PR #138, opened 2017-09-20
Feat/relay modern

thanh-nguyen-dang:
r @philloooo


data/gqlHelper.js.njk
```diff
@@ -0,0 +1,129 @@
+//
+// !!! DO NOT EDIT !!!!
+//
+// This file was auto-generated by: node data/gqlSetup.js 
+// using the Nunjucks template at data/gqlHelper.js.njx
```
philloooo: didn't see this file?

src/localconf.js
```diff
@@ -1,155 +1,206 @@
...
-  login = {
-    url: 'https://itrusteauth.nih.gov/affwebservices/public/saml2sso?SPID=https://bionimbus-pdc.opensciencedatacloud.org/shibboleth&RelayState=',
-    title: 'Login from NIH'
+  let userapiPath, submissionApiPath, submissionApiOauthPath, credentialPath, credentialOauthPath,
+    credentialCdisPath, graphqlPath, appname, navItems, login, graphqlSchemaUrl;
+  let requiredCerts = [];
+
+  // Support data/gqlSetup.js auto-generation of data-portal graphQL
+  let gqlSetup = {
+    fileTypeList: [
```
philloooo: can this be parsed from dictionary?
our goal is we don't need to update anything when we setup new commons

data/gqlHelper.js.njk
frickjack: That's the template - the generated gqlHelper.js is the thing that should not be edited ...

philloooo: `gqlHelper.js.njx` typo? actually njk?

src/localconf.js
frickjack: Yeah - we can 'curl' the dictionary from http://localhost/api/v0/submission/_dictionary/_all , but then we have a dependency on an external service.

As a practical matter we already have to edit localconf.js to deploy a new commons.

Either way is messy.

Anyway - this is a first step ...

philloooo: current editting in localconf for new commons is mainly about portal related configurations that's not described in dictionary. 
The data related configuration though is already represented in dictionary, duplicating the same info here would be error prone and also means updating dictionary for a commons every time need to update portal code.
We already have this external service dependency to get schema in data folder,  I think it's an 
 innevitable dependency? either automatically get the info from the external service or manually map the info into portal code and make sure it's in sync?

PR #144, opened 2017-09-25
fix(QueryNode): fix view-node popup

### LINKED ISSUE #142: view button doesn't work anymore
the fix/popup seems to break the view button on the same page?

frickjack:
fix #142 
Fixes the 'view node' popup that I broke when fixing the 'delete failed' popup.


frickjack:
merged the popup fix into feat/explorer which has its own pending pull request

PR #162, opened 2017-10-12
feat(collapsed): reduce displayed list

thanh-nguyen-dang:
r @frickjack @philloooo 
Make the list of choices (projects, file types, file formats) collapsible
#161


frickjack:
Hi @thanh-nguyen-dang ! 
I think that the onClick events should be changed
         <a href="#/" onClick={() => this.state.collapsed = 2}>{'More options'}</a>
to set the state via 'setState':
       this.setState( { ...this.state, collapsed: 2 } )
, so that the React component-update cycle gets triggered ...

https://reactjs.org/docs/react-component.html#setstate

PR #169, opened 2017-10-19
Test/button tests

### LINKED ISSUE #147: create regression tests for user actions on buttons

frickjack:
Just added a few tests at the 'presentation component' level to 
resolve #147


src/DataDictionary/DictionaryGraph.test.jsx
```diff
@@ -0,0 +1,47 @@
...
+
+import DictionaryGraph from './DictionaryGraph';
+import { buildTestData } from '../DataModelGraph/testData';
+
+
+describe('the DictionaryGraph', () => {
+  function buildTest() {
+    const data = buildTestData();
+    // Material-UI components require the Mui theme ...
+    const $dom = mount(
```
thanh-nguyen-dang: use enzyme to find and select the DOM

frickjack: Hi Thanh!

That 'mount' function is an enzyme function - I was just naming my variables with a '$' sign to mark them as being enzype wrappers, but I can remove the '$' from the names if  that's to weird or confusing - what do you think?

PR #176, opened 2017-10-24
Chore/submit feedback

### LINKED ISSUE #163: add summary by type to submission page
a successful submission should give some summary data by type - ex:

created:
* 55 aligned reads
* 1000 unaligned reads

Something like that.

frickjack:
resolve #163 
introduce SubmissionResult component to present submission summary:

* 200 success - list number of entities submitted by type
* 4XX error - list entities with errors
* 504 timeout - explain timeout, suggest smaller submission

### DIFF
--- a/src/Submission/ReduxProjectSubmission.js
+++ b/src/Submission/ReduxProjectSubmission.js
@@ -48,22 +48,12 @@ const submitToServer = (methodIn = 'PUT') => (dispatch, getState) => {
     body: file,
     dispatch,
   }).then(
-    ({ status, data }) => {
-      switch (status) {
-      case 200:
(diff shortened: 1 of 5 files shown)


frickjack:
Those 4 new Codacy issues result from the new code added, but do not actually reflect a decrease in code quality:

* import 'brace' is needed by AceEditor to work properly for some reason - even though we don't use it directly
* propTyps.object is not allowed is problematic, because the submission result data has a different shape depending on success or failure - this is not a regression
* indentation 8 spaces is an eslint bug - if you indent 8, then it complains about that too
* updating 'db' is idiomatic for the .reduce() function

PR #197, opened 2017-11-30
Chore/router upgrade

### LINKED ISSUE #165: update to react-router v4
We are currently on v2.
See https://github.com/ReactTraining/react-router/blob/master/packages/react-router/docs/guides/migrating.md
Pretty sure this is required to move to React v16 (currently on v15)

frickjack:
Upgraded from router v2 to v4.  API changed, so:

* introduced 'ProtectedContent' container to manage access to pages that require auth - replaces withBoxAndNav, withAuthTimeout, and requireAuth wrappers
* changed access to router-related properties in the code - new router has new and different props
* also tacked on an optional cache to fetchJsonOrText to avoid repeatedly fetching read-only data like the dictionary and schema
* introduce do-nothing Jenkinsfile - as place holder

resolve #165


src/Certificate/Quiz.jsx
```diff
@@ -1,14 +1,15 @@
 import React, { Component } from 'react';
-import { Form, FormError, RadioGroup, Radio } from 'react-form';
+import { Form, RadioGroup, Radio } from 'react-form';
```
philloooo: why is this removed?

frickjack: router-form v2 doesn't have FormError it seems ... :-(
https://react-form.js.org/#/
https://github.com/react-tools/react-form

philloooo: oh sad :/

PR #204, opened 2018-01-15
feat(ndh): add a new version for niaid

thanh-nguyen-dang:
r @philloooo @frickjack @zflamig 
Change homepage for NIAID website
- change data summary to display: case, study, lab result and socio-demographic count
- add transaction logs table
- change bar chart to display core_visit count and case count
* All tests of ExplorerPage will fail because we are not using that page in the portal.


src/Homepage/ProjectBarChart.jsx
```diff
@@ -52,7 +52,7 @@ class ProjectBarChart extends React.Component {
             <Tooltip />
             <Legend />
             <Bar dataKey="caseCount" stackId="a" fill="#8884d8" />
-            <Bar dataKey={tor.translate('experimentCount')} stackId="a" fill="#82ca9d" />
+            <Bar dataKey="coreVisitCount" stackId="a" fill="#82ca9d" />
```
philloooo: should still be translate for experimentCount and let coreVisit be the translated result for ndh

philloooo:
1. reset all dictionary/schema file replacements for ndh
2. shouldn't have if condition in each js file, should just have one centralized config in localconf.js or a new file that has for example:
```
{
   'ndh': {
      'summary_counts':[
        {'name': 'Cases', 'node': 'case'},
        {'name': 'Experiments', 'node': 'experiment'},
        ...
      ],
      'summary_bar_charts': [
        {'name': 'CoreVisits', 'node': coreVisit'},
        ...
      ]
}
```
3.  add logo image as one of the app specific configuration, if not provided the website should still work.
4. add external data link as one of the app specific configuraiton, if not provided, default to the data page.

PR #239, opened 2018-03-05
feat(multiple): multiple login

thanh-nguyen-dang:
r @frickjack 
Multiple login ticket

### DIFF
--- a/src/Login/ReduxLogin.js
+++ b/src/Login/ReduxLogin.js
@@ -1,0 +1,52 @@
+import { connect } from 'react-redux';
+
+import Login from './Login';
+import { fetchJsonOrText } from '../actions';
+import { loginPath } from '../localconf';
+
+
+export const fetchLogin = () =>
+  dispatch =>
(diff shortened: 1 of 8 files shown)


src/Login/ReduxLogin.js
```diff
@@ -0,0 +1,41 @@
...
+    })
+      .then(
+        ({ status, data }) => {
+          switch (status) {
+          case 200:
+            return {
+              type: 'RECEIVE_LOGIN_ENDPOINT',
+              providers: data.providers,
+            };
+          default:
```
frickjack: Would you make a '404' fetchLogin() response default to 'Google login', so we don't have locked dependencies between this update and a fence update?

Also - can you configure https://thanhd.planx-pla.net/ with multiple logins, so I can hit https:/thenhd.planx-pla.net/dev.html to see a working example of multiple logins?  :-)
    https://github.com/uc-cdis/cdis-wiki/tree/master/dev#k8s-dev-namespace

PR #256, opened 2018-04-03
Feat/download

zflamig:
Link file name in the Explorer tab to the download file endpoint

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -34,6 +34,10 @@ RUN COMMIT=`git rev-parse HEAD` && echo "export const portalCommit = \"${COMMIT}
     && cp nginx.conf /etc/nginx/conf.d/nginx.conf \
     && rm /etc/nginx/sites-enabled/default
 
+# In standard prod these will be overwritten by volume mounts
+# Provided here for ease of use in development and 
+# non-standard deployment environments
+
 RUN mkdir /mnt/ssl \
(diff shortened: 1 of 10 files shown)


src/Explorer/ExplorerTable.jsx
```diff
@@ -55,7 +56,9 @@ export class ExplorerTableComponent extends Component {
         <TableData c_width={columnWidths[0]}>
           <Link to={`/${file.project_id}`}>{file.project_id}</Link>
         </TableData>
-        <TableData c_width={columnWidths[1]}>{file.name}</TableData>
+        <TableData c_width={columnWidths[1]}>
```
philloooo: we should not show them as link if user doesn't have read-storage permission on this project?

src/Explorer/ExplorerComponent.jsx
```diff
@@ -50,6 +50,7 @@ class ExplorerComponent extends Component {
   mapDataToFile = filesList => filesList.map(
     file => (
       { project_id: file.project_id,
+        uuid: file.id,
```
philloooo: did？

Dockerfile
```diff
@@ -33,4 +33,8 @@ RUN COMMIT=`git rev-parse HEAD` && echo "export const portalCommit = \"${COMMIT}
     && NODE_ENV=production ./node_modules/.bin/webpack --bail \
     && cp nginx.conf /etc/nginx/conf.d/nginx.conf \
     && rm /etc/nginx/sites-enabled/default
+
+RUN mkdir /etc/nginx/ssl \
```
philloooo: Not for prod?

zflamig: wrong path, good catch.

Issue #211, opened 2018-01-24
check if need to refresh session for all api calls


philloooo:
fence sets up two cookies (after uc-cdis/fence#37 is merged):
an access_token cookie and a session_token cookie.
The default TTL for access_token is 10 min, it's a stateless token(we can't revoke it) so the TTL need to be short. The default TTL for session cookie is 30min for stale session and 8h max life time (settings here )

Whenever fence is hit, it will refresh session token if the lifetime is < 8h, and also create a new access_token if the old one is expired but the session is not expired yet.
On the user end though, any action to any api count as a 'user activity', so even if it hits sheepdog it should also refresh the session.

So we should
- wrap our non-fence api calls to handle that when the response code is 401/403, call fence /user endpoint, which will refresh the session, if fence also return 401/403, that means the session is actually expired, should pop up the login window, if it's 200, redo the previous api call.
- remove the [setinterval](https://github.com/uc-cdis/data-portal/blob/master/src/index.jsx#L40) that call fence every minute

trevars:
Estimate 6.

thanh-nguyen-dang:
spent 8 hours

PR #257, opened 2018-04-10
feat(session): support sliding session

thanh-nguyen-dang:
r @frickjack


src/actions.js
```diff
@@ -18,6 +18,38 @@ export const connectionError = () => {
...
+        }
+      } catch (e) {
+        // # do nothing
+      }
+    }
+    return { response, data, status: response.status, headers: response.headers };
+  });
+
+export const fetchAuth0 = (opts) => {
+  const { path, method = 'GET', dispatch } = opts;
```
frickjack: Hey Thanh,

It would be cool to patch `fetchAuth0` so that it does not accept a `path` or `method` - it just always does the `auth` request to fence.  

After making that change - you can wire up `fetchAuth0` with a mutex, so that only one is running at a time - ex:
```
var pendingRequest = null;
function fetchAuth0() {
   if ( pendingRequest ) { return pendingRequest; }
  pendingRequest = fetch( ... ).then( 
     function() { pendingRequest = null; ... },
     function() { pendingRequest = null; ... }
  );
  return pendingRequest;
}
```
That way multiple simultaneous overlapping async fetch requests will only result in a single Auth0 request that they'll all wait for.  What do you think?

src/actions.js
```diff
@@ -18,6 +18,38 @@ export const connectionError = () => {
...
+          fetchCache[path] = textData;
+        }
+      } catch (e) {
+        // # do nothing
+      }
+    }
+    return { response, data, status: response.status, headers: response.headers };
+  });
+
+export const fetchAuth0 = (opts) => {
```
philloooo: can we change the name? we are not talking to auth0

thanh-nguyen-dang:
resolve #211

PR #282, opened 2018-05-25
feat(nc): new commons support

### LINKED ISSUE #284: PXD-600 ⁃ fix titles for all the commons
**all commons that include cancer** should be called data commons
bhc is also Commons
only NIAID should be data hub (**"hub" should not be used in any other commons**) 
NHLBI (dcp.bionimbus.org) should be DATA STAGE (remove data commons pilot)

zflamig:
Resolves #284

### DIFF
--- a/data/parameters.js
+++ b/data/parameters.js
@@ -749,106 +749,106 @@ const params = {
     graphql: {
       boardCounts: [
         {
-            graphql: '_data_collection_count',
-            name: 'Data Collection',
-            plural: 'Data Collections',
+          graphql: '_data_collection_count',
+          name: 'Data Collection',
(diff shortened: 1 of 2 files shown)


data/parameters.js
```diff
@@ -78,8 +78,8 @@ const params = {
       appName: 'The Brain Commons Portal',
       index: {
         introduction: {
-          heading: 'Data Commons Hub',
-          text: 'The Brain Health Data Commons supports the management, analysis and sharing of brain health data for the research community and ' +
+          heading: 'Data Commons',
```
philloooo: should this be 'Brain Commons'?

PR #296, opened 2018-06-18
chore(toggle): remove unecessary libs

thanh-nguyen-dang:


src/components/tables/base/TableFoot.jsx
```diff
@@ -1,6 +1,7 @@
 import React, { Component } from 'react';
 import PropTypes from 'prop-types';
 import styled from 'styled-components';
+import TableHead from './TableHead';
```
abgeorge7: I don't think this import is being used

src/components/tables/base/TableHead.jsx
```diff
@@ -1,6 +1,7 @@
 import React, { Component } from 'react';
 import PropTypes from 'prop-types';
 import styled from 'styled-components';
+import Table from './Table';
```
abgeorge7: Same here

PR #307, opened 2018-07-11
[PXD-1108] New (hidden) route for Data Explorer

abgeorge7:
Added a new route with skeleton page for the Explorer. Hidden unless user turns on the flag in session storage. Also added `mock-local-storage` for tests, otherwise mocking `sessionStorage` was not working.


src/helpers/featureFlags.js
```diff
@@ -0,0 +1,16 @@
+/* To set a feature flag:
+
+In the browser dev tools console, type:
```
thanh-nguyen-dang: we just need to: either hide the page, point to the existing shiny page or another page instead of using featureflag. The problem is it requires another command from user/dev

thanh-nguyen-dang:
remember to squash before merging

PR #309, opened 2018-07-12
PXD-999 feat(vis): data summary card and chart components

qingyashu:


src/components/charts/PercentageStackedBarChart.jsx
```diff
@@ -0,0 +1,76 @@
+import { BarChart, Bar, Legend, Tooltip, XAxis, YAxis, CartesianGrid,
+} from 'recharts';
+import PropTypes from 'prop-types';
+import React from 'react';
+import _ from 'underscore';
+import { localTheme } from '../../localconf';
```
abgeorge7: is there a way to pass this as a prop vs. importing? I just feel like the components shouldn't be reaching into the config file

src/components/charts/PercentageStackedBarChart.jsx
```diff
@@ -0,0 +1,76 @@
...
+    {name: 'H1N1', value: 400},
+    {name: 'VN1203', value: 300},
+    {name: 'HIV', value: 300},
+    {name: 'HuCoV_EMC', value: 200},
+    {name: 'SARS_CoV', value: 278},
+    {name: 'CA04', value: 189}
+  ];
+*/
+PercentageStackedBarChart.propTypes = {
+  data: PropTypes.arrayOf(PropTypes.object).isRequired,
```
abgeorge7: so you can actually put the stuff you have in the comment in the prop!
```
data: PropTypes.arrayOf(PropTypes.shape({
     name: PropTypes.string,
     value: PropTypes.number,
})).isRequired
```

src/components/charts/SummaryBarChart.jsx
```diff
@@ -0,0 +1,60 @@
+
+import {
+  ResponsiveContainer, BarChart, Bar, Legend, Tooltip, XAxis, YAxis, LabelList, Cell,
+} from 'recharts';
+import styled from 'styled-components';
+import PropTypes from 'prop-types';
+import React from 'react';
+import { localTheme } from '../../localconf';
```
abgeorge7: Same here

src/components/charts/SummaryBarChart.jsx
```diff
@@ -0,0 +1,60 @@
...
+          </Bar>
+        </BarChart>
+      </div>
+    );
+  }
+}
+
+SummaryBarChart.propTypes = {
+  title: PropTypes.string.isRequired,
+  data: PropTypes.arrayOf(PropTypes.object).isRequired,
```
abgeorge7: Same thing here about using `PropTypes.shape`, if there's a specific data structure

src/components/charts/SummaryPieChart.jsx
```diff
@@ -0,0 +1,61 @@
...
+import PropTypes from 'prop-types';
+import React from 'react';
+import { localTheme } from '../../localconf';
+
+const pieChartContainerStyle = {
+  height: '200px',
+  width: '200px',
+};
+
+const pieChartStyle = {
```
abgeorge7: This and `pieStyle` aren't being used anywhere, do we need them?

qingyashu: Yes I just leave this for style improving, will change according to SPEC.

src/components/charts/SummaryBarChart.jsx
qingyashu: I can try let util function fetch `localConf` instead

src/DataExplorer/DataExplorer.jsx
```diff
@@ -1,10 +1,59 @@
 import React from 'react';
+import DataSummaryCardGroup from '../components/DataSummaryCardGroup';
+import SummaryChartGroup from '../components/charts/SummaryChartGroup';
+import PercentageStackedBarChart from '../components/charts/PercentageStackedBarChart';
+
+const countItems = [
```
thanh-nguyen-dang: don't hard code in the component, you should have some list parameters and the way to load data into that component.

src/components/charts/PercentageStackedBarChart.jsx
thanh-nguyen-dang: we expect to not see any localtheme or localconf in the component at all.
The component should be self-contained and everything will be loaded as properties

src/components/charts/SummaryPieChart.jsx
```diff
@@ -0,0 +1,61 @@
+
+import {
+  ResponsiveContainer, PieChart, Pie, Legend, Tooltip, Cell,
+} from 'recharts';
+import styled from 'styled-components';
+import PropTypes from 'prop-types';
+import React from 'react';
+import { localTheme } from '../../localconf';
```
thanh-nguyen-dang: again, the component should be self-contained don't use localconf or localtheme. It should not aware of any configuration of the project

src/DataExplorer/DataExplorer.jsx
qingyashu: Oh this is just dummy data here~ we are developing this page~

src/components/DataSummaryCardGroup.jsx
```diff
@@ -0,0 +1,86 @@
...
+import PropTypes from 'prop-types';
+
+const DataSummaryCardGroupWrapper = styled.div`
+  display: flex;
+  &.connected-count-card-wrapper {
+    width: 100%;
+  }
+`;
+
+const SummaryCard = styled.div`
```
thanh-nguyen-dang: we have very similar CoundCard, if you replicate code/style here, we need to change in different places when the style changed. So, please separate it using the existing CountBox in CountCard and add multiple CountBox to the group.

src/DataExplorer/DataExplorer.jsx
thanh-nguyen-dang: even you are developing the page. Don't use hardcode data inside the component @qingyashu

src/components/cards/DataSummaryCardGroup.jsx
```diff
@@ -0,0 +1,111 @@
...
+`;
+
+/**
+ * Little card with a bunch of counters on it for cases, experiments, files, ...
+ */
+class DataSummaryCardGroup extends Component {
+  render() {
+    const widthPercentage = 100 / this.props.summaryItems.length;
+    return (
+      <DataSummaryCardGroupWrapper className={'data-summary-card-group '.concat(this.props.connected ? 'connected-data-summary-card-group' : 'separated-data-summary-card-group')}>
```
abgeorge7: I think we should write a test for this
Also I don't think the `'data-summary-card-group'` class name is necessary, just the connected vs. not connected.

src/components/cards/DataSummaryCardGroup.jsx
```diff
@@ -0,0 +1,111 @@
...
+        {
+          this.props.summaryItems.map((item, index) => (
+            <SummaryCard key={`card-${item.label || item[0].label}`} widthPercentage={widthPercentage}>
```
abgeorge7: for simplicity and uniqueness, the key can just be `index`

src/components/cards/DataSummaryCardGroup.jsx
```diff
@@ -0,0 +1,111 @@
...
+              {this.props.connected && index > 0
+                && <SummaryCardLeftBorder className="left-silver-border" />
```
abgeorge7: @thanh-nguyen-dang @frickjack want your opinion on this - I think if we start mixing class names and these styled components it's going to get way too confusing... what's the better way to accomplish this?

I see that a lot of these class names have been added to `base.less` and I think we should figure out a way to have CSS in one place

src/components/cards/DataSummaryCardGroup.jsx
```diff
@@ -0,0 +1,111 @@
...
+              }
+              {
+                !item.length ? (<CountBox label={item.label} value={item.value} align="center" />) : (
+                  <SubCountBoxesWrapper>
+                    {
+                      item.map((subItem, subIndex) => (
+                        <SubCountBoxWrapper key={`sub-count-box-${subItem.label}`}>
```
abgeorge7: I don't think two wrappers is necessary here

src/stories/index.jsx
```diff
@@ -5,32 +5,76 @@ import 'react-select/dist/react-select.css';
 import '../css/base.less';
 import '../css/graphiql.css';
 
-import CountCard from '../components/CountCard';
+import CountCard from '../components/cards/CountCard';
```
abgeorge7: I think we should move all the cards to a `cards.jsx` file

src/components/charts/SummaryPieChart.jsx
```diff
@@ -0,0 +1,162 @@
...
+  justifyContent: 'flex-end',
+  marginTop: '22px',
+};
+
+const LegendWrapper = styled.div`
+  text-align: right;
+  margin-bottom: 22px;
+`;
+
+const LegendName = styled.div`
```
abgeorge7: think we can scrap this

src/components/charts/SummaryPieChart.jsx
```diff
@@ -0,0 +1,162 @@
...
+
+const LegendValue = styled.div`
+  margin-top: 8px;
+`;
+
+const calculateBarChartData = (data, showPercentage, percentageFixedPoint) => {
+  if (showPercentage) {
+    let sum = 0;
+    data.forEach((d) => { sum += d.value; });
+    let percentLeft = 100;
```
abgeorge7: can this be `percentRemaining` ? sorry picky 😝

src/components/charts/SummaryPieChart.jsx
```diff
@@ -0,0 +1,162 @@
...
+    return data.map((entry, index, array) => {
+      let percentage;
+      if (index < array.length - 1) {
```
abgeorge7: I don't think this index checking is necessary

src/components/charts/SummaryPieChart.jsx
```diff
@@ -0,0 +1,162 @@
...
+        percentage = (entry.value * 100) / sum;
+      } else {
+        percentage = percentLeft;
+      }
+      percentage = Number(Number.parseFloat(percentage).toFixed(percentageFixedPoint));
```
abgeorge7: Is this parsing needed?

src/components/charts/SummaryPieChart.jsx
```diff
@@ -0,0 +1,162 @@
...
+      percentLeft -= percentage;
+      return Object.assign({ percentage }, entry);
```
abgeorge7: This modifies the data right?

src/components/charts/SummaryPieChart.jsx
```diff
@@ -0,0 +1,162 @@
...
+    });
+  }
+  return data;
+};
+
+class SummaryPieChart extends React.Component {
+  render() {
+    const useTwoColors = this.props.data.length === 2;
+    const getCategoryColor = (index) => {
```
abgeorge7: this constant is really confusing. I think this goes back to my earlier comment about standardizing CSS and we should revisit this

src/components/charts/SummaryPieChart.jsx
```diff
@@ -0,0 +1,162 @@
...
+      const i = (index % 9) + 1;
+      return this.props.localTheme[`barGraph.bar${i}Color`];
+    };
+    const barChartData = calculateBarChartData(
+      this.props.data,
+      this.props.showPercentage,
+      this.props.percentageFixedPoint,
+    );
+    const dataKey = this.props.showPercentage ? 'percentage' : 'value';
+    const toPercentageFormatter = v => (this.props.showPercentage ? `${v}%` : v);
```
abgeorge7: lets move this out of `render`

src/components/charts/SummaryPieChart.jsx
```diff
@@ -0,0 +1,162 @@
...
+`;
+
+const pieChartStyle = {
+  flexGrow: 1,
+  display: 'flex',
+  justifyContent: 'flex-end',
+  marginTop: '22px',
+};
+
+const LegendWrapper = styled.div`
```
abgeorge7: Do we need this since we already have `PieCharLegendWrapper`?

src/components/charts/SummaryPieChart.jsx
```diff
@@ -0,0 +1,162 @@
...
+    return (
+      <PieChartWrapper>
+        <PieChartTitle className="h4-typo">
+          {this.props.title}
+        </PieChartTitle>
+        <PieChartInnerWrapper>
+          <PieChartLegendWrapper>
+            {
+              barChartData.map(entry => (
+                <LegendWrapper className="form-body" key={'text'.concat(entry.name)}>
```
abgeorge7: key can just be index

src/components/charts/SummaryPieChart.jsx
```diff
@@ -0,0 +1,162 @@
...
+                  <LegendName>
```
abgeorge7: Would push to rename this - it's not the name of the legend but the name of a legend item

src/components/charts/SummaryPieChart.jsx
```diff
@@ -0,0 +1,162 @@
...
+                    {entry.name}
+                  </LegendName>
+                  <LegendValue className="form-special-number">
+                    {
+                      this.props.showPercentage ? `${entry[dataKey]}%` : entry[dataKey]
```
abgeorge7: wasn't this defined in `toPercentageFormatter` ?
Would also push to rename that function since it doesn't necessarily return a percentage

src/components/charts/SummaryPieChart.jsx
```diff
@@ -0,0 +1,162 @@
...
+});
+
+SummaryPieChart.propTypes = {
+  title: PropTypes.string.isRequired,
+  data: PropTypes.arrayOf(ChartDataShape).isRequired,
+  localTheme: PropTypes.object.isRequired,
+  innerRadius: PropTypes.number,
+  outerRadius: PropTypes.number,
+  showPercentage: PropTypes.bool,
+  percentageFixedPoint: PropTypes.number,
```
abgeorge7: I'm not sure what this prop does exactly

src/components/charts/SummaryBarChart.jsx
```diff
@@ -0,0 +1,139 @@
...
+
+const labelValueStyle = {
+  fontSize: '10px',
+  fontWeight: 600,
+  lineHeight: '1em',
+  letterSpacing: '.02rem',
+  color: '#3283c8',
+};
+
+const calculateBarChartData = (data, showPercentage, percentageFixedPoint) => {
```
abgeorge7: this is the same function for pie chart - can take out and put into a helper file somewhere to avoid duplicate code

src/components/charts/SummaryBarChart.jsx
```diff
@@ -0,0 +1,139 @@
...
+};
+
+class SummaryBarChart extends React.Component {
+  render() {
+    const getCategoryColor = (index) => {
+      // map index to (1-9)
+      const i = (index % 9) + 1;
+      return this.props.localTheme[`barGraph.bar${i}Color`];
+    };
+    const monoFillColor = this.props.monoColor ? this.props.color : undefined;
```
abgeorge7: this seems like an extra prop - why do you need both `color` and `monoColor`?

src/components/charts/SummaryBarChart.jsx
```diff
@@ -0,0 +1,139 @@
...
+    const barChartHeight = (this.props.data.length * this.props.barSize)
+      + ((this.props.data.length + 1) * this.props.barGap) + 2;
+    const barChartData = calculateBarChartData(
+      this.props.data,
+      this.props.showPercentage,
+      this.props.percentageFixedPoint,
+    );
+    const dataKey = this.props.showPercentage ? 'percentage' : 'value';
```
abgeorge7: this line and the next were also in SummaryPieChart - can move into helper func

src/components/charts/SummaryBarChart.jsx
```diff
@@ -0,0 +1,139 @@
...
+const ChartDataShape = PropTypes.shape({
+  name: PropTypes.string.isRequired,
+  value: PropTypes.number.isRequired,
+});
+
+SummaryBarChart.propTypes = {
+  title: PropTypes.string.isRequired,
+  data: PropTypes.arrayOf(ChartDataShape).isRequired,
+  localTheme: PropTypes.object.isRequired,
+  vertical: PropTypes.bool,
```
abgeorge7: I'm not sure these `vertical` and `monoColor` do anything - when I remove them in Storybook nothing changes

src/components/charts/PercentageStackedBarChart.jsx
```diff
@@ -0,0 +1,166 @@
+import {
+  BarChart, Bar, Tooltip, XAxis, YAxis,
+  CartesianGrid, LabelList,
+} from 'recharts';
+import PropTypes from 'prop-types';
+import React from 'react';
+import _ from 'underscore';
+import styled from 'styled-components';
+
+const getPercentageData = (chartData, percentageFixedPoint) => {
```
abgeorge7: also very similar to the functions in SummaryPieChart and SummaryBarChart - I would move into a global helper function

src/components/charts/PercentageStackedBarChart.jsx
```diff
@@ -0,0 +1,166 @@
...
+  margin-left: 6px;
+`;
+
+const LegendValueText = styled.span`
+  margin-left: 6px;
+`;
+
+class PercentageStackedBarChart extends React.Component {
+  render() {
+    const getCategoryColor = (index) => {
```
abgeorge7: also a reusable function - it is in the bar chart and pie chart files

src/components/charts/PercentageStackedBarChart.jsx
```diff
@@ -0,0 +1,166 @@
...
+          <ul>
+            {
+              percentageDataLabels.map((name, index) => (
+                <li key={`label-${name}`}>
+                  <LegendColorBlock fill={getCategoryColor(index)} />
+                  <LegendNameText className="form-body">
+                    {name}
+                  </LegendNameText>
+                  <LegendValueText className="form-caption">
+                    {'('.concat(Number(this.props.data[index].value).toLocaleString()).concat(')')}
```
abgeorge7: what does this line do?

src/components/cards/CountBox.jsx
```diff
@@ -0,0 +1,76 @@
...
+
+class CountBox extends Component {
+  render() {
+    return (
+      <CountBoxWrapper className={`align-${this.props.align}`}>
+        <CountBoxTitle className="h4-typo">
+          {this.props.label}
+        </CountBoxTitle>
+        <CountBoxNumber className="special-number">
+          {Number(this.props.value).toLocaleString()}
```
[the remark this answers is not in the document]
qingyashu: Here is to transfer number to comma separated form

src/components/cards/DataSummaryCardGroup.jsx
qingyashu: I tried `index` just got some warning like `please don't use index in key` something...Anyway~

src/components/cards/DataSummaryCardGroup.jsx
qingyashu: These are for "flex" style div tags, because there are 2-level sub hierarchies. Or are you thinking to put 2-level into 1-level? I think this will make 1-st level css code more complicated

src/components/charts/SummaryBarChart.jsx
qingyashu: I agree!

src/components/charts/SummaryBarChart.jsx
qingyashu: There is something wrong with "vertical", I am still checking...I think "monoColor" should work 😛

src/components/charts/PercentageStackedBarChart.jsx
qingyashu: This is to show actual number (instead of percentage) in legends, format is like "(123,456)"

src/components/charts/SummaryPieChart.jsx
qingyashu: Do you mean calculating the last percentage? This is because we want to make sure all percentage will sum up to 100: sometimes rounding process (like toFixed(2)) will cause slightly differences and make the total sum be 99 or less...

src/components/charts/SummaryPieChart.jsx
qingyashu: Yes this add `percentage` attribute

src/components/charts/SummaryPieChart.jsx
qingyashu: `LegendWrapper` is child items inside `PieChartLegendWrapper` so we still need it to set margin of each child. Sorry I am bad at creating names

src/components/charts/SummaryPieChart.jsx
qingyashu: This is for how many digits after point, like "%.2f" in cpp.printf

src/components/cards/CountBox.jsx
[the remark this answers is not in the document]
abgeorge7: I'm pretty sure `toLocaleString()` puts numbers into the local language - I don't think it puts them into comma separated. Can you use `.join(',')` instead?

src/components/charts/SummaryPieChart.jsx
abgeorge7: Oh okay got it!

src/components/charts/SummaryPieChart.jsx
abgeorge7: Can you return a new object instead of modifying the existing data object? I feel weird modifying data directly

qingyashu: Oh Array.map function will return new data. I am sorry to be confusing...I just thought you were talking about Object.assign will modify the data object 😛

src/components/cards/DataSummaryCardGroup.jsx
```diff
@@ -0,0 +1,86 @@
...
+        className="data-summary-card-group"
+        style={{
+          width: totalWidth,
+          height: this.props.height,
+        }}
+      >
+        {
+          this.props.summaryItems.map((item, index) => (
+            <div
+              className={'data-summary-card-group__card'.concat(this.props.connected ? ' connected' : ' separated')}
```
abgeorge7: I think here there can be two separate classes of `.data-summary-card-group__connected-card` and `.data-summary-card-group__separated-card` or something.

src/components/cards/DataSummaryCardGroup.jsx
```diff
@@ -0,0 +1,86 @@
...
+              key={item.label || item[0].label}
+            >
+              {this.props.connected && index > 0
+                && <div className="data-summary-card-group__card-left-border" />
```
abgeorge7: can this be encapsulated by the `.data-summary-card-group__connected-card` class?

src/components/cards/DataSummaryCardGroup.jsx
```diff
@@ -0,0 +1,86 @@
...
+                    value={item.value}
+                    align={this.props.align}
+                  />
+                ) : (
+                  <div className="data-summary-card-group__sub-card-group">
+                    {
+                      item.map((subItem, subIndex) => (
+                        <div className="data-summary-card-group__sub-card-item" key={subItem.label}>
+                          {subIndex > 0
+                            && <div className="data-summary-card-group__sub-card-left-border" />
```
abgeorge7: Why does there need to be a separate class for this?

src/components/charts/ChartsHelper.js
```diff
@@ -0,0 +1,61 @@
+
+const percentageFormatter = showPercentage => v => (showPercentage ? `${v}%` : v);
```
abgeorge7: should this just be 
`const percentageFormatter = (showPercentage, v) => (showPercentage ? `${v}%` : v);` ?

I've never seen double arrows before

src/components/charts/ChartsHelper.js
```diff
@@ -0,0 +1,61 @@
...
+
+const addPercentage = v => (percentageFormatter(true)(v));
+
+const calculateBarChartData = (data, showPercentage, percentageFixedPoint) => {
```
abgeorge7: Is this function specific to bar charts? I think it can potentially be renamed

src/components/charts/ChartsHelper.js
```diff
@@ -0,0 +1,61 @@
+
```
abgeorge7: because this file isn't a component I think it can be renamed just to `helpers`

src/components/charts/ChartsHelper.js
```diff
@@ -0,0 +1,61 @@
...
+      }
+      percentage = Number(Number(percentage).toFixed(percentageFixedPoint));
+      percentRemaining -= percentage;
+      return Object.assign({ percentage }, entry);
+    });
+  }
+  return data;
+};
+
+const getPercentageData = (chartData, percentageFixedPoint) => {
```
abgeorge7: This function looks pretty much the same as the function above it, can we combine them somehow?

src/components/charts/PercentageStackedBarChart.jsx
```diff
@@ -0,0 +1,114 @@
...
+  render() {
+    const percentageData = ChartsHelper.getPercentageData(
+      this.props.data,
+      this.props.percentageFixedPoint,
+    );
+    const percentageDataLabels = getPercentageDataLabels(this.props.data);
+    return (
+      <div className="percentage-bar-chart">
+        <BarChart
+          width={580}
```
abgeorge7: will we ever want to change these values?

src/components/charts/PercentageStackedBarChart.jsx
```diff
@@ -0,0 +1,114 @@
...
+            style={xAxisStyle}
+            tickFormatter={ChartsHelper.addPercentage}
+          />
+          <YAxis axisLine={false} tickLine={false} dataKey="name" type="category" hide />
+          {
+            percentageDataLabels.map((name, index) => (
+              <Bar
+                key={name}
+                dataKey={name}
+                stackId="a"
```
abgeorge7: what does this mean?

src/components/charts/PercentageStackedBarChart.less
```diff
@@ -0,0 +1,32 @@
+
```
abgeorge7: picky but can we get rid of the extra space at the top 😝

src/components/charts/SummaryBarChart.jsx
```diff
@@ -0,0 +1,100 @@
...
+          >
+            <Tooltip formatter={ChartsHelper.percentageFormatter(this.props.showPercentage)} />
+            <XAxis axisLine={false} tickLine={false} type="number" hide />
+            <YAxis axisLine={false} tickLine={false} dataKey="name" type="category" style={yAxisStyle} interval={0} />
+            <Bar dataKey={dataKey}>
+              {
+                barChartData.map((entry, index) => (
+                  <Cell
+                    key={dataKey}
+                    fill={monoFillColor
```
abgeorge7: instead of doing the calculation for `const monoFillColor` on line 28, can this just be `fill={this.props.color || ChartsHelper.getCategoryColor(...)}` ? I don't love setting something to be undefined like its doing on line 28

src/components/charts/SummaryChartGroup.jsx
```diff
@@ -0,0 +1,54 @@
+import React, { Component } from 'react';
+import PropTypes from 'prop-types';
+import SummaryPieChart from './SummaryPieChart';
+import SummaryBarChart from './SummaryBarChart';
+import './SummaryChartGroup.less';
+
+class SummaryChartGroup extends Component {
+  render() {
+    const width = (typeof this.props.width === 'number') ? `${this.props.width}px` : this.props.width;
```
abgeorge7: This line has been used before as well, maybe could put it into a helper func

src/components/charts/SummaryChartGroup.jsx
```diff
@@ -0,0 +1,54 @@
...
+                      data={item.data}
+                      title={item.title}
+                      localTheme={this.props.localTheme}
+                    />
+                  ) : (
+                    <SummaryBarChart
+                      data={item.data}
+                      title={item.title}
+                      localTheme={this.props.localTheme}
+                      vertical
```
abgeorge7: do these props need to be set to something? ie. `vertical={true}`?

src/components/charts/SummaryPieChart.jsx
```diff
@@ -0,0 +1,103 @@
...
+          <div className="summary-pie-chart__legend">
+            {
+              barChartData.map(entry => (
+                <div className="summary-pie-chart__legend-item" key={'text'.concat(entry.name)}>
+                  <div className="form-body">
+                    {entry.name}
+                  </div>
+                  <div className="summary-pie-chart__legend-item-value form-special-number">
+                    {
+                      // this.props.showPercentage ? `${entry[dataKey]}%` : entry[dataKey]
```
abgeorge7: leftover comment

src/css/base.less
```diff
@@ -313,3 +313,31 @@ ul {
   /* Support for IE. */
   font-feature-settings: 'liga';
 }
+
+.left-silver-border {
```
abgeorge7: is this necessary as a class in the main file?

src/components/cards/DataSummaryCardGroup.jsx
qingyashu: The style is different for sub-card and card so I created 2 classes

src/components/charts/PercentageStackedBarChart.jsx
qingyashu: I agree with you!!! To me at first glance this design looks relatively special for this case (e.g. only one bar with all stacked items, 2 columns of labels in legend) so I haven't think much about which parameters shall we make it customize. But as you said we can first make the left width, height, and column number customizable~  😛

src/components/charts/PercentageStackedBarChart.jsx
qingyashu: This is to tell recharts library to stack all little bars together... (I am not very into recharts for this component, It might be easier and more readable if we use d3 or manually create SVG... 😂

src/components/charts/SummaryChartGroup.jsx
qingyashu: I agree with you! Just Codacy told me`boolean attribute` something and asked me to change into this...

qingyashu:
@abgeorge7  @frickjack  Finished writing unit tests. Could you help review this PR? Thanks

src/components/cards/DataSummaryCardGroup.jsx
```diff
@@ -0,0 +1,84 @@
...
+        }}
+      >
+        {
+          this.props.summaryItems.map((item, index) => (
+            <div
+              className={'data-summary-card-group__'.concat(this.props.connected ? 'connected' : 'separated').concat('-card')}
+              key={item.label || item[0].label}
+            >
+              {this.props.connected && index > 0
+                && <div className="left-border" />
```
abgeorge7: For this you just want to have all but the first child have a left border right? In your css you already have `.data-summary-card-group__connected-card:first-child { border: none; }`, so why don't you have `.data-summary-card-group__connected-card { border-left: ... }` in the css as well?

src/components/cards/DataSummaryCardGroup.jsx
```diff
@@ -0,0 +1,84 @@
...
+                    value={item.value}
+                    align={this.props.align}
+                  />
+                ) : (
+                  <div className="data-summary-card-group__sub-card-group">
+                    {
+                      item.map((subItem, subIndex) => (
+                        <div className="data-summary-card-group__sub-card-item" key={subItem.label}>
+                          {subIndex > 0
+                            && <div className="left-border" />
```
abgeorge7: same here

src/components/cards/DataSummaryCardGroup.less
```diff
@@ -0,0 +1,51 @@
...
+  }
+}
+
+.data-summary-card-group__connected-card {
+  margin: 0;
+  &:first-child {
+    border: none;
+  }
+
+  .left-border {
```
abgeorge7: With BEM we are trying to avoid nesting classes - I think you can move everything in `.left-border` (here and on line 45) to the top level of the class anyway, and then just have `&:first-child` take care of eliminating the border for the first children

src/components/cards/DataSummaryCardGroup.less
```diff
@@ -0,0 +1,51 @@
...
+.data-summary-card-group__sub-card-group {
+  display: flex;
+  height: 100%;
+}
+
+.data-summary-card-group__sub-card-item {
+  flex-grow: 1;
+  height: 100%;
+
+  .left-border {
```
abgeorge7: same here

src/components/charts/SummaryChartGroup.jsx
```diff
@@ -0,0 +1,55 @@
...
+class SummaryChartGroup extends Component {
+  render() {
+    const width = parseParamWidth(this.props.width);
+    return (
+      <div className="summary-chart-group" style={{ width }}>
+        {
+          this.props.summaries.map((item, index) => (
+            <div className="summary-chart-group__column" key={item.title}>
+              {
+                index > 0 && <div className="summary-chart-group__column-left-border" />
```
abgeorge7: same comment about left border

src/components/cards/DataSummaryCardGroup.jsx
qingyashu: This is because the design for border has some space between top and bottom, so only css's border attribute cannot have this effect

src/components/cards/DataSummaryCardGroup.less
qingyashu: Just thought nesting left-border is more semantic and easier to read...... 😛anyways, I can change to the top level and we now stick with BEM...

src/components/cards/DataSummaryCardGroup.jsx
qingyashu: Oh I just found &:first_child {border: none;} has not effects and should be deleted..Thanks for reminding me! 😂

PR #316, opened 2018-07-18
PXD-998 Added Arranger table to Data Explorer

abgeorge7:
Add the Arranger Table UI component to Data Explorer. I customized it but some of it doesn't quite align with the spec - discussing with Vivi. It is in the Storybook under "Data Explorer Table".


src/DataExplorer/index.jsx
```diff
@@ -0,0 +1,101 @@
+import React from 'react';
+import multisort from 'multisort';
+import DataExplorerTable from '../components/tables/DataExplorerTable/.';
+
+// Skeleton for connecting GraphQL data
+const dummyConfig = {
```
thanh-nguyen-dang: should we put it in parameters file and include this template as a property?

src/DataExplorer/index.jsx
```diff
@@ -0,0 +1,101 @@
...
+};
+
+const fetchGQL = ({ sort, offset, first }) => fetch(
+  '/api/v0/flat-search/search/graphql',
+  {
+    method: 'POST',
+    headers: { 'Content-Type': 'application/json; charset=utf-8' },
+    body: JSON.stringify(
+      {
+        query: '{ subject { hits { edges { node { name, project, study, gender, ethnicity, race, file_type }}}}}',
```
thanh-nguyen-dang: same for this query, put it in parameter file

src/DataExplorer/index.jsx
```diff
@@ -0,0 +1,101 @@
...
+      Header: 'Race',
+      type: 'string',
+      sortable: true,
+      canChangeShow: true,
+      accessor: 'race',
+    },
+  ],
+};
+
+const fetchGQL = ({ sort, offset, first }) => fetch(
```
thanh-nguyen-dang: write the `fetchGQL` in corresponding `action.js` file.
we have for each page in the portal:
```
- *.less
- *.jsx
- reduxer.js -> mapping redux values to properties
- reducer.js -> if you need some reduce functions for this page
- action.js -> if you need customized fetch, otherwise, you can use some default one in src/actions.js
```

abgeorge7:
@thanh-nguyen-dang @frickjack the graphQL hookup is going to change a lot (have to add more Arranger config) - this was meant to be more of a proof of concept or a place to start. Do you still want me to make the above changes?

thanh-nguyen-dang:
I think that changing those now better than later. It is pretty straightforward because we have those parameters file.

jbarno:
Definitely keep pack-lock.json under source control.

philloooo:
should not be removed from gitignore...

PR #317, opened 2018-07-20
PXD-998 (part 2) Initial filter styling

abgeorge7:
Can see the full filter on Storybook under FilterGroup, as well as all the components that it's composed of.


.eslintrc.js
```diff
@@ -40,5 +40,6 @@ module.exports = {
         "react/forbid-prop-types": "off",
         "react/prefer-stateless-function": "off",
         "react/jsx-curly-brace-presence": ["off", "ignore"],
+        "react/no-array-index-key" : ["off, ignore"],
```
qingyashu: I strongly agree with shutting down this warning! Just be careful for sortable or deletable items like table rows in future~ 😛

abgeorge7:
@thanh-nguyen-dang - done

.storybook/preview-head.html
```diff
@@ -1 +1,2 @@
 <link href="https://fonts.googleapis.com/icon?family=Source+Sans+Pro" rel="stylesheet">
+<link href="https://maxcdn.bootstrapcdn.com/font-awesome/4.7.0/css/font-awesome.min.css" rel="stylesheet" integrity="sha384-wvfXpqpZZVQGK6TAh5PVlGOfQNHSoD2xbE+QkPxCAFlNEevoEH3Sl0sibVcOQVnN" crossorigin="anonymous">
```
frickjack: Let's just use npm to pull in this dependency ...

PR #318, opened 2018-07-20
Feat/core metadata

paulineribeyre:
The travis tests fail for now because we replaced `id` by `object_id` in the `gqlHelper` and the tests need to be updated.


philloooo:
the default schema need to be updated because we added the object_id to all dictionaries, the one in github is necessary to run the tests. So we need to update the file to pass the tests

src/CoreMetadata/page.jsx
```diff
@@ -0,0 +1,38 @@
+import React, { Component } from 'react';
+import styled from 'styled-components';
+import BackLink from '../components/BackLink';
+import { ReduxCoreMetadataHeader } from './reduxer';
```
thanh-nguyen-dang: merge lines 4, 5, 6 into one line

src/components/BackLink.jsx
```diff
@@ -0,0 +1,32 @@
...
+
+export const BackComponent = styled.div`
+  display: inline;
+  padding-right: 15px;
+`;
+
+class BackLink extends Component {
+  render() {
+    return (
+      <Link to={'/files'}>
```
thanh-nguyen-dang: this BackLink component:
- if you want to keep it in `src/components` folder, it should have a parameter `backlink`
 with a link to page where you want to go back.
- otherwise, it should locate in a page of `src/CoreMetadata` (where you use it with hardcoded value `/files`

PR #321, opened 2018-07-26
fix(btn-text-color): remove styled component and move to less

qingyashu:
TopIconButton component

### DIFF
--- a/src/components/layout/TopIconButton.less
+++ b/src/components/layout/TopIconButton.less
@@ -1,0 +1,13 @@
+@import '../../css/base.less';
+
+.top-icon-button {
+  background-color: @blue;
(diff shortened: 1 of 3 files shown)


src/components/layout/TopIconButton.less
```diff
@@ -0,0 +1,13 @@
+@import '../../css/base.less';
+
+.top-icon-button.body-typo {
```
abgeorge7: I don't think we want both these class names in the definition, we want TopIconButton to always have the styles below right? Regardless of if it's also body-typo or not?

.body-typo only has two properties.... does TopIconButton even need to inherit from it since it's already overriding one of them?

PR #323, opened 2018-07-30
PXD-1214 fix(css): import ui-component and change all refer to base.less

qingyashu:
Move base.less to @gen3/ui-component and now importing this module.


src/components/cards/CountBox.less
```diff
@@ -1,4 +1,4 @@
-@import '../../css/base.less';
+@import '../../../node_modules/@gen3/ui-component/dist/css/base.less';
```
abgeorge7: I don't think we want to import directly from `node_modules`, can it just be `@import '@gen3/ui-component/...'` ?

abgeorge7:
The text we fixed from black to white is black again on this branch also just fyi

qingyashu:
Rebased. Check again~

PR #324, opened 2018-07-30
PXD-1308 fix(css): move style out of js into less (1/5)

qingyashu:
Working on CSS refactoring, 1 of 5 steps


src/DataDictionary/DataDictionaryNode.less
```diff
@@ -0,0 +1,40 @@
+.download-button:hover,
```
abgeorge7: I think these should be name `data-dictionary-node__download-button` since they are in the `DataDictionaryNode` component

src/DataDictionary/DataDictionaryNode.jsx
```diff
@@ -1,25 +1,23 @@
 import React from 'react';
-import styled, { css } from 'styled-components';
 import { Link } from 'react-router-dom';
 import PropTypes from 'prop-types';
-
-import { Table, TableData, TableRow, TableHead, Bullet } from '../theme';
+import { Table, TableRow, TableHead, Bullet } from '../theme';
```
abgeorge7: can we get rid of all of these too?

src/DataDictionary/DataDictionaryNode.jsx
```diff
@@ -40,7 +38,7 @@ const LinkTable = ({ links }) => {
       <TableHead>
         <TableRow>
           {fields.map(field =>
-            <TableData first_cr key={field}>{field}</TableData>)}
+            <td className="data-dictionary__table-data data-dictionary__table-data--head" key={field}>{field}</td>)}
```
abgeorge7: does `first_cr` need to be in this still? also why is the `table-data--head` class applied as well?

src/DataDictionary/DataDictionary.jsx
```diff
@@ -41,14 +39,14 @@ TableBullet.defaultProps = {
 };
 
 const CategoryTable = ({ nodes, category }) => (
-  <Table>
-    <TableHead>
-      <TableRow>
-        <TableData first_cr>
```
abgeorge7: does the new css take care of this `first_cr` prop?

src/DataDictionary/DataDictionary.less
```diff
@@ -0,0 +1,103 @@
...
+
+.data-dictionary__table-data--right {
+  width: 80%;
+}
+
+.data-dictionary__table {
+  table-layout:auto;
+  position: relative;
+  border: 0;
+  * {
```
abgeorge7: is there a way to achieve the same `box-sizing` without this?

src/DataDictionary/DataDictionaryNode.jsx
```diff
@@ -94,60 +91,48 @@ export const getType = (property) => {
...
+
+      <tr className="data-dictionary__table-row">
+        <td className="data-dictionary__table-data data-dictionary__table-data--head"> Description </td>
+        <td className="data-dictionary__table-data data-dictionary__table-data--right">{ node.description}</td>
+      </tr>
+
+      <tr className="data-dictionary__table-row">
+        <td className="data-dictionary__table-data data-dictionary__table-data--head"> Unique Keys </td>
+        <td className="data-dictionary__table-data data-dictionary__table-data--right">
+          {
```
abgeorge7: I don't think you need these extra parens

qingyashu:
This PR relates following pages:

- /quiz
- /DD
- /files
- /query
- /analysis

New styles should keep same as they were before

PR #325, opened 2018-07-31
PXD-1291 Make Arranger filters look like our spec

abgeorge7:
- Modified the Arranger <Aggregation /> component to divide sections into tabs
- If tab configuration is provided in parameters.js, then it will be divided into tabs. Otherwise it will just be normal Arranger aggregation filters that look more our style.
- Overrode Arranger styles


src/Arranger/AggregationTabs.jsx
```diff
@@ -0,0 +1,128 @@
+import React from 'react';
```
abgeorge7: This is the same file as in my last PR, just moved to a a different folder

src/Arranger/ArrangerWrapper.jsx
```diff
@@ -0,0 +1,77 @@
+import React from 'react';
```
abgeorge7: Same file as in my last PR, just moved to a new folder

src/Arranger/AggregationTabs.jsx
```diff
@@ -0,0 +1,128 @@
...
+      api={api}
+      projectId={projectId}
+      graphqlField={graphqlField}
+      render={(aggsState) => {
+        const aggs = aggsState.aggs.filter(x => x.show);
+        // Dividing data into tabs
+        const tabs = [];
+        filterConfig.tabs.forEach((tab, i) => {
+          const tabAggs = aggs.filter(agg => tab.fields.includes(agg.field));
+          tabs.push(
```
frickjack: This should be `aggs.find` instead of `aggs.filter` to avoid case where same field appears in more than one tab by accident?

src/Arranger/AggregationTabs.jsx
```diff
@@ -0,0 +1,128 @@
...
+                  .map((agg) => {
+                    if (aggComponents[agg.type]) {
+                      return (aggComponents[agg.type]({ ...agg, ...componentProps }));
+                    }
+                    return null;
+                  })
+              }
+            />,
+          );
+        });
```
frickjack: What happens to a field that is configured in arranger, but not registered in a `tab`?

abgeorge7: It would get left out I think - I will write a test for that

PR #326, opened 2018-08-02
PCD-1309 Fix/css refactor 2

qingyashu:


src/DataModelGraph/DataModelGraph.jsx
```diff
@@ -36,15 +34,13 @@ class DataModelGraph extends React.Component {
...
-          <ToggleButton>
-            <button
-              id="cd-dmg__toggle"
-              className="button-primary-white"
-              onClick={this.handleToggleClick}
-            >Toggle view
-            </button>
-          </ToggleButton>
+        <div className="data-model-graph">
+          <button
```
abgeorge7: do we need to add the styles from `ToggleButton`?

src/Explorer/ExplorerComponent.less
```diff
@@ -0,0 +1,5 @@
+.explorer-component {      
+  display: flex;
+  padding: 40px 0;
+  background-color: #ecebeb;
```
abgeorge7: where did this color come from?

src/Login/Login.jsx
```diff
@@ -93,29 +48,33 @@ class Login extends React.Component {
       next = basename === '/' ? queryParams.next : basename + queryParams.next;
     }
     return (
-      <LoginFrame>
-        <LeftBox style={{ height: `${this.state.height}px` }}>
+      <div className="login-page">
+        <div className="login-page__side-box">
```
abgeorge7: do we need the `height` that was being passed to `LeftBox`?

src/Login/ProtectedContent.jsx
```diff
@@ -61,7 +52,6 @@ export function intersection(aList, bList) {
  * @param history from react-router
  * @param match from react-router.match
  * @param public default false - set true to disable auth-guard
- * @param background passed through to <Box background> wrapper for page.jsx-level background
```
abgeorge7: do we need this param for anything?

src/components/layout/NavBar.jsx
```diff
@@ -89,51 +17,48 @@ class NavBar extends Component {
 
   render() {
     return (
-      <div style={{ width: '100%', backgroundColor: 'white', borderBottom: '1px solid #d1d1d1' }}>
-        <Header>
-          <NavLeft>
-            <NavLogo>
+      <div className="nav-bar">
+        <header className="nav-bar__header">
+          <nav className="nav-bar__nav nav-bar__nav--left">
```
abgeorge7: it doesn't look like `nav-bar__nav` is a class

src/components/layout/NavBar.less
```diff
@@ -0,0 +1,58 @@
...
+  overflow: hidden;
+}
+
+.nav-bar__header {
+  overflow: hidden;
+  margin: auto;
+  width: 1220px;
+}
+
+.nav-bar__link {
```
abgeorge7: I think this is missing the :hover and :active properties

src/components/layout/NavBar.less
```diff
@@ -0,0 +1,58 @@
...
+  color: #606060;
+  display: inline-block;
+  text-align: center;
+}
+
+.nav-bar__nav--right .nav-bar__link {
+  border-left: 1px solid #d1d1d1;
+  &:last-child {
```
abgeorge7: can we unnest 😬

src/components/layout/NavButton.less
```diff
@@ -0,0 +1,18 @@
+.nav-button {
+  padding: 16px 0px;
+  height: 80px;
+  width: 110px;
+  display: inline-block;
+  text-align: center;
+  opacity: 0.8;
+  &:hover {
```
abgeorge7: unnest

src/components/layout/TopBar.less
```diff
@@ -0,0 +1,32 @@
+.top-bar {
+  width: 100%;
+  background-color: #3283c8;
```
abgeorge7: I think this is `@blue`

src/components/layout/TopBar.less
```diff
@@ -0,0 +1,32 @@
...
+
+.top-bar__nav {
+  float: right;
+  color: white;
+  overflow: hidden;
+}
+
+.top-bar__link {
+  border-right: 2px solid #6896c6;
+  &:last-child {
```
abgeorge7: unnest

src/index.jsx
```diff
@@ -75,11 +76,11 @@ async function init() {
         <ThemeProvider theme={theme}>
           <MuiThemeProvider>
             <BrowserRouter basename={basename}>
-              <OuterWrapper>
+              <div>
                 { GA.init(gaTracking, dev, gaDebug) && <RouteTracker /> }
                 <ReduxTopBar />
                 <ReduxNavBar />
-                <Box background={background} style={{ width: '100%', margin: 'auto' }}>
```
abgeorge7: what happens with this `background` prop? should we get rid of the variable?

src/Login/ProtectedContent.jsx
qingyashu: I did some global searching, and found that `background-color` is only different for `ExplorerComponent` and was passed via `background` variable. So to make code cleaner I set `background-color`  inside `ExplorerComponent.less`, and remove this `background` parameter from `ProtectedContent` component.

src/components/layout/NavBar.jsx
qingyashu: 😆 Ah I am really bad at naming! how about `nav-bar__content`, `nav-bar__content--left`...

src/index.jsx
qingyashu: See comments in `src/Login/ProtectedContent.jsx`, I think this `background` variable was not a good design, think we better get rid of it.

src/components/layout/NavBar.less
qingyashu: :hover and :active is already defined in `NavButton.less`, previous code duplicately defined them. already tested in my local and hover/active effects work fine.

src/Login/ProtectedContent.jsx
abgeorge7: awesome, makes sense to me

src/components/layout/NavBar.jsx
abgeorge7: the name is fine, I just didn't see the class anywhere in the .less file

src/DataModelGraph/DataModelGraph.jsx
qingyashu: `ToggleButton` only has 2 css attribute `top` and `left`, They don't have any effect for the button in this page so I directly remove them

src/Login/Login.jsx
qingyashu: `height` is not used, so removed.

src/components/layout/NavBar.less
qingyashu: Thanks! I will unnest `.nav-bar__nav--right .nav-bar__link`. 
But I think it is reasonable to nest pseudo class names like `:hover` or `:active` because they only effect element itself and won't cause any side effects. Can we keep them?

abgeorge7: yeah I get that, but it doesn't don't adhere to the BEM rules, and also doesn't give us the preprocessor speed.

qingyashu: I think you are right about preprocessor speed, That's a good reason! 
Also if we want to have csslint in future then we should clean all `&` first (will be in next pr)
But I think BEM rule still allow reasonable nesting cases like this one: http://getbem.com/faq/#block-modifier-affects-elements

qingyashu:
Thanks for reminding me!! color fixed~

PR #328, opened 2018-08-03
PXD-1315 Fix linting and test errors in console

abgeorge7:
Fixed all our warnings/errors in the console to make linting and testing easier. These things are mostly default props, multiple components in one file, nested ternaries, default exports, etc.

Also put a test in that will fail if there are any console warnings, so that in the future we won't push code that causes these warnings.

Would appreciate a lot of eyes on this, I refactored a lot of code I'm not super familiar with so I want to be sure nothing breaks.


package.json
```diff
@@ -111,6 +111,7 @@
     "verbose": true,
     "automock": false,
     "setupFiles": [
+      "./src/shim.js",
```
frickjack: Can we just merge this into `setupJest.js`?

src/CoreMetadata/page.jsx
```diff
@@ -20,7 +20,7 @@ class CoreMetadataPage extends Component {
   render() {
     return (
       <div>
-        <BackLink url='/files' label='Back to data overview' />
+        <BackLink url="/files" label="Back to data overview" />
```
frickjack: We're supposed to use single-quotes in javascript, and double quotes in jsx?  Ugh.

src/DataDictionary/DataDictionary.jsx
```diff
@@ -90,9 +90,10 @@ export function category2NodeList(dictionary) {
   )
     .reduce(
       (lookup, node) => {
-        if (!lookup[node.category]) { lookup[node.category] = []; }
-        lookup[node.category].push(node);
-        return lookup;
+        let category = lookup[node.category];
+        if (!lookup[node.category]) { category = []; }
+        category.push(node);
+        return Object.assign({}, lookup, { [node.category]: category });
```
frickjack: Why are we doing this?
Every time through the reduce loop we're making a new copy of the accumulator dictionary - so we go from O(n) to O(n^2) ... there's no reason to do that

abgeorge7: Linting gives the error `Assignment to property of function parameter 'lookup'  no-param-reassign` so I am trying to eliminate that, I can disable this for `.reduce` though?

src/CoreMetadata/page.jsx
abgeorge7: if there are conflicting styles I can try to standardize it

package.json
abgeorge7: Unfortunately that doesn't get rid of the warning :( 
https://github.com/facebook/jest/issues/4545#issuecomment-332762365

src/CoreMetadata/page.jsx
abgeorge7: Is there one we want? I can set that in ESLint so they are the same

src/DataDictionary/DataDictionary.jsx
frickjack: Yeah - that would be great if you can eliminate that rule for .reduce :+1:

src/CoreMetadata/page.jsx
abgeorge7: Standardized it to single - so there are a lot of changed files now

PR #331, opened 2018-08-07
PXD-1430 fix(core-metadata)

paulineribeyre:
Instead of a broken link, display an error on the landing page if there's no metadata


paulineribeyre:
Yes I'm working on a PR with a test
The error displayed is just text, something like "Error: object_id 1234 not found"

PR #332, opened 2018-08-07
PXD-1325 Remove beagle.css (Arranger stylesheet)

abgeorge7:
Importing the Arranger stylesheet led to some console errors so instead we decided to grab the styles as needed from the stylesheet and have them locally, since we are also overriding most of them anyway.


qingyashu:
Checked~Style looks great!

PR #335, opened 2018-08-07
PXD-1491 Link to new Explorer page in dev

abgeorge7:
The "Exploration" button will go to the new data explorer when in the default/Generic Data Commons app. Also fixed a quick bug I noticed with outlines in the nav bar.
@thanh-nguyen-dang just wanted to make sure that only we use the default/Generic Data Commons config in parameters.js right? This won't be used by anyone external?


abgeorge7:
Not all of the apps have "buttons" in their config, but they all have "navigation." I also set the feature flag to be false for the explorer page so I think that the page will remain hidden. Want to double check with @thanh-nguyen-dang though

PR #336, opened 2018-08-08
PXD-1262 Fix VA dictionary graph crash

abgeorge7:
Checks nodes in first pass to make sure they aren't visited more than once (--> loop).


zflamig:
Can you verify the graph doesn't change (on a working commons) as a result of this? I think this will undo pushing nodes to the farthest that they are linked from in the tree.

PR #338, opened 2018-08-09
PXD-1368 fix(localconf): vis colors in localconf.js

qingyashu:
Change visualization colors in localconf.js: since only charts are using those colors, let helper access the colors, and remove all `localTheme` props from previous chart-related components.


src/components/charts/IndexBarChart/index.jsx
```diff
@@ -125,8 +120,8 @@ class IndexBarChart extends React.Component {
           >
             <h4>Project Submission status</h4>
             <XAxis
-              stroke={localTheme['barGraph.lineColor']}
-              fill={localTheme['barGraph.lineColor']}
+              stroke='#666666'
+              fill='#666666'
```
abgeorge7: is there a way to make these configurable, like pass a prop to the chart?

src/localconf.js
```diff
@@ -48,27 +48,24 @@ function buildConfig(opts) {
...
-    'summary.borderTopColor': '#c87152',
-    'summary.borderColor': '#222222',
-    'summary.countColor': '#ff4200',
-    'summary.iconColor': '#008000',
-    'projectTable.summaryRowColor': '#eeeeee',
-    'projectTable.submitButtonColor': '#dddddd',
-    'pieChartTwoColor.pie1Color': '#3283c8',
-    'pieChartTwoColor.pie2Color': '#e7e7e7',
-    tableBarColor: '#7d7474',
+  const visColor = {
```
abgeorge7: Can we rename `visColor`, `categoryColor9`, and `categoryColor2`? I feel like they could be more descriptive like `colorsForCharts` or something

src/components/charts/IndexBarChart/index.jsx
qingyashu: I think for now we only have 1 style (with white background), it is not necessary to make text colors configurable...

src/localconf.js
qingyashu: Sure~ How about `visColor`->`colorsForCharts`, `categoryColor9` -> `category9Colors`, `categoryColor2` -> `category2Colors`?

src/components/charts/IndexBarChart/index.jsx
abgeorge7: true, but if we want to use this chart anywhere else in the future we'd want it to be a prop not hardcoded

src/localconf.js
abgeorge7: `colorsForCharts` is good, but I think the other two are still not clear in terms of what they contain, can they be something like `barChartColors` and `pieChartColors`

qingyashu:
Committed by mistake, sorry about that, just remove the useless file~please check again~

PR #355, opened 2018-09-10
Updated brain and DCP config, Safari bug fix

abgeorge7:
- Chart configuration for Brain Commons and DCP
- Pulled in Arranger's api file to edit request - need to include `credentials` in request for Safari to work with Arborist
- Some style fixes for Arranger's range slider


src/arrangerApi.js
```diff
@@ -0,0 +1,41 @@
...
+const api = ({ endpoint = '', body, headers }) => fetch(urlJoin(ARRANGER_API, endpoint), {
+  method: 'POST',
+  headers: { ...alwaysSendHeaders, ...headers },
+  body: JSON.stringify(body),
+  credentials: 'include',
+}).then(r => r.json());
+
+export const graphql = body => api({ endpoint: 'graphql', body });
+
+export const fetchExtendedMapping = ({ graphqlField, projectId }) =>
```
qingyashu: I think we don't need to copy `fetchExtendedMapping`, `addHeaders` and `getAlwaysAddHeaders`here if we don't override them.

PR #360, opened 2018-09-13
PXD-1839 fix(overlap): fix chart text overlap bug

qingyashu:
Rewrite legend in div tags instead of SVG lements, in order to apply CSS


src/components/charts/SummaryPieChart/SummaryPieChart.less
```diff
@@ -20,7 +20,7 @@
 }
 
 .summary-pie-chart__legend {
-  width: 70px;
+  width: 40%;
```
abgeorge7: 🎉

PR #372, opened 2018-09-24
chore(bdbag): add export to saturn button

qingyashu:
WARN: This feature is for dcp only and depends Peregrine `feat/bagit` branch!


src/DataExplorer/custom/bdbag.js
```diff
@@ -0,0 +1,94 @@
+import { fetchWithCreds } from '../../actions';
+import { constructGraphQLQuery, hasKeyChain, MSG_FAILED_DOWNLOAD } from '../utils.js';
+
+export const getBDBagQuery = referenceIDList => `{
```
abgeorge7: this is only supposed to work with DCP right?

qingyashu: yes 😿

PR #381, opened 2018-09-27
PXD-1869 fix(dropdown): import dropdown button

qingyashu:
Usage: in `parameters.js`, enable more than one buttons pointing at same `dropdownId`

### DIFF
--- a/src/components/tables/DataExplorerTable/DataExplorerTable.less
+++ b/src/components/tables/DataExplorerTable/DataExplorerTable.less
@@ -41,13 +41,13 @@
   }
 
   .tableToolbar .group > :first-child {
-    border-top-left-radius: 10px;
-    border-bottom-left-radius: 10px;
+    border-top-left-radius: 4px;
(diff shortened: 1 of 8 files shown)


qingyashu:
> when should I see a dropdown button? not seeing one in qa-dcp.
- It is because only one enabled button is referring to a dropdown, so no necessary to group it into a dropdown. Try switch on the manifest button and you will see a dropdown~

src/components/tables/DataExplorerTable/DataExplorerTable.less
```diff
@@ -212,5 +212,14 @@
...
+
 }
 /* stylelint-enable */
+
+.data-explorer-table__arranger-table .tableToolbar .group .g3-dropdown {
+  border: none;
+  padding: 0;
+  margin: 0 10px;
+  display: inline-block;
+  z-index: 100000000;
```
abgeorge7: 😂

PR #384, opened 2018-09-28
fix(embed): embed notebook into windmill

qingyashu:


qingyashu:
@zflamig  @abgeorge7  I change EDC's workspace link for test, we might also want to remove Jupyter's `logout` button (on top-right) before we enable this feature.

PR #394, opened 2018-10-04
KF phenotype params

abgeorge7:


philloooo:
it's kfDcfInterop not kf

abgeorge7:
@philloooo Ahh thanks. Question though - qa-kidsfirst looks like it uses `kf`, should we have config in both `kf` and `kfDcfInterop` ?

philloooo:
I think qa uses the config in gitops? https://github.com/uc-cdis/gitops-qa/tree/master/qa-kidsfirst.planx-pla.net/portal

PR #395, opened 2018-10-08
fix(missing): change __missing__ to "No Data"

qingyashu:


src/components/charts/helper.js
```diff
@@ -123,6 +131,7 @@ const getCharts = (data, arrangerConfig, sqon) => {
       }
     });
   }
+  console.log({ summaries, countItems, stackedBarCharts });
```
abgeorge7: console log

PR #396, opened 2018-10-09
fix(qgl-editor): fix annoying overflow for gqleditor

qingyashu:
Make gqleditor same height as content div.


src/GraphQLEditor/GqlEditor.less
```diff
@@ -1,10 +1,13 @@
 .gql-editor {
-  height: 90vh;
+  height: calc(~"100vh - 181px");
```
abgeorge7: where is the 181px coming from? just curious :P

PR #405, opened 2018-10-17
PXD-1859 Manage explorer session

abgeorge7:
Hits Fence endpoint every 10 min if user has been active in the last 30 min to refresh Fence token and prevent empty data explorer

### DIFF
--- a/src/SessionMonitor/index.js
+++ b/src/SessionMonitor/index.js
@@ -12,6 +12,9 @@ class SessionMonitor {
   }
 
   start() {
+    if (this.interval) { // interval already started
+      return;
+    }
     window.addEventListener('mousedown', () => this.updateUserActivity(), false);
     window.addEventListener('keypress', () => this.updateUserActivity(), false);
     this.interval = setInterval(
@@ -21,9 +24,11 @@ class SessionMonitor {
   }
 
   stop() {
-    clearInterval(this.interval);
-    window.removeEventListener('mousedown', this.updateUserActivity(), false);
-    window.removeEventListener('keypress', this.updateUserActivity(), false);
+    if (this.interval) {
+      clearInterval(this.interval);
+      window.removeEventListener('mousedown', this.updateUserActivity(), false);
+      window.removeEventListener('keypress', this.updateUserActivity(), false);
+    }
   }
 
   updateUserActivity() {
@@ -53,4 +58,8 @@ class SessionMonitor {
   }
 }
 
-export default SessionMonitor;
+const singleton = new SessionMonitor();
+module.exports = {
+  singleton,
+  SessionMonitor,
+};
--- a/src/SessionMonitor/index.js
+++ b/src/SessionMonitor/index.js
@@ -3,7 +3,7 @@ import getReduxStore from '../reduxStore';
 import { logoutAPI } from '../actions';
 
 /* eslint-disable class-methods-use-this */
-class SessionMonitor {
+export class SessionMonitor {
   constructor(updateSessionTime, inactiveTimeLimit) {
     this.updateSessionTime = updateSessionTime || 10 * 60 * 1000;
     this.inactiveTimeLimit = inactiveTimeLimit || 30 * 60 * 1000;
@@ -59,7 +59,4 @@ class SessionMonitor {
 }
 
 const singleton = new SessionMonitor();
-module.exports = {
-  singleton,
-  SessionMonitor,
-};
+export default singleton;
--- a/src/SessionMonitor/index.jsx
+++ b/src/SessionMonitor/index.jsx
@@ -1,0 +1,51 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { loginPath, userapiPath } from '../localconf';
+import getReduxStore from '../reduxStore';
+import { logoutAPI } from '../actions';
+
+class SessionMonitor extends React.Component {
+  constructor(props) {
+    super(props);
+    this.state = { mostRecentActivityTimestamp: Date.now() };
+  }
+
+  componentDidMount() {
+    window.addEventListener('mousedown', this.updateUserActivity, false);
+    window.addEventListener('keypress', this.updateUserActivity, false);
+    setTimeout(this.refreshSession, this.props.refreshSessionTime); // check session every X min
+  }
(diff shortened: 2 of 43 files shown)


src/SessionMonitor/index.jsx
```diff
@@ -0,0 +1,52 @@
...
+class SessionMonitor extends React.Component {
+  constructor(props) {
+    super(props);
+    this.state = { mostRecentActivityTimestamp: Date.now() };
+  }
+
+  componentDidMount() {
+    window.addEventListener('mousedown', this.updateUserActivity, false);
+    window.addEventListener('keypress', this.updateUserActivity, false);
+    setTimeout(this.refreshSession, this.props.refreshSessionTime); // check session every X min
```
frickjack: this should be setInterval

src/SessionMonitor/index.jsx
```diff
@@ -0,0 +1,52 @@
...
+    this.setState({ mostRecentActivityTimestamp: Date.now() });
+  }
+
+  refreshSession = () => {
+    // If user has been inactive for Y min
+    if (Date.now() - this.state.mostRecentActivityTimestamp > this.props.inactiveTimeLimit) {
+      getReduxStore().then((store) => {
+        store.dispatch(logoutAPI());
+      });
+    } else {
```
frickjack: it would be good to have a guard here:
```
else if ( I haven't refreshed the session recently ) { ...
```

src/SessionMonitor/index.jsx
```diff
@@ -0,0 +1,52 @@
...
+import { logoutAPI } from '../actions';
+
+class SessionMonitor extends React.Component {
+  constructor(props) {
+    super(props);
+    this.state = { mostRecentActivityTimestamp: Date.now() };
+  }
+
+  componentDidMount() {
+    window.addEventListener('mousedown', this.updateUserActivity, false);
```
frickjack: If we're making a component out of this, then we should have logic to remove the window listeners and cancel the setInterval when the component unmounts, and also a guard against having multiple copies of this component on the page.  This thing is really a utility - not a UX component: `sessionMonitor.start()`, `sessionMonitor.end()` - but I get that people love to make a component out of everything :-p

src/SessionMonitor/index.jsx
```diff
@@ -0,0 +1,52 @@
...
+    if (Date.now() - this.state.mostRecentActivityTimestamp > this.props.inactiveTimeLimit) {
+      getReduxStore().then((store) => {
+        store.dispatch(logoutAPI());
+      });
+    } else {
+      fetch(userapiPath); // hitting Fence endpoint refreshes token
+      setTimeout(this.refreshSession, this.props.refreshSessionTime);
+    }
+  }
+
```
frickjack: Ok - calling `setTimeout` multiple times accomplishes the same thing as `setInterval` if you want to do it that way - I expected `sessionMontor.refreshSession()` to be this thing that just does a `fetch` rather than part of a state machine, but either way works :-p

src/SessionMonitor/index.jsx
```diff
@@ -0,0 +1,56 @@
+import { userapiPath } from '../localconf';
```
frickjack: Rename this file `index.js` ...

src/SessionMonitor/index.jsx
```diff
@@ -0,0 +1,56 @@
...
+class SessionMonitor {
+  constructor(updateSessionTime, inactiveTimeLimit) {
+    this.updateSessionTime = updateSessionTime || 10 * 60 * 1000;
+    this.inactiveTimeLimit = inactiveTimeLimit || 30 * 60 * 1000;
+    this.mostRecentActivityTimestamp = Date.now();
+    this.interval = null;
+  }
+
+  start() {
+    window.addEventListener('mousedown', () => this.updateUserActivity(), false);
```
frickjack: Add a guard something like - 
```
if (this.interval) { return; } // already started
```

src/SessionMonitor/index.jsx
```diff
@@ -0,0 +1,56 @@
...
+    window.addEventListener('keypress', () => this.updateUserActivity(), false);
+    this.interval = setInterval(
+      () => this.updateSession(),
+      this.updateSessionTime,
+    ); // check session every X min
+  }
+
+  stop() {
+    clearInterval(this.interval);
+    window.removeEventListener('mousedown', this.updateUserActivity(), false);
```
frickjack: Similar to start - check if it's not already stopped: 
```
if (this.interval) { ...
this.interval = null
} // else not started or already stopped ...
```

src/SessionMonitor/index.jsx
```diff
@@ -0,0 +1,56 @@
...
+  }
+
+  logoutUser() {
+    getReduxStore().then((store) => {
+      store.dispatch(logoutAPI());
+    });
+  }
+}
+
+export default SessionMonitor;
```
frickjack: Since we only ever want one SessionManager running - you can implement a `singleton` pattern here:
   https://en.wikipedia.org/wiki/Singleton_pattern
```
const singleton = new SessionMonitor();
export default singleton;
```
then the client does something like:
```
import sessionMonitor from ...;
sessionMonitor.start();
```

src/SessionMonitor/index.js
```diff
@@ -0,0 +1,65 @@
...
+  logoutUser() {
+    getReduxStore().then((store) => {
+      store.dispatch(logoutAPI());
+    });
+  }
+}
+
+const singleton = new SessionMonitor();
+module.exports = {
+  singleton,
```
frickjack: Hey - this is sort of mixing nodejs module style (`module.exports`) with es2015 module.  I think you probably want something like either:
```
export class SessionMonitor {
}
...
export default singleton;
-----------------
import sessionMonitor, { SessionMonitor } from '../SessionMonitor';
```
or
```
export default { singleton, SessionMonitor };
...
import { sessionMonitor, SessionMonitor } from '../SessionMonitor';
```
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/import
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/export

PR #411, opened 2018-10-29
PXD-1703: user agreement

thanh-nguyen-dang:
### New Features
- Implemented User Agreement page

### Deployment changes
- JSON config file.
  * We have to define questions for each common. The default values in `default.json` in the path `components/certs`
  * We need to define the required certificates for every common in `requiredCerts` (check the `default.json`)


src/UserAgreement/Quiz.less
```diff
@@ -0,0 +1,84 @@
...
+}
+
+.quiz__menu {
+  display: inline-block;
+  float: right;
+  text-align: right;
+  width: 50%;
+}
+
+.quiz__menu-bullet {
```
qingyashu: The text here is not precisely centered aligned, how about following changes: 
1. Change `.quiz__menu-bullet`'s display style to  `display: inline-block;`, remove its all paddings
2. For `.quiz__menu-bullet.h3`, add `text-align: center;` 
It will fix this :)

src/UserAgreement/Option.less
```diff
@@ -0,0 +1,81 @@
...
+.option__bullet {
+  display: table-cell;
+  width: 34px;
+  height: 34px;
+  -webkit-border-radius: 17px;
+  -moz-border-radius: 17px;
+  border-radius: 17px;
+  padding: 5px 0 0 12px;
+  background-color: var(--g3-color__lightgray);
+  &:hover {
```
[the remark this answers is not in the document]
thanh-nguyen-dang: nope, they are different. They `hover` later is for the .option_buttlet not the frame itself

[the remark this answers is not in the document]
qingyashu: Yeah I knew they are for different focus target but we only need hover status for frame (thought it was designed in SPEC: https://uchicago.app.box.com/file/317984245441

src/UserAgreement/Quiz.jsx
```diff
@@ -0,0 +1,200 @@
...
+  }
+
+  render() {
+    const { questionList, title } = this.props;
+    return (
+      <div className='quiz__form'>
+        <div className='quiz__title'>
+          <div className='quiz__icon'>
+            {
+              <IconComponent
```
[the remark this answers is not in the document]
thanh-nguyen-dang: yeah, fixed it

src/UserAgreement/Option.less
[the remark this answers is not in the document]
thanh-nguyen-dang: it will be nicer if we have the bullet are hovered together with the frame

src/UserAgreement/Option.jsx
```diff
@@ -0,0 +1,50 @@
...
+      frameClassModifier = this.props.isCorrectAnswer ? 'option__frame--correct' : 'option__frame--incorrect';
+      bulletClassModifier = this.props.isCorrectAnswer ? 'option__bullet--correct' : 'option__bullet--incorrect';
+    }
+    return (
+      <div
+        className={`option__frame ${frameClassModifier}`}
+        role='button'
+        onClick={() => this.props.onChange(this.props.idx)}
+        onKeyDown={this.handleKeyDown}
+        tabIndex={-10}
```
abgeorge7: Is this just a placeholder value? is it possible to use a real `tabIndex` to make it accessible? :)

src/UserAgreement/Quiz.jsx
```diff
@@ -0,0 +1,192 @@
...
+    this.setState({ currentSection: getQuestionSectionId(sectionIdx + 1) });
+  }
+
+  resetState() {
+    this.setState(makeDefaultState(this.props.questionList));
+  }
+
+  answerQuestion(questionId, isCorrect) {
+    const notDone = this.state.notDone.filter(item => item !== questionId);
+    let [rights] = this.state.rights;
```
abgeorge7: is this syntax correct?  I think it can just be `let rights...`

src/UserAgreement/Quiz.less
```diff
@@ -0,0 +1,88 @@
+@import '~@gen3/ui-component/dist/css/base.less';
+
+.quiz__form {
+  display: block;
+  width: 1025px;
```
abgeorge7: The quiz looks off center in my browser, I think we could make it's width responsive to fix this (ie. `100%`)

src/UserAgreement/Quiz.test.jsx
```diff
@@ -0,0 +1,68 @@
...
+      (res, item) => (res + item.options.length), 0,
+    );
+    expect($vdom.find('.quiz__questions')).toHaveLength(1);
+    expect($vdom.find('.question__content')).toHaveLength(quiz.questions.length);
+    expect($vdom.find('.question__name.h4')).toHaveLength(quiz.questions.length);
+    expect($vdom.find('.option__bullet')).toHaveLength(optionLength);
+    expect($vdom.find('.quiz__menu-bullet')).toHaveLength(quiz.questions.length);
+    expect($vdom.find('.quiz__menu-bullet--active')).toHaveLength(1);
+  });
+});
```
abgeorge7: I think it'd be good to have some more tests, like for the `handleScroll` code, `handleSubmit`, etc.

src/UserAgreement/QuizRight.jsx
```diff
@@ -0,0 +1,38 @@
...
+  static defaultProps = {
+  };
+
+  render() {
+    const finishedAnswer = (this.props.notDone.length + this.props.wrongs.length === 0);
+    const btnCls = finishedAnswer ? 'g3-button--primary' : 'g3-button--secondary';
+    return (
+      <div className='quiz__right'>
+        <div className='quiz__right--box'>
+          <QuizStatus title='Not Answer Yet:' questions={this.props.notDone} />
```
abgeorge7: "Not Answered Yet"

src/UserAgreement/QuizRight.jsx
```diff
@@ -0,0 +1,38 @@
+import React, { Component } from 'react';
+import PropTypes from 'prop-types';
+import QuizStatus from './QuizStatus';
+import './QuizRight.less';
+
+/**
+ * Little quiz component - roperites: questionList, title, onSubmit
+ */
+class QuizRight extends Component {
```
abgeorge7: This name doesn't make total sense to me, maybe `QuizSummary` ?

src/UserAgreement/QuizRight.less
```diff
@@ -0,0 +1,14 @@
+.quiz__right {
+  display: inline-block;
+  text-align: right;
+  position: relative;
+  padding-right: 30px;
+  width: 285px;
```
abgeorge7: would be good to have a percentage here as well

src/UserAgreement/QuizRight.jsx
```diff
@@ -0,0 +1,38 @@
+import React, { Component } from 'react';
+import PropTypes from 'prop-types';
+import QuizStatus from './QuizStatus';
+import './QuizRight.less';
+
+/**
+ * Little quiz component - roperites: questionList, title, onSubmit
```
abgeorge7: I think these properties might need to be updated? "notDone, wrongs, onSubmit"

src/UserAgreement/QuizStatus.jsx
```diff
@@ -0,0 +1,35 @@
+import React, { Component } from 'react';
+import PropTypes from 'prop-types';
+import './QuizStatus.less';
+
+/**
+ * Little quiz component - roperites: questionList, title, onSubmit
```
abgeorge7: I think these properties need to be updated too

src/UserAgreement/Quiz.jsx
thanh-nguyen-dang: that is the requirement of our Codacy
https://eslint.org/docs/rules/prefer-destructuring

src/UserAgreement/QuizRight.less
thanh-nguyen-dang: I think the right panel's width should be fixed while the left side could be a percentage. But, we should discuss with Vivi on this

src/UserAgreement/Quiz.less
thanh-nguyen-dang: will check with Vivi, because the design comes from her :)

src/UserAgreement/Option.jsx
thanh-nguyen-dang: codacy suggests a negative value for the button

philloooo:
if you have Nothing in a certain section of your pr description, just leave it empty instead of add "Nothing" - easier for us to compile to release note later

src/UserAgreement/Option.jsx
```diff
@@ -0,0 +1,53 @@
...
+    if (this.props.selected) {
+      frameClassModifier = this.props.isCorrectAnswer ? 'option__frame--correct' : 'option__frame--incorrect';
+      bulletClassModifier = this.props.isCorrectAnswer ? 'option__bullet--correct' : 'option__bullet--incorrect';
+    }
+    return (
+      <div
+        className={`option__frame ${frameClassModifier}`}
+        role='button'
+        onClick={() => this.props.onChange(this.props.idx)}
+        onKeyDown={this.handleKeyDown}
```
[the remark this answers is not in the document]
thanh-nguyen-dang: Codacy stuff. That's built-in function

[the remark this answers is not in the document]
thanh-nguyen-dang: for the Questions stuffs, the design coming from Vivi @abgeorge7

ZakirG:
Some thoughts: 

- After the user logs in and is redirected to the quiz, I think it would be helpful if there was a message at the top of the quiz that informed the user why they were taken to this page. Something like, "In order to gain access to the data portal, you must correctly answer the following questions."

- It feels redundant that the questions are numbered and also have titles that are spellings of their number -- "The first question". Either omit this "name" property, or provide better example quizzes in README & default.json that demonstrate the purpose of this property. 

- On a small screen, the user needs to scroll to the bottom of the page to access the submit button, which defeats the purpose of placing the submit button outside of the quiz form. I would either place the submit button at the top of the page in the right column so that the user does not have to scroll down the page twice, or place the submit button underneath all the questions to create a conventional single-column form with no internal scrollbars. This would not be an issue if the website were responsive, but it isn't :( 

- I would prefer if when questions are incomplete/incorrect and the user tries to submit the form, there is an error message near the question on the form describing the problem ("This question is required." / "Your answer was incorrect."). In general, error messages should be placed as close as possible to the location where user action is needed to fix the issue, or else form completion requires too much thinking. And since you've already written handleScroll code, you might as well automatically scroll to the first incomplete/incorrect question when the user clicks submit.

thanh-nguyen-dang:
> Some thoughts:
> 
> * After the user logs in and is redirected to the quiz, I think it would be helpful if there was a message at the top of the quiz that informed the user why they were taken to this page. Something like, "In order to gain access to the data portal, you must correctly answer the following questions."
> * It feels redundant that the questions are numbered and also have titles that are spellings of their number -- "The first question". Either omit this "name" property, or provide better example quizzes in README & default.json that demonstrate the purpose of this property.
> * On a small screen, the user needs to scroll to the bottom of the page to access the submit button, which defeats the purpose of placing the submit button outside of the quiz form. I would either place the submit button at the top of the page in the right column so that the user does not have to scroll down the page twice, or place the submit button underneath all the questions to create a conventional single-column form with no internal scrollbars. This would not be an issue if the website were responsive, but it isn't :(
> * I would prefer if when questions are incomplete/incorrect and the user tries to submit the form, there is an error message near the question on the form describing the problem ("This question is required." / "Your answer was incorrect."). In general, error messages should be placed as close as possible to the location where user action is needed to fix the issue, or else form completion requires too much thinking. And since you've already written handleScroll code, you might as well automatically scroll to the first incomplete/incorrect question when the user clicks submit.

@ZakirG All the things that you suggested need a redesign from @uchicagovivi . We will not make it in this issue. Let's do it in another one

ZakirG:
@thanh-nguyen-dang Can you attach Vivi's design wireframe in your Jira task and link to it here?

thanh-nguyen-dang:
> Some thoughts:
> 
> * After the user logs in and is redirected to the quiz, I think it would be helpful if there was a message at the top of the quiz that informed the user why they were taken to this page. Something like, "In order to gain access to the data portal, you must correctly answer the following questions."
> * It feels redundant that the questions are numbered and also have titles that are spellings of their number -- "The first question". Either omit this "name" property, or provide better example quizzes in README & default.json that demonstrate the purpose of this property.
> * On a small screen, the user needs to scroll to the bottom of the page to access the submit button, which defeats the purpose of placing the submit button outside of the quiz form. I would either place the submit button at the top of the page in the right column so that the user does not have to scroll down the page twice, or place the submit button underneath all the questions to create a conventional single-column form with no internal scrollbars. This would not be an issue if the website were responsive, but it isn't :(
> * I would prefer if when questions are incomplete/incorrect and the user tries to submit the form, there is an error message near the question on the form describing the problem ("This question is required." / "Your answer was incorrect."). In general, error messages should be placed as close as possible to the location where user action is needed to fix the issue, or else form completion requires too much thinking. And since you've already written handleScroll code, you might as well automatically scroll to the first incomplete/incorrect question when the user clicks submit.

In more detail:
- About the name of question, the question have its own name (made sense name), I just put `The first question` because don't wanna think specific name, but the people who create that question should have that idea. It is not too difficult for ones who create the question can think of that. I don't think it is necessary to put a direction. But the name of question is necessary.
- For fourth recommendation, we only enable the submit button when user answer all questions correctly.
- For other things, I don't have strong opinion on should or shouldn't we do that. But redo that need a redesign. Suggest to do in another issue @ZakirG

thanh-nguyen-dang:
> @thanh-nguyen-dang Can you attach Vivi's design wireframe in your Jira task and link to it here?

https://jira.opensciencedatacloud.org/browse/PXD-1703 @ZakirG

thanh-nguyen-dang:
> @thanh-nguyen-dang Can you attach Vivi's design wireframe in your Jira task and link to it here?

Here are the design from Vivi. One thing different is we decided to not do the popup @ZakirG :
https://uchicago.app.box.com/folder/53415902542
https://uchicago.app.box.com/file/317984245441 -> spec

qingyashu:
Just updated some scroll bar related css code so that it won't affect the whole website. All looks good to me, just I am also having same problem as Abby mentioned before: after finishing quiz it is not redirected to homepage and I am still in quiz page. This does not happens every time, but when it happens I am seeing some console error.

src/UserAgreement/ReduxCertPopup.js
```diff
@@ -0,0 +1,48 @@
...
+  );
+
+const mapStateToProps = state => ({
+  certsList: certs,
+  pendingCerts: minus(requiredCerts, state.user.certificates_uploaded),
+});
+
+const mapDispatchToProps = (dispatch, ownProps) => ({
+  onSubmit: (data, questionList, quiz) => dispatch(
+    submitForm(data, questionList, quiz, ownProps.history),
```
qingyashu: Is `submitForm` just taking 3 arguments?

src/UserAgreement/Quiz.jsx
```diff
@@ -0,0 +1,193 @@
...
+  return {
+    notDone,
+    rights: [],
+    wrongs: [],
+    currentSection: getQuestionSectionId(1),
+  };
+};
+
+/**
+ * Little quiz component - roperites: questionList, title, onSubmit
```
qingyashu: add `description` to comments?  or just remove the comments, I think proTypes is already self-explained enough.

src/UserAgreement/Quiz.jsx
```diff
@@ -0,0 +1,193 @@
...
+    }
+    const curPos = ctrl.scrollTop + ctrl.offsetTop;
+    // Get id of current scroll item
+    let { lastOffsetTop } = this;
+    let firstActiveIdx = this.state.currentSection;
+    Object.values(this.qs).forEach((item) => {
+      const dom = this.qsDom[getQuestionSectionId(item.props.idx + 1)];
+      if (dom.offsetTop >= curPos) {
+        if (dom.offsetTop <= lastOffsetTop &&
+          curPos + ctrl.offsetHeight >= dom.offsetTop + (dom.offsetHeight * 0.9)) {
```
qingyashu: This `handleScroll` function is difficult for me to understand, I think it'd be great to add more comments / explanations so that others could easily understand its logic, especially this `if` statement (e.g. what does `0.9` mean). Also could we have better name for `ctrl`? for example `scrollController`...

src/UserAgreement/Question.less
```diff
@@ -20,14 +22,27 @@
...
+
+.question__content {
+  margin: 30px 0 30px 0;
+}
+
+.question__name {
+  margin: 30px 0 0 0;
+}
+
+.question__name.h4 {
```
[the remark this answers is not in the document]
thanh-nguyen-dang: nope, we need to change color of h4

PR #420, opened 2018-11-14
PXD-1976 feat/dd graph

qingyashu:
- Branch deployed in https://qingya.planx-pla.net/DD, try it! 
- Design sheets: https://uchicago.app.box.com/folder/56881020248
- JIRA ticket link: https://jira.opensciencedatacloud.org/browse/PXD-1976

### New Features
- Implemented new graph and table view for data dictionary


ZakirG:
Looks so good!! Works very well. There was one moment where it kind of froze and I got a console error `Error: <g> attribute transform: Expected number, "…le(0) translate(NaN, NaN) ".` But I wasn't able to reproduce it :) 

Some thoughts:

- I think it would be helpful if when you click on "See it on graph", the canvas is reset (whatever the canvas reset button does), so that the node is guaranteed to be in the viewport. In the current implementation, if the graph view doesn't include the node at the time that the See it on graph button is clicked, the user won't see that node in the graph.

- I feel like the text inside the graph-drawer__node rectangles in the Graph View should not be selectable, since the nodes are buttons. Same for the text inside the Graph View and Table View tabs. 

- Can you make the x button (node-popup__close) inside the node popups bigger? I have some trouble clicking it.

- In the open node layout (after you click Open properties) the table has a slight border, which doesn't match the design. You can get rid of it with  .data-dictionary-property-table__data:last-child and :first-child selectors. 


Thoughts that might be applicable to a future PR, if there's a design discussion:

- It seems like most of the pieces in the graph view have border-radius of 0. I feel like this is inconsistent with the rest of the site -- on the homepage, all the panels and buttons are gently rounded with a small border-radius. 

- I would prefer a smaller border-radius on the Graph View / Table View selector, it feels very round compared to the site's norm.

philloooo:
just comment about the pr description: let's put these miscellaneous info at the top? the new feature section should just have one new feature that's the new graph page

qingyashu:
Thanks for the helpful comments! 😍   I have fixed the 2 mentioned bugs and improved the styles according to suggestions (except some I will confirm with designers to find better solutions, like background color for overlay table). Please have a check and let me know if there's other bugs I missed.

ZakirG:
Looks perfect!

src/DataDictionary/table/DataDictionaryTable/index.js
```diff
@@ -0,0 +1,21 @@
+import { connect } from 'react-redux';
+import localconf from '../../../localconf';
```
abgeorge7: It might be better to just import `appname` since thats all thats being used from here

src/GraphUtils/utils.js
```diff
@@ -1,3 +1,28 @@
+const getGraphToDotString = (nodes, edges, treeLevel2Names) => {
+  const whRatio = 1;
+  let graphString = 'digraph dictionary {\n';
+  // graphString += 'concentrate=true\n';
```
abgeorge7: extra comments

src/index.jsx
```diff
@@ -21,15 +21,15 @@ import ExplorerPage from './Explorer/ExplorerPage';
 import CoreMetadataPage from './CoreMetadata/page';
 import { fetchCoreMetadata } from './CoreMetadata/reduxer';
 import IndexPage from './Index/page';
-import DataDictionary from './DataDictionary/ReduxDataDictionary';
-import DataDictionaryNode from './DataDictionary/ReduxDataDictionaryNode';
+import DataDictionary from './DataDictionary/.';
+// import DataDictionaryNode from './DataDictionary/ReduxDataDictionaryNode';
```
abgeorge7: extra comment

src/DataDictionary/graph/GraphCalculator/GraphCalculator.jsx
```diff
@@ -0,0 +1,156 @@
...
+    const newHighlightingNode = nextProps.highlightingNode;
+    const newFurtherHighlightingNodeID = nextProps.furtherHighlightingNodeID;
+    if (this.oldHighlightingNode !== newHighlightingNode) {
+      const relatedHighlightedNodeIDs = calculateHighlightRelatedNodeIDs(
+        newHighlightingNode,
+        this.props.nodes,
+      );
+      this.props.onHighlightRelatedNodesCalculated(relatedHighlightedNodeIDs);
+      const furtherClickableNodeIDs = newHighlightingNode ? newHighlightingNode.outLinks : [];
+      this.props.onFurtherClickableNodeIDsCalculated(furtherClickableNodeIDs);
```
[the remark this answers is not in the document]
qingyashu: 😛  It means when a node (let's say node A) is clicked, after highlighting A and all related nodes we will also allow some of the highlighted nodes clickable (we allow all A's descendent nodes clickable, but this interaction spec might change if we change design).... How about I change the variable name to `secondStepClickableNodes`... sorry I know interactions here is very complicated couldn't think a better name....

src/DataDictionary/graph/GraphDrawer/GraphDrawer.jsx
```diff
@@ -0,0 +1,249 @@
...
+    return (
+      <g
+        className='graph-drawer'
+        transform={`scale(${fittingScale}) translate(${fittingTransX}, ${fittingTransY}) `}
+        ref={this.graphDomRef}
+      >
+        { // edges
+          this.props.edges.map((edge, i) => {
+            let edgeRelatedClassModifier = '';
+            if (this.props.highlightingNode) {
```
thanh-nguyen-dang: You should split this function to multiple ones to be more readable.
- draw edges (in some condition)
- draw nodes (in some condition)

zflamig:
Not a big thing, but the zoom buttons are a bit unfriendly. When we zoom out/zoom in can we make sure that what is centered on the screen stays centered? Right now it moves off to the right. I also think we could make it zoom in/out faster at maybe 2x the current rate per click?

src/DataDictionary/graph/GraphCalculator/graphCalculatorHelper.js
```diff
@@ -0,0 +1,520 @@
...
+    dictionary,
+    counts_search: countsSearch,
+    links_search: linksSearch,
+  }, true, []);
+  const dotString = createDotStrinByNodesEdges(nodes, edges);
+  const viz = new Viz({ Module, render });
+  return viz.renderJSONObject(dotString)
+    .then((renderedJSON) => {
+      // draw nodes
+      const renderedNodes = renderedJSON.objects
```
thanh-nguyen-dang: should we separate to sub-functions for this `renderNodes` and `renderEdges` as well?

src/DataDictionary/graph/GraphCalculator/graphCalculatorHelper.js
```diff
@@ -0,0 +1,520 @@
...
+export const getNodeIDsThatHaveNoInLinks = (
+  subgraphNodeIDs,
+  subgraphEdges,
+  wholeGraphNodes,
+) => {
+  const resultIDs = [];
+  subgraphNodeIDs.forEach((nodeID) => {
+    const node = wholeGraphNodes.find(n => n.id === nodeID);
+    const inLinks = node.inLinks
+      .filter(inNodeID => subgraphNodeIDs.includes(inNodeID))
```
thanh-nguyen-dang: this logic and the later one in `getNodeIDsThatHaveNoOutLinks` are very similar. Should we:
- create a filter function receive lists as parameters
- for filter that are different you can pass it as a parameter
- and give the filter functions meaningful names?

src/DataDictionary/graph/GraphCalculator/graphCalculatorHelper.js
```diff
@@ -0,0 +1,520 @@
...
+import { truncateLines, graphStyleConfig } from '../../utils';
+import { getCategoryColor } from '../../NodeCategories/helper';
+
+export const getAllTypes = (nodes) => {
+  const types = nodes.map(node => node.type);
+  const uniqueTypes = _.uniq(types);
+  return uniqueTypes;
+};
+
+export const getAllChildrenNodeIDs = (highlightingNode, nodes) => {
```
thanh-nguyen-dang: In general, could you put some comments in this file so that we can understand more about the logic that you use to compute the position of nodes and edges.
You can also name the condition and some specific calculating functions meaningful name to increase the readability.

qingyashu:
Hey~ thanks for feedbacks! I have update the code: 
1. refine zoom speed and zoom center, also add animation for zooming
2. refactoring some code according to @thanh-nguyen-dang 's comments. 

Please have a check again (https://qingya.planx-pla.net/dd) and see what bug I missed. Thanks~~

qingyashu:
Also I am having a list of style improvement feedbacks from our designer but I will make them to next PRs..(after search function)

PR #421, opened 2018-11-15
PXD-2125 Feat/make bhc non indexable

tshao87:

### Improvements
- inject robots meta tag for data.braincommons.org (bhc), addresses jira ticket #PXD-2125


zflamig:
Can we add this as a flag in the config json instead of using the app name? So if others want it we can turn it on easier. We should also consider doing this here vs in the revproxy... I think doing it in the revproxy is better because we need a way to inject code per commons there for things like redirects or robots.txt.

tshao87:
> Can we add this as a flag in the config json instead of using the app name? So if others want it we can turn it on easier. We should also consider doing this here vs in the revproxy... I think doing it in the revproxy is better because we need a way to inject code per commons there for things like redirects or robots.txt.

Yea that's a better approach. I can make it to read in an argument and we can supply it in manifest.json? I'm not sure how to let it read from local json files and may need a bit more time to investigate it. Not very familiar with JS :(

tshao87:
New section added to local config file (such as bhc.json) 
` "featureFlags": {
    "explorer": true,
    "noindex": true
  },`
if the `noindex` flag is set to true, the robots meta tag will be injected, if that flag is omitted or set to false, no change will happen.

The header tag injection is made possible by using the React Helmet package.

data/config/bhc.json
```diff
@@ -119,7 +119,8 @@
     }
   },
   "featureFlags": {
-    "explorer": true
+    "explorer": true,
+    "noindex": true
```
abgeorge7: This is super picky but can you camel case?

src/index.jsx
```diff
@@ -72,17 +75,23 @@ async function init() {
   library.add(faAngleUp, faAngleDown);
 
   render(
-    <div>
+    < div >
```
abgeorge7: something is off with the linting here, there shouldn't be spaces

src/index.jsx
```diff
@@ -72,17 +75,23 @@ async function init() {
...
-        <ThemeProvider theme={theme}>
-          <MuiThemeProvider>
-            <BrowserRouter basename={basename}>
-              <div>
-                { GA.init(gaTracking, dev, gaDebug) && <RouteTracker /> }
-                <ReduxTopBar />
-                <ReduxNavBar />
-                <div className='main-content'>
-                  <Switch>
+      <ThemeProvider theme={theme}>
```
abgeorge7: Also I think the indentation is off

package.json
```diff
@@ -11,6 +11,7 @@
     "@gen3/ui-component": "0.0.18",
     "@storybook/addon-actions": "^3.4.8",
     "@storybook/react": "^3.4.8",
+    "ajv": "^6.5.5",
```
abgeorge7: what is this package for?

tshao87: removed

src/index.jsx
tshao87: linting has been fixed

PR #424, opened 2018-11-28
PXD-2248 fix/sponsor image

qingyashu:
Allow adding sponsor image to footer via configuration json file. By default the footer only has CTDS and gen3 logo. 

### Improvements
- allow adding sponsor image to footer


qingyashu:
Thanks for the comments! I have updated the code to remove Footer from reduxer file. 
About the multiple image, one solution is adding cusotmized stylesheets common by common. I added a customized css file for our first data commons (niaid) which needs sponsor image, please have a check qa-niaid and see how it looks, gitops PR at: https://github.com/uc-cdis/gitops-qa/pull/40/

PR #426, opened 2018-12-03
PXD-2168 Add another GraphQL endpoint

abgeorge7:

Adds a way to switch between the Peregrine ("Graph") and Arranger ("Flat") endpoints, so users can query the data coming from the data explorer.

Didn't add a button to take the user from the explorer to the proper endpoint in GraphiQL, but set it up so it should be easy if we want to do it in the future, by adding functionality for `?endpoint={index}` in the url.

Example Arranger query:

```
{
  case {
    extended
  }
}
```

Example Peregrine query:
```
{
  project {
    project_id
  }
}
```

### New Features
- Button to switch between endpoints on the `/query` page
- Ability to switch between endpoints by using query params, ie `/query?endpoint=0`

### Improvements
- Users can query the arranger api

### Deployment changes


src/GraphQLEditor/GqlEditor.less
```diff
@@ -1,3 +1,5 @@
+@import '~@gen3/ui-component/dist/css/base.less';
```
qingyashu: Stop using `base.less` file and use `--g3-color__xxx` instead of `@xxx`? (and I promise I will remove this `base.less` file later some day 😂

src/GraphQLEditor/ReduxGqlEditor.js
```diff
@@ -1,11 +1,15 @@
 import { connect } from 'react-redux';
-
+import querystring from 'querystring';
```
qingyashu: I think it's ``` import queryString from 'query-string';```
And using `queryString` instead of `querystring` won't need to replace the prepend question mark just FYI

src/GraphQLEditor/GqlEditor.jsx
```diff
@@ -1,45 +1,95 @@
...
 const parameters = {};
 
-const GqlEditor = ({ schema }) => {
-  if (!schema) {
-    return <Spinner />; // loading
+class GqlEditor extends React.Component {
+  constructor(props) {
+    super(props);
+    this.state = {
+      selectedEndpointIndex: props.endpointIndex || 0,
```
qingyashu: What if not `0` or `1`? will it break if `endpoint=2`?

abgeorge7: Probably 🙈 will add a check for that

PR #427, opened 2018-12-04
PXD-2365 chunk tsv submission

mfshao:
This pr addresses JIRA ticket PXD-2365 https://jira.opensciencedatacloud.org/browse/PXD-2365 and has been deployed on https://mingfei.planx-pla.net

### New Features
- chuck tsv files into smaller pieces in windmill before submitting
- chunk size (in lines) can be specified in local config file, if not specified, default is 30 lines per req


src/Submission/ReduxProjectSubmission.js
```diff
@@ -36,24 +38,57 @@ const submitToServer = (fullProject, methodIn = 'PUT') => (dispatch, getState) =
...
+  let i = 0;
+  if (submission.file_type === 'text/tab-separated-values') {
+    const fileSplited = file.split(/\n/g);
+    let fileHeader = fileSplited.shift();
+    fileHeader += '\r';
+    if (fileSplited.length > lineLimit) {
+      i = lineLimit;
+      let fileChunk = fileHeader;
+      while (fileSplited.length > 0) {
+        fileChunk += fileSplited.shift();
```
abgeorge7: I think using `.shift()` might not be the most efficient way to iterate through the array:

> To remove the returned item without re-addressing the array and invalidating all references to it, shift() requires moving the entire array around; (https://stackoverflow.com/questions/6501160/why-is-pop-faster-than-shift)

Do you need to delete entries from `fileSplited`, or can you just iterate through the array without modifying it?

tshao87: Right! This has been resolved by only iterating through `fileSplited` array

src/Submission/ReduxProjectSubmission.js
```diff
@@ -36,24 +38,58 @@ const submitToServer = (fullProject, methodIn = 'PUT') => (dispatch, getState) =
...
+      let count = lineLimit;
+      let fileChunk = fileHeader;
+
+      for (let i = 1; i < fileSplited.length; i += 1) {
+        fileChunk += fileSplited[i];
+        fileChunk += '\r';
+        count -= 1;
+        if (count === 0) {
+          fileArray.push(fileChunk);
+          fileChunk = fileHeader;
```
abgeorge7: do you need to `+= '\r'` here?

tshao87: no `fileheader` already has `\r` at the end after line 47

zflamig:
I just tested this and it doesn't appear to work (I don't see it creating chunks)

tshao87:
> This doesn't behave correctly if using TSVs saved directly from Excel on Mac. The line endings need to be converted first for it to work, but lets just correct that in the code here.
> 
> Is it possible to add some UI elements too so that the user knows its submitting chunks and what chunk it is on? Also it overwrites the entire log output but it should probably just append.

I'll check the TSVs saved directly from Excel and fix that part.

Regarding to the submission result display part, I was thinking about that too. One way is that we could just append the display of each response on the page so there will be a list of green/red bars? Or do we want something like a progress bar?

zflamig:
I think just a counter text above the logs would be sufficient. "Submitting chunk 8/23..." something like that.

PlanXCyborg:
Jenkins Build 10 : time taken 1 min 4 sec

PlanXCyborg:
Jenkins Build 11 : time taken 35 min

mfshao:
Right now it has a simple counter message displaying. But for better user experience we should create another ticket to maybe redesign part of the UI.

PlanXCyborg:
Jenkins Build 12 : time taken 30 min

PlanXCyborg:
Jenkins Build 14 : time taken 1 hr 7 min

PlanXCyborg:
Jenkins Build 14 : time taken 24 min

mfshao:
This pr has been updated with the following changes:
1. chunks are dispatched sequentially
2. chunk numbers are displayed in order
3. response messages from each chunk are appended in the `Details` field for reference

This PR has been deployed and tested at https://mingfei.planx-pla.net and https://qa-brain.planx-pla.net 
Could @zflamig and @abgeorge7 review this PR again? Thanks!

PR #449, opened 2019-01-16
fix(nginx.conf): ignore . in folder names

frickjack:
pidgin has paths like `/file/bla.foo/id` - so the nginx rule that treats any path with `.` as a file and 404 if it doesn't exist needs update


PlanXCyborg:
Jenkins Build 1 : time taken 24 min

PR #456, opened 2019-01-24
fix(file-img): display the right file type image

paulineribeyre:
Display the right file type image in the core metadata landing page

### Bug Fixes
The default image was displayed all the time


PlanXCyborg:
Jenkins Build 1 : time taken 24 min

PR #458, opened 2019-01-25
fix(order): make generated files in front, and also order them by time

qingyashu:
### Improvements
make generated files in front, and also order them by time


PlanXCyborg:
Jenkins Build 1 : time taken 27 min

PR #460, opened 2019-01-31
fix(download)

paulineribeyre:
If we use `expires_in=10`, we get the following error:
Fail to reach AWS: Parameter validation failed: Invalid range for parameter DurationSeconds, value: 10, valid range: 900-inf

### Bug Fixes
- set `expires_in` to 900 when requesting a presigned URL for file download


PlanXCyborg:
Jenkins Build 1 : time taken 41 min

PlanXCyborg:
Jenkins Build 2 : time taken 1 hr 0 min

PR #463, opened 2019-02-05
fix(graph): filter unwanted edges and remove useless force simulation

qingyashu:
### Bug Fixes
- Some edges in data model graph shouldn't be rendered and will cause "Error: <path> attribute d: Expected number, "Mundefined,undefi…"."

### Improvements
 - Remove useless d3 force simulation related code to enhance render performance


qingyashu:
To check the bug, please see `QA-test` in qa-bpa env.

src/DataModelGraph/SvgGraph.jsx
```diff
@@ -21,10 +20,15 @@ const d3 = {
...
  */
-export function createSvgGraph(nodesIn, edges) {
-  assignNodePositions(nodesIn, edges);
+export function createSvgGraph(nodesIn, edgesIn) {
+  assignNodePositions(nodesIn, edgesIn);
   // some nodes may not be linked under the root, so filter them out ...
   const nodes = nodesIn.filter(nd => !!nd.position);
+
+  // filter out edges that shouldn't be rendered in this graph
+  const edges = edgesIn
```
abgeorge7: I think here you could do `.filter(e => nodes.find(n => n.id === e.source.id || n.id === e.target.id))` and save yourself O(n).

In general this is adding O(en) cost (e = # edges, n = # nodes)... is there a way to make that more efficient? why would an edge that is not in this graph be passed into the function at all?

qingyashu: > why would an edge that is not in this graph be passed into the function at all?
This is because windmill is receiving a graph with orphan nodes, and our current graph traversal logic is not robust so there're some nodes missing in results, causing some edges don't belong to result nodes, and it will cause thousands of console error in browser :(   
So this filtering logic is just an error proof to avoid thousands of errors in browser.

qingyashu: So I think a better solution would be I digging deep into graph traversal logic to fix missing nodes. 
And I agree with you that we'll improve UI for this page, I don't think we have ticket for that yet, but it'd be great that we have careful discussion and design first.

PR #468, opened 2019-02-13
fix(subgouplink): recursively get subgoup links

qingyashu:
### Bug Fixes
Recursively get subgroup links when building graph


src/GraphUtils/utils.js
```diff
@@ -1,3 +1,29 @@
...
+  let subgroupLinks = [];
+  if (link.subgroup) {
+    link.subgroup.forEach((sgLink) => {
+      if (sgLink.subgroup) {
+        subgroupLinks = subgroupLinks.concat(getSubgroupLinks(sgLink, nameToNode, sourceId));
+      } else {
+        subgroupLinks.push({
+          source: nameToNode[sourceId],
+          target: nameToNode[sgLink.target_type],
+          exists: 1,
```
abgeorge7: what is "exists" for?

qingyashu: that's for previous logic in `createNodesAndEdges`, I think it's used for filtering out 0 count links for data model graph.

src/GraphUtils/utils.js
```diff
@@ -1,3 +1,29 @@
+/**
+ * Get subgroup links from link
+ * @param {array} link - array of links
```
abgeorge7: this is picky but usually if something is an array it feels like it should be plural, so `links` instead of `link`

abgeorge7: The code seems like `link` is not an array but an object though?

qingyashu:
yeah it's typo... fixed

PR #474, opened 2019-02-20
Fix/checkbox efficiency

abgeorge7:
Deployed at https://abby.planx-pla.net/submission/files

Added two things to make the frontend handle large lists faster
 - virtualized table for each group
- used object set implementation instead of javascript set

### New Features
- Faster UI! Added virtualized tables so there's no need to render everything at once

### Improvements
- UI will render faster and not be slow with the virtualized tables
- Checkbox functionality itself is slightly faster by removing the Javascript Set


ZakirG:
Checkboxes work great! Files mapped successfully. Whole flow was quick and smooth.


Probably out of scope of this PR, but -- I think this message could be formatted better.
screen shot 2019-02-22 at 11 48 44 am

PR #476, opened 2019-02-22
PXP-2464 fix/homepage chart REST API

qingyashu:
This PR change homepage chart to hit REST API instead of using relayer, and also enable `public` option for index page. 
- To set the chart nodes, a data commons should add `components.index.homepageChartNodes` for required nodes and chart labels. 
- To enable public index page, by setting `components.index.public`. 
- For backward compatibility, we still keep relayer to avoid crash. Until we gradually change all other data common' `gitops.json`/`${APP}.json` towards using REST API, then relayer-related code could be moved. 

### New Features
- Customizable public index page, by setting `components.index.public: true`. 

### Breaking Changes
- Change relayer to REST API for homepage chart: use `components.index.homepageChartNodes` for required nodes and chart labels.


qingyashu:
Branch deployed in https://qingya.planx-pla.net/, now its index page is open to public.

src/Login/Login.jsx
```diff
@@ -68,7 +78,7 @@ class Login extends React.Component {
           <hr className='login-page__separator' />
           <div className='body-typo'>{this.props.data.text}</div>
           {
-            this.props.providers.map(
+            this.props.providers && this.props.providers.map(
```
abgeorge7: is this necessary to add if you added default props?

qingyashu: Oh I changed the code yesterday and forgot to commit... let me update that! thanks

src/Index/page.jsx
```diff
@@ -1,15 +1,21 @@
...
 import { components } from '../params';
-import getProjectsList from './relayer';
+import getProjectNodeCounts from './utils';
 import './page.less';
 
 class IndexPageComponent extends React.Component {
   constructor(props) {
     super(props);
-    getProjectsList();
+    getProjectNodeCounts((res) => {
```
abgeorge7: I think that this should go in `componentDidMount`, not the constructor.
https://reactjs.org/docs/react-component.html#constructor

PR #481, opened 2019-03-05
Export To Workspace

benJrohrer:
Export to workspace button, status toaster, post manifest to manifests endpoint.

### New Features
- Export to workspace button, cohort is exported and made available in workspace
- Toaster display for status of export proess

### DIFF
--- a/src/DataExplorer/DataExplorerVisualizations.jsx
+++ b/src/DataExplorer/DataExplorerVisualizations.jsx
@@ -9,9 +9,10 @@ import SummaryChartGroup from '../components/charts/SummaryChartGroup/.';
 import PercentageStackedBarChart from '../components/charts/PercentageStackedBarChart/.';
 import DataSummaryCardGroup from '../components/cards/DataSummaryCardGroup/.';
 import { getCharts } from '../components/charts/helper';
-import { downloadManifest, downloadData, getManifestEntryCount } from './actionHelper';
+import { downloadManifest, downloadData, getManifestEntryCount, exportToWorkspace } from './actionHelper';
 import { calculateDropdownButtonConfigs, humanizeNumber } from './utils';
 import { exportAllSelectedDataToCloud } from './custom/bdbag';
+import { workspaceUrl } from '../localconf';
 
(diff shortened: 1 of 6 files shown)


src/DataExplorer/DataExplorerVisualizations.jsx
```diff
@@ -53,6 +60,45 @@ class DataExplorerVisualizations extends React.Component {
     );
   }
 
+  exportToWorkspaceCallback = () => {
+    this.setState({toasterOpen: true});
```
abgeorge7: just for style can you add spaces between the curly braces and text?
`this.setState({ toasterOpen: true });`

abgeorge7: these can also be combined into one line
`this.setState({ toasterOpen: true, exportedToWorkspace: true });`

src/DataExplorer/DataExplorerVisualizations.jsx
```diff
@@ -125,6 +176,27 @@ class DataExplorerVisualizations extends React.Component {
       : null;
     const selectedTableRowsCount = this.props.selectedTableRows.length;
     const dropdownConfigs = calculateDropdownButtonConfigs(this.props.explorerTableConfig);
+    const toaster =
```
abgeorge7: instead of defining this here and reusing the css from `MapDataModel.jsx` it would be good to abstract it out into its own component

src/localconf.js
```diff
@@ -52,6 +52,7 @@ function buildConfig(opts) {
   const graphqlSchemaUrl = `${hostname}data/schema.json`;
   const workspaceUrl = '/lw-workspace/';
   const workspaceErrorUrl = '/no-workspace-access/';
+  const manifestUrl = `${hostname}manifests/`
```
abgeorge7: I think ESLint is mad at you because you didn't export this

philloooo:
the button should gray out until there are cases/subjects selected that has files. should use the similar logic for that as the 'download manifest' button

PR #484, opened 2019-03-11
flow for workspace-token-service

benJrohrer:
Connect to Workspace Token Service from Workspace

### New Features
- When Workspace component is mounted, check if user has token. If so, continue. If not, connect to workspace token service to generate token.


philloooo:
merged to the other

PR #486, opened 2019-03-15
Files counts for buttons and chart

abgeorge7:

It is confusing to users when exporting or downloading data if the number does not match the number in the table - files have a different number, so we've added an optional file count chart as well as the counts for the buttons.

### New Features
- Configuration for a file chart
- Counts on the buttons to clarify what is being downloaded and exported

### DIFF
--- a/src/DataExplorer/DataExplorerVisualizations.jsx
+++ b/src/DataExplorer/DataExplorerVisualizations.jsx
@@ -81,14 +81,16 @@ class DataExplorerVisualizations extends React.Component {
     return clickFunc;
   }
 
-  isFileButton = (buttonConfig) => buttonConfig.type === 'manifest' ||
+  isFileButton = buttonConfig => buttonConfig.type === 'manifest' ||
       buttonConfig.type === 'export' ||
       buttonConfig.type === 'export-to-workspace';
 
   refreshManifestEntryCount = () => {
     if (this.props.dataExplorerConfig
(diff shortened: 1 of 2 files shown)


src/DataExplorer/DataExplorerVisualizations.jsx
```diff
@@ -81,10 +81,16 @@ class DataExplorerVisualizations extends React.Component {
...
   }
 
+  isFileButton = buttonConfig => buttonConfig.type === 'manifest' ||
+      buttonConfig.type === 'export' ||
+      buttonConfig.type === 'export-to-workspace';
+
   refreshManifestEntryCount = () => {
     if (this.props.dataExplorerConfig
       && this.props.dataExplorerConfig.buttons
-      && this.props.dataExplorerConfig.buttons.some(btnCfg => btnCfg.type === 'manifest' && btnCfg.enabled)) {
```
qingyashu: I haven't checked with other commons' gitops but here `manifest` type is removed, will it affect other commons if it's `manifest`?

abgeorge7: I made a function to determine if a button has anything to do with files! `isFileButton`

qingyashu: Oh I didn't see the that. 😆   looks good to me!

PR #488, opened 2019-03-21
fix(docker): add flag to npm ci and remove unused package to avoid warning

qingyashu:
There's warning message from docker image build, and ui-component is using npm prepared compiled files under `dist`. This PR solves the warning and add a flag to `npm ci` to allow docker generate prepared files. 

### Deployment changes
  - Fix warning from fsevents packages during docker image building
  - Add flag to `npm ci` in dockerfile to support npm prepared files from `ui-component` library


qingyashu:
Branch tested in my dev env https://qingya.planx-pla.net/ , server starts correctly.

PR #489, opened 2019-03-21
fix(icons): add more icons in data dictionary

qingyashu:
Add more category icons to data dictionary viewer

### Improvements
 - Add more category icons to data dictionary viewer


qingyashu:
Yeah there's some subgroups in svg files that'll cause bug during graph rendering, so I simplify it somehow and now the bug is gone.

PR #491, opened 2019-03-25
Upgrade Webpack and minify bundle

abgeorge7:

This upgrades a lot of outdated packages and minifies our bundle.js to make it faster.

### New Features
- Updated packages, most notably Webpack v4


package.json
```diff
@@ -5,14 +5,14 @@
   "main": "index.js",
   "dependencies": {
```
qingyashu: There're several packages that are no in use anymore, here's a list and could you check and delete them: 
d3(already have partial packages so we don't need this whole package anymore) , fast-csv, js-file-download, material-design-icons, material-ui-icons, multisort, npm, react-form, react-highlight, react-motion, react-remarkable, svg-fill-loader

abgeorge7: I used `depcheck` and remove the ones it showed - the ones you listed + cssnano and polished

PR #494, opened 2019-03-26
fix(dd-button-bug): tiny css changes to fix dd graph button off style

qingyashu:
There're some style bugs for buttons in the dd graph viewer's node popup and in graph table. This PR changes some CSS to fix them. 

### Bug Fixes
  - Fix some dd graph button off style bugs


ZakirG:
Very good!!

PR #498, opened 2019-03-27
Toaster Component

benJrohrer:
Implemented new toaster component. 

### New Features
- Toaster component


src/components/Toaster.less
```diff
@@ -0,0 +1,26 @@
+@import '~@gen3/ui-component/dist/css/base.less';
```
qingyashu: Remove this

src/components/Toaster.jsx
```diff
@@ -0,0 +1,82 @@
...
+      null
+    ;
+  }
+}
+
+Toaster.propTypes = {
+  isEnabled: PropTypes.bool.isRequired,
+  primaryText: PropTypes.string,
+  otherText: PropTypes.string,
+  hasCloseButton: PropTypes.bool,
```
qingyashu: I think 3 props for a button is too many: how about remove this `hasCloseButton` and only check `onCloseButton` is null or not for rendering. also same suggestion for action button

src/components/Toaster.jsx
```diff
@@ -0,0 +1,82 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import Button from '@gen3/ui-component/dist/components/Button';
+import './Toaster.less';
+
+class Toaster extends React.Component {
+  constructor(props) {
+    super(props);
+    this.state = {
+      closed: false,
```
qingyashu: Is this `closed` state used anywhere?

src/components/Toaster.jsx
```diff
@@ -0,0 +1,82 @@
...
+    );
+
+    return this.props.isEnabled ?
+      (
+        <div className='toaster__div'>
+          {closeButton}
+          {actionButton}
+          <div className='toaster__div-text'>
+            <div> {this.props.primaryText} </div>
+            <div> {this.props.otherText} </div>
```
qingyashu: Why using 2 props for text here? Why don't we just use 1 `text` prop instead, since you are going to combine them anyways, and they don't have any style differences.

src/components/Toaster.less
qingyashu: also rename this `Toaster.less` file into `Toaster.css` since we are not using any less syntax here

src/components/Toaster.jsx
```diff
@@ -0,0 +1,82 @@
...
+          </div>
+        </div>
+      )
+      :
```
abgeorge7: for style, this should be `: null` on one line
and don't think you need the semicolon on line 53 after it

PR #503, opened 2019-04-08
feat/niaid-decision-tree-poc

ZakirG:
NIAID Decision Tree POC App. New pieces in windmill:
- Cohort Tools button in navbar
- New cohort-tools page with a left-sidebar for filter application
- 2 Download Cohort buttons which return a JSON file for PTC or Control case

The elastic search query is performed asynchronously; the counts in the diagram will be spinners in the meantime, and the "Confirm" button that applies the filters should be disabled during that time. 

See the ticket for a description of the subject-filtering algorithm:
https://ctds-planx.atlassian.net/browse/PXP-2771

One known limitation of the algorithm is that when visits are missing between 2 follow ups, we take them as adjacent even though they are not, so that visits 10 and 14 will be treated as 6 months apart, even though the time may be much greater.


src/HIVCohortFilter/HIVCohortFilter.jsx
```diff
@@ -0,0 +1,364 @@
...
+
+    const slidingWindowSize = Math.ceil(thisContext.state.numConsecutiveMonthsFromUser / 6);
+    if (visitArray.length < slidingWindowSize + 1) {
+      thisContext.setState({ subjectNeither:
+                [...thisContext.state.subjectNeither, subjectWithVisits] });
+      return;
+    }
+
+    // The sliding window step. window is of size this.state.numConsecutiveMonthsFromUser
+    for (let i = 0; i < visitArray.length - slidingWindowSize; i += 1) {
```
thanh-nguyen-dang: The algorithm will be faster if you do in reversed direction:
- Find the sliding window of interested therapy from the end. Let's call it `follow_ups[i, i + interested_length]`.
- If that sliding windows continue with a visit that has `thrpyv != interested therapy` and `viral_load  <= 4000`. Return that windows. If not, you can skip/jump to all sliding windows `follow_ups[h, i]` where `for all j`, such that `h <= j <=i, follow_ups[j] == interested_therapy`. Because, at then end of whatever the sliding windows, the post treatment condition will be violated.

PR #513, opened 2019-04-22
Feat: NDH's EC & LTNP Cohort Filter Apps

ZakirG:
Deployed at: https://qa-niaid.planx-pla.net/cohort-tools

Before deployment to production, the leu3n and fposdate fields need to be added to the follow_up document in NIAID production.

Notes about LTNP: 
- This algorithm assumes that once a subject is HIV positive, they remain HIV positive. 
- If followups are missing between fposdate and currentYear, they don't count against the subject -- the CD4 count is assumed to remain above the threshold for the missing visits.


package.json
```diff
@@ -10,9 +10,11 @@
     "@fortawesome/fontawesome-svg-core": "^1.2.2",
     "@fortawesome/free-solid-svg-icons": "^5.2.0",
     "@fortawesome/react-fontawesome": "^0.1.0",
-    "@gen3/ui-component": "0.2.0",
+    "@gen3/ui-component": "0.2.2",
     "@storybook/addon-actions": "^5.0.0",
     "@storybook/react": "^5.0.0",
+    "babel-core": "^6.26.0",
```
qingyashu: Could you remove those two babel packages? we're using `@bable/core` above already

src/HIVCohortFilter/HIVCohortCases.jsx
```diff
@@ -0,0 +1,1205 @@
...
+    // Overridden by LTNP
+    const fileName = `control-cohort-vload-${this.state.viralLoadFromUser.toString()
+    }-months-${this.state.numConsecutiveMonthsFromUser.toString()}.json`;
+
+    const blob = this.makeCohortJSONFile(this.state.subjectControl);
+    FileSaver.saveAs(blob, fileName);
+  }
+}
+
+class PTCCase extends HIVCohortFilterCase {
```
qingyashu: Is it possible `PTCCase`, `ECCase`, `LTNPCase` to separate files? this files over 1k+ lines 😛

src/HIVCohortFilter/HIVCohortCases.jsx
```diff
@@ -0,0 +1,1205 @@
...
+    this.numConsecutiveMonthsInputRef = React.createRef();
+    this.showCount = this.showCount.bind(this);
+  }
+
+  getBucketByKeyWithHAARTVAR(bucketKey, isHAART) {
+    // The below query differs from the PTC case in that there is no viral_load check.
+    // This is because the EC and LTNP cases uses this function to find people who have
+    // never received haart treatments; we need to look at *all* their followups.
+    // (Read the function getBucketByKey() defined in the HIVCohortFilterCase class.)
+    const query = `
```
qingyashu: This is just code structure and refactoring: since we're having more and more queries it'd be good to separate requests and UI rendering apart, how about make those queries into another util files for UI component to import?

package.json
abgeorge7: these things shouldn't be necessary? they are old versions of babel

src/HIVCohortFilter/HIVCohortCases.jsx
```diff
@@ -0,0 +1,1205 @@
...
+import CohortPTCSvg from '../img/cohort-PTC.svg';
+import CohortECSvg from '../img/cohort-EC.svg';
+import CohortLTNPSvg from '../img/cohort-LTNP.svg';
+import Spinner from '../components/Spinner';
+
+class HIVCohortFilterCase extends React.Component {
+  // Base class for the 3 NDH cohort filter apps. Meant to facilitate code reuse
+  constructor(props) {
+    super(props);
+    this.updateFilters = this.updateFilters.bind(this);
```
abgeorge7: if you use arrow functions (ES6 syntax) you don't have to bind, just fyi

src/HIVCohortFilter/HIVCohortCases.jsx
```diff
@@ -0,0 +1,1205 @@
...
+        query: queryString,
+      }),
+      method: 'POST',
+    })
+      .then(
+        ({ status, data }) => data, // eslint-disable-line no-unused-vars
+      );
+  }
+
+  makeSubjectToVisitMap(followUps) { // eslint-disable-line class-methods-use-this
```
abgeorge7: what are you disabling with eslint here?

src/HIVCohortFilter/HIVCohortCases.jsx
```diff
@@ -0,0 +1,1205 @@
...
+
+class HIVCohortFilterCase extends React.Component {
+  // Base class for the 3 NDH cohort filter apps. Meant to facilitate code reuse
+  constructor(props) {
+    super(props);
+    this.updateFilters = this.updateFilters.bind(this);
+    this.checkReadyToCalculate = this.checkReadyToCalculate.bind(this);
+    this.downloadControl = this.downloadControl.bind(this);
+    this.therapyValuesOfInterest = ['HAART'];
+  }
```
abgeorge7: there is no default state but I see you adding state later - should add a default state, and maybe `therapyValuesOfInterest` can be in it ?

src/index.jsx
```diff
@@ -183,6 +184,15 @@ async function init() {
                         />)
                       }
                     />
+                    <Route
```
abgeorge7: remove this since we're using apps :)

src/HIVCohortFilter/HIVCohortCases.jsx
ZakirG: class-methods-use-this

ZakirG: this is like a "static" method kinda. a helper function. but i didnt mark it as static because it seems like that means something else in react

abgeorge7: right, but why all the eslint disabling?

src/HIVCohortFilter/HIVCohortCases.jsx
ZakirG: This is true! But I'd argue this particular refactor would make the code less readable -- the algorithm logic is easier to follow if the query is in the same place. After splitting the three cases into 3 files, each file is under 400 lines, i feel like that's an appropriate file length already?

src/index.jsx
ZakirG: ✓

src/HIVCohortFilter/HIVCohortCases.jsx
ZakirG: ✓

package.json
ZakirG: ✓

src/HIVCohortFilter/HIVCohortCases.jsx
ZakirG: ✓

package.json
ZakirG: ✓

src/HIVCohortFilter/HIVCohortCases.jsx
ZakirG: removed this eslint-disable ✓ 
but i left a few in the codebase. there are some that i don't know how to fix. for example: 
```
return fetchWithCreds({
      path: `${arrangerGraphqlPath}`,
      body: JSON.stringify({
        query: queryString,
      }),
      method: 'POST',
    })
      .then(
        ({ status, data }) => data, // eslint-disable-line no-unused-vars
      );
```

eslint talks about "status" being unused. but i can't remove that variable :O that's how callbacks go right?

PR #517, opened 2019-04-25
Feat/tiered access data explorer

mfshao:

### New Features
- Implemented blocking table for authenticated but not authorized view
- Utilized `ConnectFilter` from `Guppy`
- Data explorer now supports tiered access by using `Guppy`

### Improvements
- Overall UI/UX improvement to meet design requirements for tiered access data explorer page


src/GuppyDataExplorer/ExplorerButtonGroup/index.jsx
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable max-len */
```
qingyashu: Could we resolve the max-len warning?

src/GuppyDataExplorer/ExplorerTopMessageBanner/index.jsx
```diff
@@ -0,0 +1,55 @@
...
+      </div>
+    );
+  }
+}
+
+ExplorerTopMessageBanner.propTypes = {
+  className: PropTypes.string,
+  getAccessButtonLink: PropTypes.string,
+  tierAccessLevel: PropTypes.string.isRequired,
+  tierAccessLimit: PropTypes.number, // inherit from GuppyWrapper
```
qingyashu: Remove this ` //inherit from GuppyWrapper` comment?

src/GuppyDataExplorer/GuppyDataExplorer.jsx
```diff
@@ -35,12 +37,21 @@ class GuppyDataExplorer extends React.Component {
...
+            guppyConfig={this.props.guppyConfig}
+            getAccessButtonLink={config.dataExplorerConfig.getAccessButtonLink}
+          />
+          <ExplorerFilter
             className='guppy-data-explorer__filter'
-            filterConfig={this.props.filterConfig}
-            guppyConfig={{ type: this.props.guppyConfig.dataType, ...this.props.guppyConfig }}
-            fieldMapping={this.props.guppyConfig.fieldMapping}
+            guppyConfig={this.props.guppyConfig}
+            getAccessButtonLink={config.dataExplorerConfig.getAccessButtonLink}
```
qingyashu: Move this `getAccessButtonLink` to `/GuppyDataExplorer/index.jsx` to keep all config items in one place?

.travis.yml
```diff
@@ -7,10 +7,6 @@ git:
 node_js:
   - "10"
 
-cache:
-  directories:
-    - node_modules # NPM packages
```
mfshao: Have to remove this caching feature, otherwise the build will fail due to using old dependencies.

This appears to be a known issue for travis-ci https://github.com/travis-ci/travis-ci/issues/3758

abgeorge7: @zflamig is this an issue?

package.json
```diff
@@ -10,8 +10,8 @@
     "@fortawesome/fontawesome-svg-core": "^1.2.2",
     "@fortawesome/free-solid-svg-icons": "^5.2.0",
     "@fortawesome/react-fontawesome": "^0.1.0",
-    "@gen3/guppy": "git+https://github.com/uc-cdis/guppy.git",
-    "@gen3/ui-component": "0.2.2",
+    "@gen3/guppy": "git+https://github.com/uc-cdis/guppy.git#feat/locked-fields",
+    "@gen3/ui-component": "git+https://github.com/uc-cdis/gen3-ui-component.git#feat/lockable-filter",
```
abgeorge7: I don't think we should merge branches into master....

src/GuppyDataExplorer/ExplorerButtonGroup/index.jsx
```diff
@@ -53,7 +53,8 @@ class ExplorerButtonGroup extends React.Component {
         });
     }
     if (nextProps.totalCount !== this.props.totalCount) {
-      this.refreshManifestEntryCount();
+      this.refreshManifestEntryCount().catch(() => {
```
abgeorge7: what is the point of the catch?

.travis.yml
qingyashu: @abgeorge7  I think Mingfei will recover this change when we merge ui-component and guppy

src/GuppyDataExplorer/ExplorerTable/index.jsx
```diff
@@ -1,10 +1,12 @@
+/* eslint-disable max-len */
```
abgeorge7: instead of disabling linting can you just put a break after the <p> tags in the problem line (88 I think)

abgeorge7: unless the problem line is due to a comment, then nvm

src/GuppyDataExplorer/ExplorerTopMessageBanner/index.jsx
```diff
@@ -0,0 +1,55 @@
...
+                  className='top-message-banner__button'
+                  buttonType='default'
+                  onClick={
+                    (this.props.getAccessButtonLink) ? (
+                      () => { window.open(this.props.getAccessButtonLink); }
+                    ) : (() => {})
+                  }
+                />
+              </div>
+              <span className='top-message-banner__normal-text'>To protect data security, you can only narrow the cohort down to </span>
```
abgeorge7: Can we change this message? I think "to protect data security" sounds awkward
Maybe something like "Due to lack of access, you are only able to narrow the cohort down to X ${whatever they call the records}. Please request additional access if necessary"

I think we also shouldn't hardcode "subjects" - some groups say cases, some patients, etc

src/GuppyDataExplorer/ExplorerVisualization/index.jsx
```diff
@@ -59,6 +61,10 @@ class ExplorerVisualization extends React.Component {
     const chartData = this.getData(this.props.aggsData, this.props.chartConfig, this.props.filter);
     const tableColumns = (this.props.tableConfig.fields && this.props.tableConfig.fields.length > 0)
       ? this.props.tableConfig.fields : this.props.allFields;
+    const isComponentLocked = checkForAnySelectedUnaccessibleField(this.props.aggsData,
+      this.props.accessibleFieldObject, this.props.guppyConfig.accessibleValidationField);
+    const lockMessage = `You cannot see this chart because it contains items under ${this.props.tierAccessLimit} ${this.props.nodeCountTitle.toLowerCase()}`;
```
abgeorge7: I don't think the ${this.props.nodeCountTitle.toLowerCase()} accounts for plurals, "50 case" sounds awkward

src/GuppyDataExplorer/TierAccessSelector/index.jsx
```diff
@@ -0,0 +1,118 @@
...
+  };
+
+  handleToggle = () => {
+    this.setState(prevState => ({ toggled: !prevState.toggled }));
+  }
+
+  render() {
+    return (
+      <div className='tier-access-selector'>
+        <h4 // eslint-disable-line jsx-a11y/no-noninteractive-element-interactions
```
abgeorge7: also don't think we should disable it - it just needs a `role` and `tabIndex`
good for accessibility :)

src/GuppyDataExplorer/TierAccessSelector/index.jsx
```diff
@@ -0,0 +1,118 @@
...
+                </div>
+              </div>
+              {
+                this.state.selected !== 'with-access' && (
+                  <div className='tier-access-selector__button-wrapper'>
+                    <Button
+                      label='Get Access'
+                      className='tier-access-selector__button'
+                      buttonType='default'
+                      onClick={
```
abgeorge7: can we just make the get access prop required? it seems weird to have a button that goes nowhere if the prop isn't passed in

package.json
mfshao: Yeah, I'll wait for guppy and ui-component's PR to be merged and tagged. After that I'll update this package.json

src/GuppyDataExplorer/TierAccessSelector/index.jsx
qingyashu: `tabIndex` work for `div` but not `h4` according to this rules https://github.com/evcohen/eslint-plugin-jsx-a11y/blob/master/docs/rules/no-noninteractive-element-interactions.md .  😂

src/GuppyDataExplorer/TierAccessSelector/index.jsx
qingyashu: yeah sure could we have a default url for this button?

src/GuppyDataExplorer/TierAccessSelector/index.jsx
abgeorge7: can we make it a div then? :)

src/GuppyDataExplorer/ExplorerVisualization/index.jsx
mfshao: The lock message will be rephrased as suggested by @ZakirG here https://github.com/uc-cdis/gen3-ui-component/pull/39#discussion_r282515232

src/GuppyDataExplorer/ExplorerTable/index.jsx
mfshao: yeah it is because of a comment, I'll disable `max-len` for that line only

src/GuppyDataExplorer/TierAccessSelector/index.jsx
mfshao: I think we left this prop to be optional is because we want this button's behavior to be more customizable, like displaying a toaster by default?

But if we give it a default URL, what would the URL be?

src/GuppyDataExplorer/ExplorerButtonGroup/index.jsx
mfshao: There was an error when the component is loading at the first time.

Actually found a right way to stop the error from happening, will remove this `catch`.

src/GuppyDataExplorer/ExplorerVisualization/index.jsx
mfshao: also for plurals, how about we use this? https://github.com/blakeembrey/pluralize

src/GuppyDataExplorer/ExplorerTable/index.jsx
```diff
@@ -82,35 +83,48 @@ class ExplorerTable extends React.Component {
...
+          pages={(this.props.isLocked) ? 0 : visiblePages} // Total number of pages, don't show 10000+ records in table
           loading={this.state.loading}
           onFetchData={this.fetchData}
           defaultPageSize={this.props.defaultPageSize}
           className={'-striped -highlight '}
-          minRows={0} // hide empty rows
+          minRows={3} // make room for no data component
           resizable={false}
+          NoDataComponent={() => (this.props.isLocked ? (
+            <div className='rt-noData'>
```
ZakirG: this className looks strange to me. should it be rt-no-data for consistency?

mfshao: ah `rt-noData` is actually defined by `ReactTable`. I'm just overriding its content here

ZakirG: aha!! i see

src/GuppyDataExplorer/TierAccessSelector/index.jsx
mfshao: change default url to https://gen3.org/

.travis.yml
mfshao: restored

PR #526, opened 2019-05-07
Fix login logout issues - PXP-3000

ZakirG:
This PR focuses on the user session lifecycle. Deployed at https://zakir.planx-pla.net/

### Bug fixes
-> The query page now calls updateSession() before executing a query. This fixes the bug where query results come back empty with a 200 response in the case where the user's session has expired. 

-> The refreshSession function now hits /user/user instead of just /user, so that we can notify the user with the ReduxAuthTimeoutPopup if their session is over (/user is public, /user/user is not). 

### Code changes
-> The SessionMonitor no longer logs out users. If a user has been inactive and the config has logoutInactiveUsers == true, we just decline to refresh their session and allow Fence to log them out. At this point, we start using Sheepdog to check their auth status instead of Fence, so that we can check their auth status without refreshing the access token. 

-> Users are allowed to keep a workspace tab open for the duration of the Fence session lifetime without being logged out (likely 8 hours). We can't detect user activity within the workspace iframe, so a solution to detect their activity is not trivial. (Note that once a Jupyter pod is spawned, it stays alive regardless of whether or not we redirect their browser to the login page.) Allowing "inactivity" on this page will make it easier for users to perform long-running analysis operations in a notebook overnight (the redirect didn't work before anyway, because the browser displays that "You have unsaved changes" popup which can be declined).

-> The session is now updated every 5 minutes, rather than every 10 minutes. If a user's auth has expired and the session is not updated, the user won't know they've been logged out until they navigate to a new page (or until they notice that pages are suddenly refusing to serve them data). Increasing the updateSession frequency is helpful in this situation. 

### Deployment changes
New manifest.json variable: logout_inactive_users. Defaults to true if not provided.

https://github.com/uc-cdis/cloud-automation/pull/818

### DIFF
--- a/runWebpack.sh
+++ b/runWebpack.sh
@@ -39,7 +39,6 @@ gitops_config() {
   local manifestFile
   local portalApp
   local portalGitopsFolder
-  local logoutInactiveUsers
 
   gitRepo="cdis-manifest"
   hostname="$1"
--- a/runWebpack.sh
+++ b/runWebpack.sh
@@ -12,6 +12,8 @@
 export APP="${APP:-dev}"
 export NODE_ENV="${NODE_ENV:-dev}"
 export HOSTNAME="${HOSTNAME:-"revproxy-service"}"
+echo "THIS"
+echo $LOGOUT_INACTIVE_USERS
 export LOGOUT_INACTIVE_USERS="${LOGOUT_INACTIVE_USERS:-"true"}"
 
 
--- a/runWebpack.sh
+++ b/runWebpack.sh
@@ -70,15 +70,8 @@ gitops_config() {
     return 1
   fi
   echo "INFO: gitops_config - setting APP=$portalApp"
-  logoutInactiveUsers="$(jq -r .global.logout_inactive_users < $manifestFile)"
-  if [[ -z "$logoutInactiveUsers" ]]; then
-    echo "ERROR: unable to determine logout_inactive_users from manifest at $manifestFile"
-    return 1
-  fi
-  echo "INFO: gitops_config - setting LOGOUT_INACTIVE_USERS=$logoutInactiveUsers"
   export HOSTNAME="$hostname"
   export APP="$portalApp"
-  export LOGOUT_INACTIVE_USERS="$logoutInactiveUsers"
   if [[ "$portalApp" == "gitops" ]]; then
     portalGitopsFolder="../$gitRepo/$hostname/portal"
     local it=0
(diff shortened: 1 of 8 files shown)


runWebpack.sh
```diff
@@ -67,8 +70,15 @@ gitops_config() {
     return 1
   fi
   echo "INFO: gitops_config - setting APP=$portalApp"
+  logoutInactiveUsers="$(jq -r .global.logout_inactive_users < $manifestFile)"
```
qingyashu: I think you already add env var in cloud automation, https://github.com/uc-cdis/cloud-automation/pull/818/files#diff-95b4ad3baa4d2705112b07887f1cd448R82,  so you wont need to parse manifest json here

webpack.config.js
```diff
@@ -23,12 +23,14 @@ const plugins = [
...
   new webpack.EnvironmentPlugin(['APP']),
   new webpack.EnvironmentPlugin({'BASENAME': '/'}),
+  new webpack.EnvironmentPlugin(['LOGOUT_INACTIVE_USERS']),
   new webpack.EnvironmentPlugin(['REACT_APP_PROJECT_ID']),
   new webpack.EnvironmentPlugin(['REACT_APP_ARRANGER_API']),
   new webpack.EnvironmentPlugin(['REACT_APP_DISABLE_SOCKET']),
   new webpack.DefinePlugin({ // <-- key to reducing React's size
     'process.env': {
       'NODE_ENV': JSON.stringify(process.env.NODE_ENV || 'dev'),
+      LOGOUT_INACTIVE_USERS: !(process.env.LOGOUT_INACTIVE_USERS === 'false'),
```
qingyashu: For better readability, could we change this line to `process.env. LOGOUT_INACTIVE_USERS || 'false'` (assume by default is `false`) ?

src/localconf.js
```diff
@@ -50,6 +50,7 @@ function buildConfig(opts) {
     title: 'Login from Google',
   };
   const loginPath = `${userapiPath}login/`;
+  const logoutInactiveUsers = !(process.env.LOGOUT_INACTIVE_USERS === 'false');
```
qingyashu: Same thought here as for the one in `webpack.config.js`

src/SessionMonitor/index.js
```diff
@@ -35,25 +36,62 @@ export class SessionMonitor {
...
+    return this.refreshSession();
   }
 
   refreshSession() {
+    if (this.isUserOnPage('login')) {
+      return Promise.resolve(0);
+    }
     // hitting Fence endpoint refreshes token
-    fetch(userapiPath);
+    const self = this;
```
qingyashu: I think ES6's arrow function will take `this` as the one from outer scope, so you don't need to create an extra `self` here

webpack.config.js
ZakirG: Hmm by default it's set to true. The tricky part for me was that the bash script brings the boolean in as a string :/

PR #528, opened 2019-05-13
fix: account for trailing slashes in URLs

ZakirG:
Qingya found a small bug in the function isUserOnPage in the SessionMonitor -- the updateSession function logs out an inactive user in the workspace if the URL ends with a slash. This PR fixes that issue.

### DIFF
--- a/src/SessionMonitor/SessionMonitor.test.js
+++ b/src/SessionMonitor/SessionMonitor.test.js
@@ -24,4 +24,27 @@ describe('SessionMonitor', () => {
     sessionMonitor.updateSession();
     expect(refreshSessionSpy).toHaveBeenCalledTimes(1);
   });
+
+  it('detects the page correctly', () => {
+    const sessionMonitor = new SessionMonitor(500, 10000000);
+    expect(
+      sessionMonitor.pageFromURL("https://niaid.bionimbus.org/workspace/") 
+    ).toEqual('workspace');
+
+    expect(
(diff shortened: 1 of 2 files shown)


src/SessionMonitor/index.js
```diff
@@ -37,7 +37,8 @@ export class SessionMonitor {
   }
 
   isUserOnPage(pageName) {
-    const paths = window.location.href.split('/').filter(x => x !== 'dev.html');
+    let paths = window.location.href.split('/').filter(x => x !== 'dev.html');
+    paths = paths.filter(path => path !== ""); // To account for trailing slashes
```
qingyashu: Could you chain this `filter` with previous line, and also make the double quote to single quote. 
And it'd be better to add unit tests for this function :)

src/SessionMonitor/SessionMonitor.test.js
```diff
@@ -24,4 +24,27 @@ describe('SessionMonitor', () => {
...
+    expect(
+      sessionMonitor.pageFromURL("https://niaid.bionimbus.org/workspace") 
+    ).toEqual('workspace');
+
+    expect(
+      sessionMonitor.pageFromURL("https://niaid.bionimbus.org/dev.html/workspace/") 
+    ).toEqual('workspace');
+
+    expect(
+      sessionMonitor.pageFromURL("data.braincommons.org/login")
```
qingyashu: Thanks! another small one this time: let's not add real sponsor's url here, could you change them to "http://example.com/workspace" etc.?

PR #529, opened 2019-05-13
feat(sanity-check): check portal config works for homepage charts

qingyashu:
Before starting the server, let's check if the homepage chart config in portal config is correct. 

### Improvements
- Before starting, check if the homepage chart config in portal config is correct

### DIFF
--- a/sanity-check.js
+++ b/sanity-check.js
@@ -1,32 +1,35 @@
 const dict = require('./data/dictionary.json');
 const params = require('./src/params.js');
 
-/**
- * this transfer '_xxx_count' to 'xxx'
- * @param {string} countName
- */
-function parseNodeType(countName) {
-  return countName.substring(1, countName.length - '_count'.length);
-}
-const boardCountNodes = params.config.graphql.boardCounts.map(e => parseNodeType(e.graphql));
-const chartCountNodes = params.config.graphql.boardCounts.map(e => parseNodeType(e.graphql));
-
-/**
- * Check if all nodes exist in dictionary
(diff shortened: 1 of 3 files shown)


sanity-check.js
```diff
@@ -0,0 +1,32 @@
...
+
+/**
+ * this transfer '_xxx_count' to 'xxx'
+ * @param {string} countName
+ */
+function parseNodeType(countName) {
+  return countName.substring(1, countName.length - '_count'.length);
+}
+const boardCountNodes = params.config.graphql.boardCounts.map(e => parseNodeType(e.graphql));
+const chartCountNodes = params.config.graphql.boardCounts.map(e => parseNodeType(e.graphql));
```
mfshao: should be `graphql.chartCounts` ?

qingyashu: 🤔  why? it's getting the node name from chartCounts config....

qingyashu: Oh I get it....typo here !!

sanity-check.js
```diff
@@ -0,0 +1,32 @@
...
+ */
+function checkNodesExistInDict(nodeList) {
+  for (let i = 0; i < nodeList.length; i += 1) {
+    const nodeID = nodeList[i];
+    if (typeof dict[nodeID] === 'undefined') {
+      console.error(`ERR: "${nodeID}" does not exist`);
+      process.exit(1);
+    }
+  }
+  return true;
```
mfshao: is this return necessary?

PR #536, opened 2019-05-24
Updated logo and homepage charts for BRAIN

abgeorge7:

Changes the logo and homepage charts to meeting BRAIN needs.

### Improvements
- Updated logo
- Relevant homepage stats


data/config/bhc.json
```diff
@@ -70,7 +70,6 @@
       ]
     },
     "navigation": {
-      "title": "BRAIN Commons",
```
ZakirG: why this change?

abgeorge7: Because the logo has the title in it so just going to use that as the nav title.
Similar to NIAID!

PR #537, opened 2019-05-29
fix(ui): bump ui and guppy version, revise some UI, add canine config

qingyashu:
This PR does following things: 
 - bump ui component version 0.3.2, better bar charts, and "show more"/"show less" buttons
 - bump guppy version to make an explorer config optional to avoid console err
 - Fix a bug that "export to workspace" button is not enabled at first time. 
 - Add canine portal config

### Bug Fixes
 - Fix a bug that guppy explorer's "export to workspace" button not enabled at first time

### Improvements
 - Better bar charts for guppy explorer with "show more"/"show less" buttons
 - Revise some guppy explorer options
 - Add canine portal config


src/GuppyDataExplorer/ExplorerButtonGroup/index.jsx
```diff
@@ -53,8 +53,7 @@ class ExplorerButtonGroup extends React.Component {
         });
     }
     if (nextProps.totalCount !== this.props.totalCount
-      && nextProps.totalCount
-      && this.props.totalCount) {
```
mfshao: is this to fix the export to manifest btn disable bug? I sort of remembered that we put this implementation to prevent an error msg from showing in console? 🤔

mfshao: tested again but didn't see any errors. i guess my memory is not right. Anyway, pls ignore this and sorry for the confusions 🤦‍♂

PR #543, opened 2019-06-04
fix(chart): fix index chart crash when less than 3 nodes provided

qingyashu:
It happens in IBDGC portal: For index chart config block inside portal config, when only 2 or less nodes are provided, the chart component will crash. This PR add more out-of-index check and refine the initialization logic. 

### Bug Fixes
  - fix  index chart crash bug when less than 3 nodes

### DIFF
--- a/src/Index/relayer.js
+++ b/src/Index/relayer.js
@@ -2,6 +2,7 @@ import { fetchQuery } from 'relay-runtime';
 import environment from '../environment';
 import { GQLHelper } from '../gqlHelper';
 import getReduxStore from '../reduxStore';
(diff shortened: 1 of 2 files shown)


qingyashu:
For index chart, we recommend gradually change to use peregrine's another endpoint for fetching index chart data, just add another `homepageChartNodes` config block in gitops.json, under `components.index`, so that portal won't use graphql (or relay), and there won't be any strange multi rendering and refreshing for index chart. Pls see example config at: https://github.com/uc-cdis/cdis-manifest/blob/master/gen3.datastage.io/portal/gitops.json#L71

PR #545, opened 2019-06-13
Fix Integer casting for Data Upload

abgeorge7:

### Bug Fixes
- For the data upload process, we were casting everything to a string. This created issues when the required input was a number. Fixes to parse as an when needed integer. 
- Also fixes UI bug with the floating table header on the Data Upload page

### DIFF
--- a/src/Submission/MapDataModel.jsx
+++ b/src/Submission/MapDataModel.jsx
@@ -101,11 +101,20 @@ class MapDataModel extends React.Component {
     const fields = this.state.requiredFields;
     fields[prop] = null;
     if (option && option.target) {
-      fields[prop] = option.target.value;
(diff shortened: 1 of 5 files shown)


src/Submission/MapDataModel.jsx
```diff
@@ -100,14 +100,26 @@ class MapDataModel extends React.Component {
...
     }
+    fields[prop] = castOption || null;
     this.setState({ requiredFields: fields });
   }
 
+  isInteger = (prop) => {
+    if (this.state.nodeType && this.props.dictionary[this.state.nodeType] &&
+      this.props.dictionary[this.state.nodeType].properties[prop]) {
+      return this.props.dictionary[this.state.nodeType].properties[prop].type === 'integer' ||
+          this.props.dictionary[this.state.nodeType].properties[prop].type === 'number';
```
mfshao: could its value be non-integer if `type === number`?

PR #547, opened 2019-06-18
Merge File Explorer with Data Explorer

abgeorge7:

### New Features
- Merges the file explorer with the data explorer

### Improvements
- File explorer that uses ES index instead of slow graphQL queries


src/GuppyDataExplorer/ExplorerVisualization/index.jsx
```diff
@@ -81,18 +81,26 @@ class ExplorerVisualization extends React.Component {
...
+          chartData.countItems.length > 0 && (
+            <div className='guppy-explorer-visualization__summary-cards'>
+              <DataSummaryCardGroup summaryItems={chartData.countItems} connected />
+            </div>
+          )
+        }
+        {
+          chartData.summaries.length > 0 && (
+            <div className='guppy-explorer-visualization__charts'>
+              <SummaryChartGroup
```
mfshao: Can we make more spaces between bar charts? Right now the tooltip is overlapping with the chart below

abgeorge7: I think thats fine - the tool tip could have a ton of data and we can't account for that so I just made the z-index higher so its the top most element

PR #549, opened 2019-06-25
fix: guppy explorer using incorrect manifest format

ZakirG:
I believe this PR will fix the original bug that Sean reported.

### DIFF
--- a/src/GuppyDataExplorer/ExplorerButtonGroup/index.jsx
+++ b/src/GuppyDataExplorer/ExplorerButtonGroup/index.jsx
@@ -101,7 +101,7 @@ class ExplorerButtonGroup extends React.Component {
     resultManifest = resultManifest.filter(
       x => typeof x[fileFieldInFileIndex] !== 'undefined'
     );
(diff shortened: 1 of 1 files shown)


src/GuppyDataExplorer/ExplorerButtonGroup/index.jsx
```diff
@@ -98,7 +98,12 @@ class ExplorerButtonGroup extends React.Component {
       },
       [caseFieldInFileIndex, fileFieldInFileIndex],
     );
-    resultManifest = resultManifest.filter(x => typeof x[fileFieldInFileIndex] !== 'undefined');
+    resultManifest = resultManifest.filter(
+      x => typeof x[fileFieldInFileIndex] !== 'undefined'
+    );
+    for(let i = 0; i < resultManifest.length; i += 1) {
+      resultManifest[i][caseField] = [ resultManifest[i][caseField] ];
```
mfshao: should be `caseFieldInFileIndex` instead of `caseField ` here?

PR #553, opened 2019-07-09
feat(privacy-policy): add privacy policy page and footer

rudyardrichter:
Notes for people trying this in cdistest:
- set the portal app to `gitops` in the manifest
- make a `portal` directory in your manifest folder (same directory as `g3k_manifest_path`)
  - add a `portal/gitops.json` for the config.
    - for this PR, it should have a line `privacyPolicyFile: "/src/privacy_policy.md"`
- run `gen3 kube-setup-portal`

### New Features

- Add privacy policy link in footer
- Add `/privacy-policy` route

### DIFF
--- a/runWebpack.sh
+++ b/runWebpack.sh
@@ -78,7 +78,6 @@ gitops_config() {
   export HOSTNAME="$hostname"
   export APP="$portalApp"
   if [[ "$portalApp" == "gitops" ]]; then
-    gitopsFolder="../$gitRepo/$hostname"
     portalGitopsFolder="../$gitRepo/$hostname/portal"
     local it=0
     local copySource
--- a/runWebpack.sh
+++ b/runWebpack.sh
@@ -78,6 +78,7 @@ gitops_config() {
   export HOSTNAME="$hostname"
   export APP="$portalApp"
   if [[ "$portalApp" == "gitops" ]]; then
+    gitopsFolder="../$gitRepo/$hostname"
     portalGitopsFolder="../$gitRepo/$hostname/portal"
     local it=0
     local copySource
--- a/src/PrivacyPolicy/PrivacyPolicy.jsx
+++ b/src/PrivacyPolicy/PrivacyPolicy.jsx
@@ -1,0 +1,36 @@
+import React from 'react';
+import Spinner from '../components/Spinner';
+
+class PrivacyPolicy extends React.Component {
+  constructor(props) {
+    super(props);
+  }
+
+  render() {
+    this.props.loadPolicy();
+    console.log('render');
+    console.log(this.props.loaded)
+    if (!this.props.loaded) {
+      console.log('not loaded')
+      return <Spinner />;
+    } else if (this.props.asdf) {
+      console.log('loaded')
+      return (
+        <div className='privacy-policy'>
+          {this.props.asdf}
+        </div>
+      )
+    } else {
+      // otherwise we'll try to redirect to the url according to the environment variable
+      const url = process.env.PRIVACY_POLICY_URL;
+      if (url) {
+        // redirect to given URL
+        this.props.history.push(url);
+      } else {
+        return null
+      }
+    }
+  }
+}
+
+export default PrivacyPolicy;
--- a/src/PrivacyPolicy/PrivacyPolicy.jsx
+++ b/src/PrivacyPolicy/PrivacyPolicy.jsx
@@ -1,23 +1,43 @@
 import React from 'react';
 import Spinner from '../components/Spinner';
+import showdown from 'showdown';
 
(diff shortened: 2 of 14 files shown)


runWebpack.sh
```diff
@@ -78,6 +78,7 @@ gitops_config() {
   export HOSTNAME="$hostname"
   export APP="$portalApp"
   if [[ "$portalApp" == "gitops" ]]; then
+    gitopsFolder="../$gitRepo/$hostname"
```
abgeorge7: is this used anywhere? just saw its the only change in the file

src/PrivacyPolicy/PrivacyPolicy.jsx
```diff
@@ -0,0 +1,55 @@
+import PropTypes from 'prop-types';
+import React from 'react';
+import showdown from 'showdown';
+import Spinner from '../components/Spinner';
+import './PrivacyPolicy.less';
+
+class PrivacyPolicy extends React.Component {
+  componentDidMount() {
+    this.props.dispatch(dispatch => fetch('/src/privacy_policy.md').then(
```
abgeorge7: what if there isn't a privacy policy here? wondering if this can be a var vs a hardcoded path

abgeorge7: also think the dispatch stuff should live in the Redux component vs. the UI component

src/PrivacyPolicy/PrivacyPolicy.jsx
```diff
@@ -0,0 +1,55 @@
...
+        </div>
+      );
+    }
+    // otherwise we'll try to redirect to the url according to the environment variable
+    const url = process.env.PRIVACY_POLICY_URL;
+    if (url) {
+      // redirect to given URL
+      this.props.history.push(url);
+    }
+    return '';
```
abgeorge7: can return null if nothing is rendered

src/PrivacyPolicy/PrivacyPolicy.less
```diff
@@ -0,0 +1,8 @@
+.privacy-policy {
```
abgeorge7: for CSS we try to use BEM instead of nesting:
https://gen3.org/resources/developer/style-guide/#css-style-guide

src/components/layout/Footer.less
```diff
@@ -53,6 +54,46 @@
...
+  vertical-align: middle;
+  display: inline-flex;
+  flex-direction: column;
+  flex-grow: 1;
+  flex-basis: 20%;
+  justify-content: center;
+  height: 100%;
+  color: white;
+  text-decoration: none;
+  &:hover {
```
abgeorge7: same here with BEM we try to avoid nesting

src/PrivacyPolicy/PrivacyPolicy.less
rudyardrichter: recording conversation here for posterity/thoroughness: the privacy policy HTML will be dynamically generated from markdown, so there's not a way to apply BEM in that case

src/PrivacyPolicy/PrivacyPolicy.jsx
rudyardrichter: this path is determined by the thing in `customize.sh` that copies it into the `src` directory, so it seems like this would have to be hardcoded

src/components/layout/Footer.less
rudyardrichter: also for this case, is there a way to get the `&:hover` functionality without nesting? seems like this pattern is used [elsewhere in the code](https://github.com/uc-cdis/data-portal/blob/6723c9f14061f81db57ee26e124567e0ddd29265/src/UserAgreement/Quiz.less#L74-L76) too

mfshao: I think you can write the same way as original css? Something like https://github.com/uc-cdis/data-portal/blob/63408352a83eb428827ee9f04254bb310a2c9fab/src/components/layout/NavBar.less#L97

runWebpack.sh
rudyardrichter: not used. fixed in 6723c9f

src/PrivacyPolicy/PrivacyPolicy.jsx
```diff
@@ -0,0 +1,43 @@
...
+    if (url) {
+      // redirect to given URL
+      this.props.history.push(url);
+    }
+    return null;
+  }
+}
+
+PrivacyPolicy.propTypes = {
+  loadPrivacyPolicy: PropTypes.function.isRequired,
```
mfshao:
```suggestion
  loadPrivacyPolicy: PropTypes.func.isRequired,
```

PR #558, opened 2019-07-26
fix/HIV app and Guppy

mfshao:
Let NIAID's HIV app support Guppy

### New Features
- Replace Arranger by Guppy for HIV apps

### DIFF
--- a/src/GuppyDataExplorer/ExplorerFilter/index.jsx
+++ b/src/GuppyDataExplorer/ExplorerFilter/index.jsx
@@ -38,6 +38,10 @@ class ExplorerFilter extends React.Component {
         }
       }
     }
+    return null;
+  }
+
+  componentDidUpdate() {
   }
 
   /**
--- a/src/HIVCohortFilter/HIVCohortFilterCase.jsx
+++ b/src/HIVCohortFilter/HIVCohortFilterCase.jsx
@@ -1,15 +1,21 @@
 import React from 'react';
 import FileSaver from 'file-saver';
-import { arrangerGraphqlPath } from '../localconf';
 import { fetchWithCreds } from '../actions';
+import {
+  guppyGraphQLUrl,
+  arrangerGraphqlPath,
+  useGuppyForExplorer,
+} from '../configs';
 
 class HIVCohortFilterCase extends React.Component {
   // Base class for the 3 NDH cohort filter apps. Meant to facilitate code reuse
-  static performQuery(queryString) {
+  static performQuery(queryString, variableString) {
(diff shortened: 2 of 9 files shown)


src/GuppyDataExplorer/ExplorerFilter/index.jsx
```diff
@@ -38,6 +38,10 @@ class ExplorerFilter extends React.Component {
         }
       }
     }
+    return null;
```
mfshao: to address a warning found in console while testing using dev.html.
`Warning: ExplorerFilter.getSnapshotBeforeUpdate(): A snapshot value (or null) must be returned. You have returned undefined.`

src/GuppyDataExplorer/ExplorerFilter/index.jsx
```diff
@@ -38,6 +38,10 @@ class ExplorerFilter extends React.Component {
...
+  }
+
+  componentDidUpdate() {
```
mfshao: to address a warning found in console while testing using dev.html.
`Warning: ExplorerFilter: getSnapshotBeforeUpdate() should be used with componentDidUpdate(). This component defines getSnapshotBeforeUpdate() only.`

mfshao:
deployed in https://qa-niaid.planx-pla.net

mfshao:
> When I run this locally (qa-niaid) I get: SyntaxError: Unexpected token u in JSON at position 0
> 
> but maybe my local config is out of date? Is it deployed to qa-niaid where we can test there?

hmmm... yeah it is deployed on QA-NIAID now

src/HIVCohortFilter/HIVCohortFilterCase.jsx
```diff
@@ -1,15 +1,21 @@
...
   // Base class for the 3 NDH cohort filter apps. Meant to facilitate code reuse
-  static performQuery(queryString) {
+  static performQuery(queryString, variableString) {
+    const graphqlUrl = useGuppyForExplorer ? guppyGraphQLUrl : arrangerGraphqlPath;
     return fetchWithCreds({
-      path: `${arrangerGraphqlPath}`,
+      path: `${graphqlUrl}`,
       body: JSON.stringify({
         query: queryString,
+        variables: JSON.parse(variableString),
```
ZakirG: In the case where we use arranger, variableString will be undefined right? And I think that will lead to the error Uncaught SyntaxError: Unexpected token u in JSON at position 0

mfshao: nice catch! fixed in https://github.com/uc-cdis/data-portal/pull/558/commits/1dd79067b948a0cd77857bbc25b6548e37931ed9

PR #561, opened 2019-08-01
Feat/dat

paulineribeyre:
Updates for the Data availability tool

Needs a new "dataAvailabilityToolConfig" block in gitops.json. Example:
```
  "dataAvailabilityToolConfig": {
    "guppyConfig": {
      "dataType": "follow_up",
      "mainField": "harmonized_visit_number",
      "mainFieldTitle": "Visit number",
      "mainFieldIsNumeric": true,
      "aggFields": [
        "age_at_visit",
        "abc",
        "abcv",
        "drug_used",
        "thrpy",
        "thrpyv",
        "trz",
        "trzv",
        "cd4dt",
        "chol",
        "leu2n",
        "viral_load",
        "cocuse",
        "drinkcat",
        "eductn",
        "emotl",
        "employ",
        "income",
        "insurance",
        "paidsex"
      ],
      "fieldMapping": [
        { "field": "abc", "name": "Abcavir_since_last_visit" },
        { "field": "abcv", "name": "Abcavir_at_visit" },
        { "field": "drug_used", "name": "drug_taken_frequency_6mons" },
        { "field": "thrpy", "name": "therapy_type_since_last_visit" },
        { "field": "thrpyv", "name": "therapy_type_at_visit" },
        { "field": "trz", "name": "Terazol_since_last_visit" },
        { "field": "trzv", "name": "Terazol_at_visit" },
        { "field": "cd4dt", "name": "year_CD4_collected" },
        { "field": "chol", "name": "Cholesterol(mg/dL)" },
        { "field": "leu2n", "name": "CD8_positive_cells" },
        { "field": "drinkcat", "name": "drink_category" },
        { "field": "eductn", "name": "highest_education" },
        { "field": "emotl", "name": "emotion_wellness" }
      ],
      "colorRange": ["#09e8d9", "#e36702"]
    }
  }
```
`mainFieldIsNumeric`: optional, default is false
`fieldMapping`: optional
`colorRange`: optional, default is ['#EBF7FB', '#3188C6']

TODOS:
- [x] un-hardcode grid height

### New Features
- Data Availability Tool in data explorer

### Dependency updates
- DAT requires Guppy >= 0.3.0 

### Deployment changes
- DAT requires Guppy >= 0.3.0


package.json
```diff
@@ -10,7 +10,7 @@
     "@fortawesome/fontawesome-svg-core": "^1.2.2",
     "@fortawesome/free-solid-svg-icons": "^5.2.0",
     "@fortawesome/react-fontawesome": "^0.1.0",
-    "@gen3/guppy": "0.1.2",
+    "@gen3/guppy": "git://github.com/uc-cdis/guppy.git#feat/dat",
```
mfshao: just to note down here in case we forget: change this after a new Guppy has been released

src/GuppyDataExplorer/ExplorerHeatMap/index.jsx
```diff
@@ -0,0 +1,183 @@
...
+    );
+    const yAxisVarsMapping = this.props.guppyConfig.fieldMapping;
+    const data = this.getTransformedData(yAxisVars);
+    const height = '450px'; // TODO need to display all yaxis vars https://github.com/hustcc/echarts-for-react/issues/208
+
+    return (
+      <React.Fragment>
+        {
+          data && data.length ? (
+            <div className={'explorer-heat-map'}>
```
abgeorge7: just a nitpick thing, no need for the `{` and `}`

src/GuppyDataExplorer/GuppyDataExplorer.jsx
```diff
@@ -62,6 +63,7 @@ class GuppyDataExplorer extends React.Component {
             nodeCountTitle={this.props.guppyConfig.nodeCountTitle || capitalizeFirstLetter(
               this.props.guppyConfig.dataType)}
             tierAccessLimit={this.props.tierAccessLimit}
+            nestedAggsData={this.nestedAggsData}
```
abgeorge7: does nestedAggsData ever change from `{}` ?

src/GuppyDataExplorer/ExplorerHeatMap/index.jsx
```diff
@@ -0,0 +1,183 @@
...
+import './ExplorerHeatMap.less';
+
+function round(value, decimals) {
+  return Number(`${Math.round(`${value}e${decimals}`)}e-${decimals}`);
+}
+
+class ExplorerHeatMap extends React.Component {
+  constructor(props) {
+    super(props);
+    this.maxCellValue = 0; // updated when data is received
```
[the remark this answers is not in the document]
paulineribeyre: it's updated during `getTransformedData()` which is used during `render()`, i can't update the state while rendering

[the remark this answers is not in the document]
paulineribeyre: i could make `getTransformedData` return both `transformedData` and `maxCellValue` though if you think that's cleaner

package.json
mfshao:
```suggestion
    "@gen3/guppy": "^0.3.0",
```

src/GuppyDataExplorer/ExplorerHeatMap/index.jsx
```diff
@@ -0,0 +1,185 @@
...
+    visualMap: { // legend
+      min: 0,
+      max: this.maxCellValue, // round up (1 decimal) of max value
+      precision: 2, // 2 decimals in label
+      calculable: true, // handles on legend to adjust selected range
+      orient: 'horizontal',
+      left: 'center', // horizontal alignment
+      align: 'right', // position of bar relatively to handles and label
+      inRange: {
+        color: ['#EBF7FB', '#3188C6'], // [smallest value, greatest value]
```
mfshao: could you make these colors customizable? I'm thinking `['#EBF7FB', '#3188C6']` could be the default colors but if there is `categorical2Colors` defined in config files then use those two colors are heatmap color?

Something like this https://github.com/uc-cdis/data-portal/blob/master/src/GuppyDataExplorer/ExplorerVisualization/index.jsx#L67

src/GuppyDataExplorer/ExplorerHeatMap/index.jsx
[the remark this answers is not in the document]
abgeorge7: its fine haha

paulineribeyre:
when tiered access is enabled, the query that gets the list of IDs from the current filter is unauthorized (the user doesn't have access to query these IDs), so the heatmap would be displayed without a filter.
i am choosing to completely hide the heatmap instead of displaying "This chart is hidden because it contains fewer than xxx subjects" like the other charts do, because the reason for hiding it is not the number of subjects

PR #564, opened 2019-08-14
DAT improvements

paulineribeyre:
### Improvements
- DAT: do not automatically capitalize field names when a mapping is provided
- DAT: add "subject_id" to the configurable field name mapping
- DAT: show X axis values when the field is not numeric

### DIFF
--- a/docs/data_availability_tool.md
+++ b/docs/data_availability_tool.md
@@ -26,8 +26,10 @@ The DAT requires the `dataAvailabilityToolConfig` block in `gitops.json`. An exa
 }
 ```
 
-- `mainFieldIsNumeric`: optional, default is false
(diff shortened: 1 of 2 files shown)


docs/data_availability_tool.md
```diff
@@ -0,0 +1,33 @@
...
+            {"field": "thrpyv", "name": "Therapy Type"},
+            {"field": "trzv", "name": "Terazol Use"},
+            {"field": "subject_id", "name": "Subjects"}
+        ],
+        "colorRange": ["#09e8d9", "#e36702"]
+    }
+}
+```
+
+- `mainFieldIsNumeric`: optional, default is false
```
mfshao: maybe provide some more explanation about this option?

PR #565, opened 2019-08-16
Feat/button auth

paulineribeyre:
PXP-3189 and part of PXP-3508

Disable UI components based on arborist access.
Requires https://github.com/uc-cdis/cloud-automation/pull/959

Needs a new "componentToResourceMapping" block in gitops.json. Example:

```
"componentToResourceMapping": {
  "Workspace" : {
    "resource": "/workspace",
    "method": "access",
    "service": "jupyterhub"
  },
  "SomeUIComponent" : {
    "resource": "/abc/xyz",
    "method": "read",
    "service": "someService"
  }
}
```

When a user logs in, we ask arborist whether this user has access to each resource defined in `componentToResourceMapping`. We send multiple requests to Arborist's `auth/proxy` endpoint.

~A possible improvement is to expose the `auth/resources` endpoint so that we can fetch all the resources the user has access to in a single call.~ We are not using the `auth/resources` endpoint because it lists all the resources the user has access to, regardless of which actions the user is allowed to perform on the resource

### New Features
- Disable UI components based on arborist access

### Deployment changes
- Needs a version of cloud-automation that exposes arborist's `auth/proxy` endpoint

### DIFF
--- a/data/config/default.json
+++ b/data/config/default.json
@@ -283,5 +283,12 @@
         "file_format"
       ]
     }
+  },
+  "componentToResourceMapping": {
+    "Workspace" : {
+      "resource": "/workspace",
+      "method": "access",
+      "service": "jupyterhub"
+    }
   }
 }
--- a/data/config/default.json
+++ b/data/config/default.json
@@ -284,11 +284,5 @@
       ]
     }
   },
-  "componentToResourceMapping": {
-    "Workspace" : {
-      "resource": "/workspace",
-      "method": "access",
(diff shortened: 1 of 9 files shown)


paulineribeyre:
Note that this PR only implements the disabling of UI components based on arborist permissions for the big buttons in the navigation bar and the "Export to workspace" button in the explorer

But the functionality can be implemented easily for other components by using the redux store's `state.userAccess.access` prop

PR #570, opened 2019-09-06
feat: use indexd authz config setting

ZakirG:
TLDR: Before centralized auth, we use the ACL field to check for unmapped files, but after centralized auth, we use the AUTHZ field. 

### Deployment changes
Windmill lists unmapped files by checking Indexd. Before centralized auth's deployment, a file was determined to be unmapped by checking if the acl field was null. After centralized auth, we use the authz field instead. The new manifest configuration setting `use_indexd_authz` in the global block specifies which of the two fields should be used. This field defaults to false.

Deployed in https://qa-canine.planx-pla.net with fence version < 3, index authz field set to false 
(by default).

Deployed in https://qa-brain.planx-pla.net with fence version > 3, index authz field set to true.


README.md
```diff
@@ -181,6 +181,17 @@ Then, specify all the required certificates that need to be completed before usi
 ```
 Default is an empty list.
 
+### File mapping configuration
```
mfshao: these should go into the template of this PR and probably in a seperated doc under `/docs`

probably also an good idea to update https://github.com/uc-cdis/cloud-automation/blob/master/kube/services/portal/README.md as well

README.md
```diff
@@ -181,6 +181,17 @@ Then, specify all the required certificates that need to be completed before usi
...
+Windmill lists unmapped files by checking Indexd. Before centralized auth's deployment, a file was determined to be
+unmapped by checking if the acl field was null. After centralized auth, we use the authz field instead. The manifest
+configuration setting `use_indexd_authz` specifies which of the two fields should be used. This field defaults to false.
+```
+"global": {
+    ...
+    "use_indexd_authz": "true"
```
mfshao:
```suggestion
    "use_indexd_authz": true
```

src/Submission/ReduxMapFiles.js
```diff
@@ -2,38 +2,43 @@ import { connect } from 'react-redux';
...
-    case 200:
-      total = total.concat(data.records);
-      if (data.records.length === fetchLimit) {
-        return dispatch(
-          fetchUnmappedFiles(user, total, data.records[fetchLimit - 1].did, fetchLimit),
-        );
+const fetchUnmappedFiles = (user, total, start, fetchLimit) => (dispatch) => {
+  let unmappedFilesCheck = useIndexdAuthz ? 'authz=null' : 'acl=null';
+  console.log('ReduxMapFiles.js -- useIndexdAuthz =  ', useIndexdAuthz);
+  console.log('ReduxMapFiles.js -- unmappedFilesCheck =  ', unmappedFilesCheck);
```
mfshao: remove these console prints?

src/Submission/ReduxSubmissionHeader.js
```diff
@@ -2,42 +2,47 @@ import { connect } from 'react-redux';
...
-        type: 'RECEIVE_UNMAPPED_FILE_STATISTICS',
-        data: {
-          count: totalSize.length,
-          totalSize: totalSize.reduce((total, current) => total + current.size, 0),
-        },
-      };
+const fetchUnmappedFileStats = (user, totalSize, start, fetchLimit) => dispatch => { 
+  let unmappedFilesCheck = useIndexdAuthz ? 'authz=null' : 'acl=null';
+  console.log('ReduxMapFiles.js -- useIndexdAuthz =  ', useIndexdAuthz);
+  console.log('ReduxMapFiles.js -- unmappedFilesCheck =  ', unmappedFilesCheck);
```
mfshao: also these?

README.md
ZakirG: hmm are you sure about this one? i see quotation marks around boolean values in production manifests

mfshao: yeah in manifest.json there are `"Ture"`, `"true"` and `true`...

in portal configs we use `true` only. I think any of these styles will work, just trying to make a bit of consistance. But no big deal if leave as is

mfshao:
Also we should test in QA first before merging

README.md
ZakirG: Ah, but this is not a portal config setting, it is a manifest.json setting :O

PR #575, opened 2019-09-26
feat(tabs): allow multiple tabs in explorer

qingyashu:
User could customize more than one tab in explorer. Take tb.json as example, just add config JSON blocks under `explorerConfigs`, like: 
```
{
   explorerConfigs: {
       Data: {
           // data explorer config goes here
       },
       File: {
           // file explorer config goes here
       },
       "Follow Up": {
           // follow up explorer config goes here
       }
   }
}
```

Notice that we will gradually deprecate `dataExplorerConfig` and `fileExplorerConfig` later if we want multiple filter tabs in explorer page, use `explorerConfigs` instead. 
But for backward compatibility, previous `dataExplorerConfig` and `fileExplorerConfig` is still valid if there's no `explorerConfigs` block. so that old config won't break portal. 

### New Features
 -  Support multiple tabs in explorer in addition to Data, e.g. File, Folowups, etc.

### DIFF
--- a/src/GuppyDataExplorer/index.jsx
+++ b/src/GuppyDataExplorer/index.jsx
@@ -108,7 +108,7 @@ class Explorer extends React.Component {
               </div>
             ) : null
           }
-          <div className={Object.keys(config.explorerConfigs).length === 1 ? 'guppy-explorer__main' : ''}>
+          <div className={Object.keys(config.explorerConfigs).length > 1 ? 'guppy-explorer__main' : ''}>
             <GuppyDataExplorer
               key={this.state.tab}
               chartConfig={guppyExplorerConfigs[this.state.tab].charts}
--- a/src/GuppyDataExplorer/index.jsx
+++ b/src/GuppyDataExplorer/index.jsx
@@ -48,15 +48,24 @@ const defaultFileConfig = {
   dropdowns: {},
 };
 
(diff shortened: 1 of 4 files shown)


src/GuppyDataExplorer/index.jsx
```diff
@@ -48,16 +48,25 @@ const defaultFileConfig = {
...
+let routes = [
   '/explorer',
   '/files',
 ];
 
+const converNameToLink = (name) => {
+  const linkName = name.replace(' ', '_').toLowerCase();
+  return `/explorer/${linkName}`;
+};
+if (config.explorerConfigs) { // using new multi-tab configuration
```
mfshao: use `Object.keys(config.explorerConfigs).length > 1` to check?

otherwise it will error and not fall back to use old configs if it has `explorerConfigs : {}`

src/GuppyDataExplorer/index.jsx
```diff
@@ -48,16 +48,25 @@ const defaultFileConfig = {
...
   _.merge(defaultFileConfig, config.fileExplorerConfig),
 ];
-
-const routes = [
+let routes = [
   '/explorer',
   '/files',
 ];
 
+const converNameToLink = (name) => {
```
ZakirG: converNameToLink -> convertNameToLink

PR #584, opened 2019-10-01
feat(idx-button-auth): hide idx button if no auth

vpsx:
Builds on this PR https://github.com/uc-cdis/data-portal/pull/565 

Disables index button bars (card thingies?) based on Arborist access.

So if for example you have the following `componentToResourceMapping` block in gitops.json: 

```
"componentToResourceMapping": {
  "Analyze Data" : {
    "resource": "/workspace",
    "method": "access",
    "service": "jupyterhub"
  }
}
```
then only users with `/workspace` will see the Analyze Data card on the home page. 

### New Features
- Disable index button bars based on Arborist access

### DIFF
--- a/src/components/IndexButtonBar.jsx
+++ b/src/components/IndexButtonBar.jsx
@@ -9,35 +9,42 @@ import './IndexButtonBar.css';
  * @param {dictIcons, buttons} params
  */
 class IndexButtonBar extends Component {
+  canUserSeeComponent = (componentName) => {
+    const authResult = this.props.userAccess[componentName];
+    return typeof authResult !== 'undefined' ? authResult : true;
+  }
+
   render() {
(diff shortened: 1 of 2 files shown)


src/components/IndexButtonBar.jsx
```diff
@@ -9,35 +9,42 @@ import './IndexButtonBar.css';
...
-                    this.props.history.push(`${item.link}`);
-                  }}
-                  label={item.label}
-                  buttonType='primary'
-                />
-              </div>
-            ),
-          )
-        }
+        { indexButtonItems }
```
paulineribeyre: why did you choose to pull the code out into a variable?

vpsx: an excellent question indeed...

PR #586, opened 2019-10-03
fix/redirect to hostname

mfshao:
To address https://ctds-planx.atlassian.net/browse/PXP-4115
Use `hostname` for logout redirect URL so it could work for NDE

### Bug Fixes
- Use `hostname` as logout redirect URL


mfshao:
deployed at https://qa-test2.planx-pla.net

PR #587, opened 2019-10-03
Feat/presigned url btn

mfshao:
`Generate Sign URL` button for files, which can be enabled by putting `"signedURLButton": true` in `featureFlags` block of portal config

### New Features
- New `Generate Sign URL` button in file info page, requires `"signedURLButton": true` in `featureFlags` block

### Bug Fixes
- Add check for files that don't have required `data_format` field in metadata and prevent file info page from crashing in that case

### DIFF
--- a/src/CoreMetadata/CoreMetadataHeader.jsx
+++ b/src/CoreMetadata/CoreMetadataHeader.jsx
@@ -83,6 +83,10 @@ class CoreMetadataHeader extends Component {
         />);
       }
 
+      if (!this.props.metadata.data_format) {
+        /* eslint no-console: ["error", { allow: ["error"] }] */
+        console.error('WARNING: null value found for mandatory field \'data_format\', please verify the correctness of metadata');
+      }
       const properties = `${this.props.metadata.data_format} | ${fileSizeTransform(this.props.metadata.file_size)} | ${this.props.metadata.object_id} | ${this.dateTransform(this.props.metadata.updated_datetime)}`;
 
       return (
--- a/src/CoreMetadata/CoreMetadataHeader.jsx
+++ b/src/CoreMetadata/CoreMetadataHeader.jsx
@@ -8,6 +8,7 @@ import { userapiPath } from '../configs';
 const DOWNLOAD_BTN_CAPTION = 'Download';
 const SIGNED_URL_BTN_CAPTION = 'Generate Signed URL';
 const SIGNED_URL_MSG = 'Please copy your signed URL below (this generated signed URL will expires after 3600 seconds):';
+const SIGNED_URL_ERROR_MSG = 'A error has occurred when generating signed URL:';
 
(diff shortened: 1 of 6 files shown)


mfshao:
locally tested against `gen3qa-brain`, will figure out which QA env to deploy later

mfshao:
deployed on my dev env

here is a good file that can gives you a signed url: https://mingfei.planx-pla.net/files/1cb65d78-fd84-456a-b5f2-24571293b823
here is a dummy file that can only gives you an error: https://mingfei.planx-pla.net/files/00e6f5f4-0c71-4bc9-bd5e-77a2a494b3b9

src/CoreMetadata/CoreMetadataHeader.jsx
```diff
@@ -1,10 +1,14 @@
...
 import React, { Component } from 'react';
-import {
-  userapiPath,
-} from '../configs';
+import Popup from '../components/Popup';
+import { userapiPath } from '../configs';
 
 const DOWNLOAD_BTN_CAPTION = 'Download';
+const SIGNED_URL_BTN_CAPTION = 'Generate Signed URL';
+const SIGNED_URL_MSG = 'Please copy your signed URL below (this generated signed URL will expires after 3600 seconds):';
```
qingyashu: `expires` -> `expire`

qingyashu: `3600 seconds` -> `1 hour` ? (idk, just more human readable

src/CoreMetadata/CoreMetadataHeader.jsx
```diff
@@ -1,10 +1,14 @@
...
+const SIGNED_URL_ERROR_MSG = 'A error has occurred when generating signed URL:';
```
qingyashu: `A` -> `An`

src/CoreMetadata/CoreMetadataHeader.jsx
```diff
@@ -44,6 +48,15 @@ function canUserDownload(user, projectAvail, projectID) {
 }
 
 class CoreMetadataHeader extends Component {
+  onGenerateSignedURL = () => {
+    this.props.onGenerateSignedURL(this.props.metadata.object_id);
+  };
+
+  onSignedURLPopupClose = () => {
+    this.props.onUpdatePopup({ signedURLPopup: false });
+    this.props.onClearSignedURL();
```
qingyashu: 🤔 I don't understand here why do we need to clear url?

mfshao: I was just thinking to be secure? So if u close the popup then you lose it

but might be an overkill...

qingyashu: i see... maybe we don't need to worry about that :p up to you

PR #589, opened 2019-10-04
fix/display table

mfshao:
Make core metadata table appears agian

### Bug Fixes
- Fixed a bug that not showing the `CoreMetadataTable`


paulineribeyre:
where is this deployed?

ZakirG:
> where is this deployed?
I ran this locally against niaid.bionimbus 
`HOSTNAME=niaid.bionimbus.org APP=ndh NODE_ENV=dev bash ./runWebpack.sh
and then visited this url to see the table
https://niaid.bionimbus.org/dev.html/files/dg.0896/fabc2863-c2a3-443c-adfe-559bd9db7902

src/CoreMetadata/page.less
```diff
@@ -1,18 +1,35 @@
...
-.core-metadata-page__column {
-  float: left;
-  width: 50%;
+.core-metadata-page__grid { 
+  display: grid; 
+  grid-template-columns: repeat(2, 1fr); 
+  grid-auto-rows: auto;
+} 
+
+.core-metadata-page__box1 {
```
paulineribeyre: picture, header, table instead of box1, box2, box3 would be easier to understand

mfshao:
> where is this deployed?

in qa-canine now!

here is an example https://qa-canine.planx-pla.net/files/22ec1330-9a3e-492f-a24d-a1c4fe6617e9

mfshao:
> could you explain why it was not displayed before, which changes are fixing it and which changes are just improvements?

perviously it was css issues. If you set page column with `float: left` and `width: 50%` the two columns will take all the spaces and squzze the table into a element with width = 0px;

really the fix here is by setting `display: grid` and configures for each child box how many grids it will take. Other stuff like `overflow: auto` or `height:auto` are just some sytle improvement

PR #595, opened 2019-10-11
Feat/incommon login

paulineribeyre:
Adds an optional dropdown for each login button on the login page, to handle InCommon login.
Related to [this fence PR](https://github.com/uc-cdis/fence/pull/703)

- All "login entry" components (login button, description and optional dropdown) have the same width
- Slow filter when the dropdown contains all InCommon login options (3,000+):
I added a dependency to NPM package `react-select-fast-filter-options` to use the `createFilterOptions` function. This helps avoiding a long lag time when users type in the dropdown filter.
We are using `react-select` v1, so we don't have access to the `createFilter` function introduced in v2 (updating to v2 or v3 now would cause other problems) which we could also use to reduce lag time, for example:
```
import createFilter from 'react-select';
filterOption={createFilter({ ignoreAccents: false })}
```

### New Features
- Optional login dropdowns to allow logging in through InCommon

### Deployment changes
- InCommon login needs fence 4.8.0


src/Login/Login.jsx
```diff
@@ -36,6 +45,14 @@ class Login extends React.Component {
     this.state = getInitialState(window.innerHeight - 221);
     this.resetState = this.resetState.bind(this);
     this.updateDimensions = this.updateDimensions.bind(this);
+
+    // using objects for autoCompleteRefs, suggestionLists and
+    // currentLoginUrl because we need one for each login dropdown
+    this.autoCompleteRefs = {};
+    this.state = {
```
mfshao: add `height: getInitialState(window.innerHeight - 221),` into this block? This should fix the DNA image display issue

src/Login/Login.jsx
```diff
@@ -36,6 +45,14 @@ class Login extends React.Component {
     this.state = getInitialState(window.innerHeight - 221);
```
mfshao: you could remove this line

src/Login/Login.jsx
paulineribeyre: thank you 😭

paulineribeyre:
the only failing jenkins test is the workspace test that hasn't been fixed yet - merging

PR #592, opened 2019-10-10
Arborist UI integration

vpsx:
arborist ui integration--change UI based on user's authz. 
see spec [here](https://docs.google.com/document/d/1IsUlRwoNLUNtT5I9D2tzmepC3y8vdhjmuLfUh-wumvU)

Needs Arborist's GET /auth/mapping endpoint to be exposed via revproxy: 
https://github.com/uc-cdis/arborist/pull/110 
https://github.com/uc-cdis/cloud-automation/pull/1006 

To use this feature, add `"useArboristUI": true` in gitops.json

### New Features
- get auth mapping from arborist if useArboristUI true in config
- per-project submit data buttons change verbiage based on user's authz
- top bar submit data button changes verbiage based on user's authz
- toggle visibility of recent submissions table based on user's authz
- toggle visibility of map my files card based on user's authz
- toggle form submission and upload file button visibility on project page based on user's authz
- toggle node browser delete button visibility based on user's authz 

### Deployment changes 
- if deployed with "useArboristUI": true, needs Arborist's GET /auth/mapping endpoint to be exposed via revproxy. (useArboristUI false by default)

PR #598, opened 2019-10-14
PXP-4449 Refactor/auth mapping

vpsx:
Refactor and tidy up code for Arborist UI ( https://github.com/uc-cdis/data-portal/pull/592 )

see individual commits

### Improvements
- Refactor code for Arborist UI


src/utilsAuth.js
```diff
@@ -0,0 +1,39 @@
+// Functions for checking user's auth mapping
```
qingyashu: I noticed some code here doesn't have a property check / out-of-index check. It would be safe to handle those edge cases so that the website won't crash for null/undefined data. Thanks!

vpsx: Do you mean the string manipulation in `resourcePathFromProjectID`? If so: 
```
> split = ''.split('-')
[ '' ]
> split
[ '' ]
> program = split[0]
''
> project = split.slice(1).join('-')
''
> resourcePath = ["/programs", program, "projects", project].join('/')
'/programs//projects/'
>
```

vpsx: Or if you mean the `userAuthMapping`s, these should always be defined because of the `useArboristUI` checks, but if you want, I can add extra checks anyway?

qingyashu: Oh yeah maybe just things like `x['method'] === 'file_upload' && x['service'] === 'fence'` makes me nervous but I should read code more carefully. it's all up to you~ :p

vpsx: Ah, that particular fn is passed to `Array.some`, so either the array is empty and the fn isn't called at all, or the array is a list of actions from Arborist (so those fields are guaranteed by the API)--if the Arborist API changes, we should just be updating this code 😄  But I added checks for `userAuthMapping`!

vpsx:
PXP-4449

PR #599, opened 2019-10-15
PXP-4448 fix(button-css): button width to px

vpsx:
PXP-4448

### Bug Fixes
Give submit data button min and max widths


vpsx:
> how about setting both `max-width` and `min-width` to arbitry px values while also set `width` to a specifc percentage value?

Ah yes. I was following the precedent set by `g3-button`, but this makes a lot more sense. Thank you!!

PR #608, opened 2019-10-25
fix(hiv): change follow_up to visit

qingyashu:
### Bug Fixes
 - Fix HIV app by changing "follow_up" to "visit" according to dictionary update


mfshao:
this will be a breaking change since NDH is still using `follow_up`

I think we should note that down in the PR template

PR #610, opened 2019-10-25
Fix/permissions PXP-4524

vpsx:
This PR makes the profile page show authz info from fence or arborist or both or neither, depending on the config. Additionally it fixes the file page's download button. See details below. 

For download button, here is why we are throwing out the Fence authz checks: 
1. They appear to have been ad hoc in the first place 
2. They are broken in some cases in commons using centralized auth, whenever fence's userinfo project_access is no longer in sync with Arborist authz (e.g. synapse). 
3. To fall back on the Fence checks if no centralized auth would have required a third new "useArboristAuthz" config var and there was already pushback from the team against more config 
4. We are transitioning to centralized auth anyway

### New Features
- add config vars showArboristAuthzOnProfile and showFenceAuthzOnProfile
- make profile page show authz info from fence/arborist/both/neither based on configs 
- make file page 'download' button toggle according to arborist authz

### Breaking Changes
- file page 'download' button will always render if useArboristUI is false (it no longer respects Fence authz). If useArboristUI is true, it will render if user has read-storage. Previously, it checked user.project_access (from Fence userinfo) and rendered if user had read-storage.

### DIFF
--- a/src/CoreMetadata/CoreMetadataHeader.jsx
+++ b/src/CoreMetadata/CoreMetadataHeader.jsx
@@ -44,12 +44,12 @@ class CoreMetadataHeader extends Component {
 
   render() {
     if (this.props.metadata) {
-      // display the download button if the user can download this file
       const { projectAvail } = this.props;
       const projectId = this.props.metadata.project_id;
       let downloadButton = null;
       let signedURLButton = null;
 
+      // downloadButton should always render if useArboristUI false. Otherwise according to authz.
       if (
         !useArboristUI
         || userHasMethodOnProject('read-storage', projectId, this.props.userAuthMapping)
--- a/src/CoreMetadata/CoreMetadataHeader.jsx
+++ b/src/CoreMetadata/CoreMetadataHeader.jsx
@@ -3,8 +3,10 @@ import PropTypes from 'prop-types';
 import copy from 'clipboard-plus';
 import React, { Component } from 'react';
 import Popup from '../components/Popup';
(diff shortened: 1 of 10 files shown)


mfshao:
I'm still a bit of lost about the download button rendering logic. Would it be better if `useArboristUI 
 === false` then it falls back to check Fence authz?

vpsx:
@mfshao See PR description. Doesn't make sense to use `useArboristUI` as a flag for this case, because a commons can be using centralized auth but not be using the auth-based UI.

src/CoreMetadata/CoreMetadataHeader.jsx
```diff
@@ -63,12 +45,16 @@ class CoreMetadataHeader extends Component {
...
-      const canDownload = canUserDownload(user, projectAvail, projectId);
       let downloadButton = null;
       let signedURLButton = null;
-      if (canDownload) {
+
+      if (
+        !useArboristUI
+        || userHasMethodOnProject('read-storage', projectId, this.props.userAuthMapping)
+        || projectIsOpenData(projectAvail, projectId)
+      ) {
```
mfshao: can you add a bit of comment here explaining that this button will always render if `useArboristUI === false` so other people don't need to search for clues from all the PR comments?

vpsx: Yes, thanks!! Also removing the outdated comment a few lines prior, which I completely missed before this 🤦‍♀

PR #620, opened 2019-11-12
Fix(PXP-3514): Correctly handle unicode file uploads

mpingram:
Jira: https://ctds-planx.atlassian.net/browse/PXP-3514

Unicode files were handled incorrectly in the 'Upload File' view of the node explorer. This was due to the uploaded files being read using FileReader.readAsBinaryString(), which does not preserve unicode encoding -- this results in non-ascii characters, like 'š', being displayed as each individual byte, like 'Å¡'. Fixed by using FileReader.readAsTextString(). (Also automatically formatted the code.)

Deployed in my dev commons at https://mpingram.planx-pla.net. Steps to test:
0. Create a .tsv file with these contents:
```
type	project_id	submitter_id	acknowledgee	projects.id#1	projects.code#1
acknowledgement	DEV-mpingram-test	Tešt2	Tešt2	d288d790-2019-592e-a067-add692828d09	mpingram-test
```
1. Click on 'Submit Data' in the top navbar
2. Click on the 'Submit Data' button for project 'DEV-mpingram-test'
3. Click 'Upload File' and upload the .tsv file.
_Old behavior:_ special characters are corrupted. _New behavior_: special characters are preserved.

### Bug Fixes
- Fixed a bug where special characters in uploaded files, such as accents or diacriticals, would not be preserved.

### DIFF
--- a/src/Submission/SubmitTSV.jsx
+++ b/src/Submission/SubmitTSV.jsx
@@ -23,26 +23,6 @@ const SubmitTSV = ({ project, submission, onUploadClick,
   //
   const processUpload = (event) => {
     const f = event.target.files[0];
-    if (FileReader.prototype.readAsBinaryString === undefined) {
-      FileReader.prototype.readAsBinaryString = (fileData) => {
-        let binary = '';
-        const pt = this;
-        const reader = new FileReader();
-        // listener for when all the bytes have been read
-        //  https://developer.mozilla.org/en-US/docs/Web/API/FileReader
-        reader.onload = () => {
-          const bytes = new Uint8Array(reader.result);
-          const length = bytes.byteLength;
-          for (let i = 0; i < length; i += 1) {
-            binary += String.fromCharCode(bytes[i]);
-          }
-          // pt.result  - readonly so assign content to another property
-          pt.content = binary;
-          pt.onload();
-        };
-        reader.readAsArrayBuffer(fileData);
-      };
-    }
     const reader = new FileReader();
     let fileType = f.type;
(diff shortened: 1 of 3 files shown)


src/Submission/SubmitTSV.jsx
```diff
@@ -23,26 +23,6 @@ const SubmitTSV = ({ project, submission, onUploadClick,
   //
   const processUpload = (event) => {
     const f = event.target.files[0];
-    if (FileReader.prototype.readAsBinaryString === undefined) {
```
mfshao: i think the section you are remving here was meant to be used to support IE...

IDK if we are still going to support it, but if we do we might need to test this from a Windows env...

src/Submission/__test__/data.json
```diff
@@ -1,99 +1,124 @@
...
+    },
+    {
+      "acl": [
+        "QA",
+        "test"
+      ],
+      "baseid": "a388a5-eb87-e3",
+      "created_date": "2018-09-11T22:50:20.055349",
+      "did": "00041a-3114-4d9f00344",
+      "file_name": "File 6 - With a special character: 💩",
```
mfshao: awesome filename! 😆

src/Submission/SubmitTSV.jsx
mpingram: Yes, I was looking at the support tables for `FileReader.readAsTextString` -- `readAsTextString` works without polyfill for IE (https://developer.mozilla.org/en-US/docs/Web/API/FileReader#Browser_compatibility)

PR #621, opened 2019-11-14
(PXP-4707) Add pending state to explorer buttons

mpingram:
Jira: https://ctds-planx.atlassian.net/browse/PXP-4707?

Adds 'pending' state to buttons in Explorer when requests to Guppy are in flight.

Uses a combination of props passed from guppy and internal state in ExplorerButtonGroup to determine whether a request is in flight that will enable or disable the explorer's export / download buttons.

### New Features
- Adds 'pending' state to buttons in Explorer when requests to Guppy are in flight.

### DIFF
--- a/src/GuppyDataExplorer/ExplorerButtonGroup/index.jsx
+++ b/src/GuppyDataExplorer/ExplorerButtonGroup/index.jsx
@@ -295,14 +295,14 @@ class ExplorerButtonGroup extends React.Component {
     const caseField = this.props.guppyConfig.manifestMapping.referenceIdFieldInDataIndex;
     const caseFieldInFileIndex =
       this.props.guppyConfig.manifestMapping.referenceIdFieldInResourceIndex;
-    this.setState(prevState => ({
-      pendingManifestEntryCountRequestNumber: prevState.pendingManifestEntryCountRequestNumber + 1,
-      manifestEntryCount: 0,
-    }));
     if (this.props.buttonConfig
       && this.props.buttonConfig.buttons
       && this.props.buttonConfig.buttons.some(
         btnCfg => this.isFileButton(btnCfg) && btnCfg.enabled)) {
+      this.setState(prevState => ({
+        pendingManifestEntryCountRequestNumber: prevState.pendingManifestEntryCountRequestNumber + 1,
+        manifestEntryCount: 0,
+      }));
       const caseIDResult = await this.props.downloadRawDataByFields({ fields: [caseField] });
       if (caseIDResult) {
         const caseIDList = caseIDResult.map(i => i[caseField]);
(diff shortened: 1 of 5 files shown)


mpingram:
This change is running in my dev commons at https://mpingram.planx-pla.net/explorer! To really see it, you'll probably need to throttle your internet speed in the dev tools -- there's not a huge number of cases being filtered in my dev commons so the pending state is pretty fast on our wifi.

mpingram:
Note -- I need to revert commits d72c6c9 and 33f925b before merging (ie, revert the Dockerfile changes and pin Guppy to the next released version) but we'll need to merge the Guppy PR first / publish the npm package. Please review this commit after I've made those changes!

src/GuppyDataExplorer/ExplorerButtonGroup/index.jsx
```diff
@@ -295,14 +295,14 @@ class ExplorerButtonGroup extends React.Component {
...
-      manifestEntryCount: 0,
-    }));
     if (this.props.buttonConfig
       && this.props.buttonConfig.buttons
       && this.props.buttonConfig.buttons.some(
         btnCfg => this.isFileButton(btnCfg) && btnCfg.enabled)) {
+      this.setState(prevState => ({
+        manifestEntryCountSemaphore: prevState.manifestEntryCountSemaphore + 1,
+        manifestEntryCount: 0,
+      }));
```
mpingram: The request for a new `manifestEntryCount` is only made inside the if statement on line 298, but the manifestEntryCountSemaphore was updated outside the if statement -- so if the if statement was never executed (this can happen on the File explorer), the semaphore would stay nonzero forever. (This caused a visual bug in the file explorer, where buttons would appear to be pending forever.)

mfshao: wow good catch!

mpingram:
> Only concern is if the aggsDataIsLoading bool in guppy will behave normally if there are councurrent queries. But might not be an issue for now

Yes, I agree! I thought about adding another semaphore to the guppy code similar to the `manifestEntryCountSemaphore` in data-portal, but I decided against it thinking that any race condition bugs resulting from it would only be temporary visual bugs due to the fact that whether a request succeeds or fails it sets `aggsDataIsLoading` to false. Is that accurate? (I was thinking along the lines of, 2 requests go out simultaneously -> request 1 fails and sets `aggsDataIsLoading` to false -> buttons stop being pending, are disabled -> request 2 succeeds and sets `aggsDataIsLoading` to false -> buttons are enabled.)

mfshao:
> > Only concern is if the aggsDataIsLoading bool in guppy will behave normally if there are councurrent queries. But might not be an issue for now
> 
> Yes, I agree! I thought about adding another semaphore to the guppy code similar to the `manifestEntryCountSemaphore` in data-portal, but I decided against it thinking that any race condition bugs resulting from it would only be temporary visual bugs due to the fact that whether a request succeeds or fails it sets `aggsDataIsLoading` to false. Is that accurate? (I was thinking along the lines of, 2 requests go out simultaneously -> request 1 fails and sets `aggsDataIsLoading` to false -> buttons stop being pending, are disabled -> request 2 succeeds and sets `aggsDataIsLoading` to false -> buttons are enabled.)

yes! I think this is acceptable to me

PR #628, opened 2019-12-12
PXP-4533 Fix/stack bar title

mfshao:
This PR fulfills https://ctds-planx.atlassian.net/browse/PXP-4533

Dployed in https://mingfei.planx-pla.net/ check the File Explorer page to see the changes

### Bug Fixes
- Fixed a bug that prevents PercentageStackBarChart from showing its title
- Fixed a bug that prevents PercentageStackBarChart from showing its full x-axis

### Improvements
- Made PercentageStackBarChart more responsive
- Removed some duplicated UI components and import them only from `gen3-ui-component`

### Dependency updates
- Updated `gen3-ui-component` to 0.3.12


src/components/charts/PercentageStackedBarChart/PercentageStackedBarChart.less
```diff
@@ -1,23 +1,50 @@
...
   padding: 12px 12px;
   font-size: 14px;
   line-height: 1em;
   letter-spacing: .03rem;
-  color: #606060;
+  color: var(--g3-color__gray);
 }
 
-.percentage-bar-chart__legend {
+.percentage-bar-chart__title-box {
```
mpingram: What's the reason these styles are duplicated in data portal and gen3-ui-component? src/components/charts/PercentageStackedBarChart/PercentageStackedBarChart.css

mfshao: Pretty sure they are there before we put them into `gen3-ui-component`. And I think only the old data explorer (powered by arranger) are still importing those.

I'll see if I can let the arranger data explorer to import from gen3-ui-component and get rid of those duplicates. But I hope we can get rid of the whole arranger thing soon. 🤦‍♂

PR #631, opened 2019-12-20
(PXP-4883) More descriptive file manifests

mpingram:
Resolves https://ctds-planx.atlassian.net/browse/PXP-4883.

This PR adds the `md5sum`, `file_size`, and `file_name` fields to the data file manifest that can be downloaded from the data explorer and the file explorer.

This change is currently running in my dev commons at https://ctds-planx.atlassian.net/browse/PXP-4883

### New Features
- Added `file_name`, `md5sum`, and `file_size` fields to the data explorer file manifest.


src/GuppyDataExplorer/ExplorerButtonGroup/index.jsx
```diff
@@ -118,7 +129,15 @@ class ExplorerButtonGroup extends React.Component {
       filter.data_format = this.props.filter.data_format;
     }
     let resultManifest = await this.props.downloadRawDataByTypeAndFilter(
-      resourceType, filter, [refFieldInResourceIndex, resourceFieldInResourceIndex],
+      resourceType,
+      filter,
+      [
+        refFieldInResourceIndex,
+        resourceFieldInResourceIndex,
+        fileNameField,
```
qingyashu: if there's no `file_name` or `file_size` field in ES index, will this request fail?

src/GuppyDataExplorer/ExplorerButtonGroup/index.jsx
```diff
@@ -118,7 +129,15 @@ class ExplorerButtonGroup extends React.Component {
...
+        md5Field,
+        fileSizeField,
+      ],
     );
     resultManifest = resultManifest.filter(
       x => !!x[resourceFieldInResourceIndex],
```
mfshao: you probably want also to change logics in here so the manifest downloaded from data tab will also has these new fields

src/GuppyDataExplorer/ExplorerButtonGroup/index.jsx
mpingram: I don't know -- I'll test it now and find out.

src/GuppyDataExplorer/ExplorerButtonGroup/index.jsx
mpingram: Gotchu, I missed this -- Could you tell me what this line is doing? I thought it was removing all elements from the manifest that had a blank `resourceFieldInResourceIndex` property.

src/GuppyDataExplorer/ExplorerButtonGroup/index.jsx
mpingram: Yes, the request fails, and causes an unhandled Promise rejection, so the manifest is not downloaded.

src/GuppyDataExplorer/ExplorerButtonGroup/index.jsx
mfshao: wait... actually you are right. I was getting a manifest without new fields, but this line is irrlevant. Sorry for the false alarm. I'll make this as resolved 🤦‍♂

mpingram: No worries, thank you for looking over my code in such detail!

src/GuppyDataExplorer/ExplorerButtonGroup/index.jsx
mpingram: Update: After talking with @qingyashu we've decided to pre-check the hard-coded file manifest field names against the field names in Guppy's file index. This lets us safely request the new field names without having to make portal config changes.

mpingram: @qingyashu Let me know if the new approach makes sense -- I just wrapped the Guppy call in a try/catch block, and retry the request without the additional fields if it fails. (I experimented with pre-requesting the fields or using the allFields prop from the Guppy wrapper but this seemed simplest)

src/GuppyDataExplorer/ExplorerButtonGroup/index.jsx
```diff
@@ -114,12 +136,27 @@ class ExplorerButtonGroup extends React.Component {
...
-    );
+    let resultManifest;
+    try {
+      resultManifest = await this.props.downloadRawDataByTypeAndFilter(
+        resourceType,
+        filter,
+        [
+          refFieldInResourceIndex,
+          resourceFieldInResourceIndex,
+          ...additionalFields,
```
mpingram: @mfshao Something @svburke brought up, will the extra fields be a breaking change for gen3-client? https://github.com/uc-cdis/cdis-data-client/blob/531982ea94edb98cbd706ab6656e4a7d790690e9/gen3-client/g3cmd/download-multiple.go#L336

mpingram: Update: No, it shouldn't  -- plus, I tested it and gen3-client knew what to do with the new manifests.

src/GuppyDataExplorer/ExplorerButtonGroup/index.jsx
```diff
@@ -97,15 +97,37 @@ class ExplorerButtonGroup extends React.Component {
 
   getManifest = async (indexType) => {
     const refField = this.props.guppyConfig.manifestMapping.referenceIdFieldInDataIndex;
+    const md5Field = 'md5sum';
+    const fileNameField = 'file_name';
+    const fileSizeField = 'file_size';
+    const additionalFields = [md5Field, fileNameField, fileSizeField];
```
mfshao: trivial: since we are hard coding these anyways, seems a bit redundent to me for giving each of them a variable name and then put into array

PR #639, opened 2020-01-28
Fix(PXP-5088): Refactor index page (`/`) to not share state with `/submission` page

mpingram:
This PR resolves https://ctds-planx.atlassian.net/browse/PXP-5088 (Windmill submission page hitting wrong Peregrine endpoint for getting project lists). When `homepageChartNodes` is enabled, users would always see all projects in the /submission page, including projects they didn't have access to.

This issue was caused by the submission page (`Homepage`) and the landing page (`Index`) using the same Redux state to store project counts. This meant that, when a commons shows project counts for ALL projects on the landing page (when `homepageChartNodes` is set in the portal config), the submission page would use the same counts for ALL projects in its list of projects.

This PR resolves the issue by separating the Redux state between the submission page and the landing page. It also duplicates the code used to fetch the data from Peregrine's graphql endpoint (`Index/relayer.js` and `Homepage/relayer.js`), because this code is tightly coupled to the redux store.

Test plan at: https://github.com/uc-cdis/gen3-qa/blob/85f6f4e95f11a64013c45826a53f5c8af6bc1eeb/suites/portal/homepageChartNodesExecutableTestPlan.js

Currently deployed in my dev environment at https://mpingram.planx-pla.net, which has `homepageChartNodes` enabled. I've enabled mock auth in my dev commons, so you can run the test plan by logging in as `dummy-one@planx-pla.net`. (Using the cookie `dev_login: dummy-one@planx-pla.net`, I usually set it in the `Application` tab of Chrome's dev tools). `dummy-one@planx-pla.net` has access to the projects QA-test and DEV-project_A (which is open access) but not DEV-project_B.

### Bug Fixes
- Fixed bug in which, under some conditions, the submission page would appear to show all projects as submittable regardless of user's permissions


src/Index/utils.js
```diff
@@ -48,6 +47,7 @@ const getProjectNodeCounts = async (callback) => {
     case 404:
       // Shouldn't happen, this means peregrine datasets endpoint not enabled
       console.error(`REST endpoint ${datasetUrl} not enabled in Peregrine yet.`);
+      resultStatus.needLogin = true;
```
mpingram: Added a redirect to login page if Peregrine does not have /dataset endpoint.

src/Index/reducers.js
```diff
@@ -0,0 +1,80 @@
...
+    const projectsByName = Object.assign({}, state.projectsByName || {});
+    action.data.projectList.forEach((proj) => {
+      const old = projectsByName[proj.name] || {};
+      projectsByName[proj.name] = Object.assign(old, proj);
+    });
+    const summaryCounts = Object.assign(
+      {}, state.summaryCounts || {}, action.data.summaryCounts,
+    );
+    const lastestListUpdating = Date.now();
+    // const { error, ...state } = state;
```
ZakirG: (is this comment important)

PR #653, opened 2020-02-19
fix(branch): add search to filter and some more code fixes

qingyashu:
### New Features
 - Use newer library version to add search options in explorer's filter

### Improvements
  - Stop rendering invisible component to reduce heavy calculation and boost UI


mfshao:
closing since it is outdated

PR #657, opened 2020-02-24
[PXP-4583] Fix program/project submission UI

vpsx:
Related sheepdog PR https://github.com/uc-cdis/sheepdog/pull/319 
Related commons-users PR https://github.com/uc-cdis/commons-users/pull/810

Fixes submission UI in the program/project submission case. Previously the ArboristUI logic did not differentiate between when a user was submitting program/project nodes and when they were submitting other nodes. CRUD on program/project nodes requires different authz; pre-centralized auth it required `admin: true` in the jwt; post-centralized auth it requires the `services/sheepdog/submission/[program,project]` resources. This PR changes ArboristUI so that it renders the submission UI components ("upload file" button, "use form submission" toggle) on program/project creation only when the user has permissions on the new resources. When ArboristUI is not enabled, the submission UI is always shown regardless of the user's permissions.

Also adds TECHDEBT.md per discussion in Engineering Retro a while ago. (To clarify, the thing described in the document was not introduced here.)

Tested on dev env:
- Old per-project ArboristUI still works as expected (submission UI appears only on projects where I have create or update)
- Program/project ArboristUI works as expected (for program and project submission pages, submission UI only appears if I have sheepdog program/project admin respectively)
- Turning off Arborist UI still works (submission UI will appear)

nodeType fix reverted because more issues were found. The bug is unrelated to this PR. Opening a separate ticket for it.

### New Features
Implement ArboristUI for program/project submission

### Improvements
Adds TECHDEBT.md so we can track tech debt

### DIFF
--- a/TECHDEBT.md
+++ b/TECHDEBT.md
@@ -1,0 +1,37 @@
+# Tech debt
+
+### Arborist auth mapping
+Observed: October 2019
+Impact: (if this tech debt affected your work somehow, add a +1 here with a
+date and optionally a note)
++1 Zoe 2020 Feb 20 This is an example of a +1
+##### Problem:
+The ArboristUI features rely on hitting Arborist /auth/mapping one time to get
+the auth mapping, and then consult the auth mapping to decide which components
+should be rendered and which should not.
+##### Why it was done this way:
+This was done to avoid having to make many (tens?) of calls to Arborist upon
+each page load, for example upon loading the profile page.
+##### Why this way is problematic:
+The auth mapping returns a bunch of resource paths, and the resource paths are
+part of a hierarchy. When Arborist resolves auth requests it understands the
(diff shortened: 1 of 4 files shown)


vpsx:
Did you mean you should be able to create a _program_ (not project) with those permissions, while in the UI you can't? If that's the case then yes, that is indeed the expected behavior 😄 
I can put it in deployment changes but actually right now, if you have arboristUI on, it's already the case that the submission UI won't show up on root/program pages--whether you are an admin or not 😛 so it wouldn't really make sense to do that. It _is_ in the deployment changes for the Sheepdog PR. Which I should have linked... will go do that now

vpsx:
@mfshao Thank you for the review 🙇 🙇‍♀

PR #660, opened 2020-03-05
(PXP-5571) Allow contact to be a link on login page

mpingram:
https://ctds-planx.atlassian.net/browse/PXP-5571

Add a new configurable contact link to the login page, which can be used in place of the contact email. This allows us to specify email links (by using `mailto:email` as the link) as well as normal URLs. This is needed for BDCat as we want to use https://biodatacatalyst.nhlbi.nih.gov/contact as the Login page contact form.

Deployed in https://mpingram.planx-pla.net/login -- try using the contact link.

### New Features
- Allow use of any link for a contact on the Login page, rather than only emails.

### Deployment changes
- Added a new property to gitops.json: `components.login.contact_link: { text: string, href: string }`. If `login.contact_link` is present, it will be displayed instead of `login.email`. Email links can be represented in `contact_link` using "mailto:email@address" as the "href".

### DIFF
--- a/src/Login/Login.jsx
+++ b/src/Login/Login.jsx
@@ -17,27 +17,6 @@ const getLoginUrl = (providerLoginUrl, next) => {
 };
 
 class Login extends React.Component {
-  static propTypes = {
-    providers: PropTypes.arrayOf(
-      PropTypes.objectOf(PropTypes.any),
-    ),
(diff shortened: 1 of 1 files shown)


mpingram:
Thanks for the r! I'll make `login.contact_link.text` optional and have it take the value of `login.contact_link.href` if it's not specified.

PR #666, opened 2020-04-07
Multi-commons Workspace: login buttons POC

paulineribeyre:
### New Features
- Multi-commons Workspace: login buttons POC


paulineribeyre:
Note that this is a POC and the WorkspaceLogin.jsx code will probably be replaced later when we have a real design for this. We need to merge it so that we can deploy it for demos without pinning data-portal to a branch

paulineribeyre:
relates to https://github.com/uc-cdis/workspace-token-service/pull/19

mfshao:
tried it with some commons and it looks cool. Only question is since it is a POC for a demo do we want to merge it back to `master` and release it to other commons in the future? or should we just tag this branch and make a temp release?

paulineribeyre:
nothing will be change visually if the WTS is not configured with extra OIDC clients. but we can definitely just tag it if you want

mfshao:
> nothing will be change visually if the WTS is not configured with extra OIDC clients. but we can definitely just tag it if you want

yeah i test it locally against your dev env (that is sneaky sorry) and the those buttons are good.

mfshao:
> nothing will be change visually if the WTS is not configured with extra OIDC clients. but we can definitely just tag it if you want

yeah i test it locally against your dev env (that is sneaky sorry) and the those buttons are good.

for uc-cdis/workspace-token-service#19 , does the `login_options` in `external_oidc` created automatically as they have been configured in the fence config file? or you can specify in which login options in fence config will be included in the `external_oidc` of wts?

paulineribeyre:
> does the login_options in external_oidc created automatically as they have been configured in the fence config

you have to specify which login options you want manually. listing them automatically was adding too much complexity for a POC. Also right now the portal code can handle multiple URLs for the same login option (display a dropdown) but the WTS can't. So for incommon you'd have to list the options you want manually for now

mfshao:
> > does the login_options in external_oidc created automatically as they have been configured in the fence config
> 
> you have to specify which login options you want manually. listing them automatically was adding too much complexity for a POC. Also right now the portal code can handle multiple URLs for the same login option (display a dropdown) but the WTS can't. So for incommon you'd have to list the options you want manually for now

oh cool! thanks for the explanation! just want to make sure if we merge this and when the new releases has rolled out these new buttons will not be showing up automatically on any prod envs if we haven't configured it intentionally 😃

PR #669, opened 2020-04-16
Navbar tooltip on hover

paulineribeyre:
Screen Shot 2020-04-16 at 5 42 43 PM
Jira Ticket: COV-98 (OCC Jira)
Allows us to display tooltips on hover instead of (or in addition to 🤷) the usual homepage cards

```
"navigation": {
  "items": [
    {
      "name": "Notebook Browser",
      "link": "/resource-browser",
      "icon": "query",
      "tooltip": "The notebooks pull data from various external sources to generate and output useful tables, charts, graphs, and models."
    }
  ]
}
```

Currently deployed at https://pauline.planx-pla.net

\+ add `tooltip` to the default config, and remove `color` because as far as i can tell it's not used anywhere... let me know if i shouldn't
\+ small fix for a 5px empty space displayed under navbar buttons when the text is a bit long
\+ add padding around the Resource Browser description

### New Features
- Navigation bar display a tooltip on hover for each button

PR #671, opened 2020-04-17
fix menu responsive CSS

paulineribeyre:
My previous change https://github.com/uc-cdis/data-portal/pull/669/commits/e588a0cfab76206c9dfa6e1cda4d2774fa7c2129 fixed a CSS issue but created a new one on mobile screens:

Screen Shot 2020-04-17 at 4 45 33 PM

This change is to fix it:

Screen Shot 2020-04-17 at 4 48 32 PM

The original change was to fix an annoying extra empty space under the navbar buttons - it turns out we just need to reduce the `.nav-bar__logo` bottom padding, and I need to remove the CSS override in covid19's `gitops.css`

### Bug Fixes
- Fix responsive main menu

### DIFF
--- a/src/components/layout/NavBar.less
+++ b/src/components/layout/NavBar.less
@@ -114,7 +114,7 @@
 }
 
 .nav-bar__logo {
-  padding-top: 8px;
+  padding: 8px 0;
   display: inline-block;
 }
 
--- a/src/components/layout/NavBar.less
+++ b/src/components/layout/NavBar.less
@@ -7,7 +7,6 @@
 .nav-bar__nav--info {
   float: left;
   display: inline-flex;
-  height: 0;
 }
 
 @media screen and (max-width: 1090px) {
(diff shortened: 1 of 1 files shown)


src/components/layout/NavBar.less
```diff
@@ -115,7 +114,7 @@
 }
 
 .nav-bar__logo {
-  padding: 8px 0;
+  padding-top: 8px;
```
mfshao, line 117: i'm not sure about this because i'm in favor of keeping an equal padding on top and buttom. Maybe when it shrinks into media width we'd better add some boarders around those nav buttons to clearly indicate the padding vs the button area?

sorry just nitpicking about the details. i wish we have a designer 😥

paulineribeyre: then i can keep the original line 👍 and i can override it in gitops.css if needed. it looks like most commons are overriding this anyway :/

PR #674, opened 2020-04-20
env var mapboxApiAccessToken

diw2:
add MapboxAccessToken env var to portal from cloud-automation


webpack.config.js
```diff
@@ -36,6 +36,7 @@ const plugins = [
   new webpack.EnvironmentPlugin(['WORKSPACE_URL']),
   new webpack.EnvironmentPlugin(['WTS_URL']),
   new webpack.EnvironmentPlugin(['MANIFEST_SERVICE_URL']),
+  new webpack.EnvironmentPlugin(['MapboxAccessToken']),
```
mfshao, line 39: can we change the name of this env var to sth like `MAPBOX_ACCESS_TOKEN` so all env vars have similar format?

diw2: https://github.com/uc-cdis/cloud-automation/pull/1200/files FYI

runWebpack.sh
```diff
@@ -160,6 +160,7 @@ if [[ "$NODE_ENV" == "dev" || "$NODE_ENV" == "auto" ]]; then
   echo ./node_modules/.bin/webpack-dev-server
   ./node_modules/.bin/webpack-dev-server
 else
+  export NODE_OPTIONS='--max-old-space-size=4096'
```
paulineribeyre, line 163: what is this?

mfshao: this is to allow node to allowcate more mem (default is 512MB i think). 

we have this in the Dockerfile but don't know why not here. And portal is so fat that it need this to spin up...

paulineribeyre: ok, is this related to covid19? if not, can we make a separate PR to merge this to master? this PR will only be merged to branch feat/covid19dashboard for now

mfshao: yeah portal on this branch won't start up if don't have this 😞 

i guess by adding more into portal eventually the master branch will gonna need this one day

paulineribeyre: just adding 1 environment variable triggered this?? 😱

mfshao: it sounds unrealistic but 😕

paulineribeyre:
> could we add a warning at startup when MAPBOX_API_TOKEN is not set?

sorry, changed my mind 🤦‍♀️it shouldn't be at startup because commons that don't use map should not get a warning all the time haha. We can put the warning in the covid19dashboard/index.jsx file instead

PR #677, opened 2020-04-22
Fix covid19 map in safari

atharvar28:
This PR consists of changes needed to fix the dashboard map in Safari for qa-covid19 env 

the Initial error was -> `Refused to connect to https://www.google-analytics.com/...... because it does not appear in the connect-src directive of the Content Security Policy.` 

The changes here is ->
adding `child-src blob:` in Content Security Policy code block in response to this error -> `[Error] Refused to load blob:https://qa-covid19.planx-pla.net/89835289-eb1a-4867-ab55-0ecf1c5f61cc because it appears in neither the child-src directive nor the default-src directive of the Content Security Policy.`

Update -> the dashboard map works without adding `https://www.google-analytics.com` to the connect-src directive (tested on qa-covid19)

JIRA ticket -> https://occ-data.atlassian.net/browse/COV-58


dev.html
```diff
@@ -5,8 +5,9 @@
     <!-- Content-Security-Policy connect-src:
       - https://*.mapbox.com for COVID-19 Dashboard maps
       - https://opendata.datacommons.io to fetch COVID-19 Dashboard data
+      - https://www.google-analytics.com for the dashboard map
```
paulineribeyre, line 8: can you please check if it works without adding `https://www.google-analytics.com`? since apparently the issue was something else. and if we do need it, can you update the comment in both files with a bit more detail - `for COVID-19 Dashboard maps in Safari` or something

atharvar28: I will try doing that

src/index.ejs
```diff
@@ -5,8 +5,9 @@
     <!-- Content-Security-Policy connect-src:
       - https://*.mapbox.com for COVID-19 Dashboard maps
       - https://opendata.datacommons.io to fetch COVID-19 Dashboard data
+      - https://www.google-analytics.com for the dashboard map
```
paulineribeyre, line 8: remove comment

dev.html
paulineribeyre, line 8: remove comment

PR #679, opened 2020-04-23
fix: invalid presigned URL issue

ZakirG:
### Bug Fixes
- Fixes an issue in the indexing GUI where the spinner spins forever if the presigned URL has an incorrect request signature

### DIFF
--- a/src/actions.js
+++ b/src/actions.js
@@ -128,6 +128,31 @@ export const fetchWithCreds = (opts) => {
   );
 };
 
+export const fetchWithCredsAndTimeout = (opts, timeoutInMS) => {
+  let didTimeOut = false;
+
+  return new Promise(((resolve, reject) => {
+    const timeout = setTimeout(() => {
+      didTimeOut = true;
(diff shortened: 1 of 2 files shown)


ZakirG:
A screenshot of this code change in action in qa-dcf:

Screen Shot 2020-04-23 at 3 29 52 PM

src/actions.js
```diff
@@ -128,6 +128,31 @@ export const fetchWithCreds = (opts) => {
...
+      }
+    })
+      .catch((err) => {
+        // Rejection already happened with setTimeout
+        if (didTimeOut) return;
+        // Reject with error
+        reject(err);
+      });
+  }));
+};
```
ZakirG, line 154: i copied this function verbatim from the NDE https://github.com/uc-cdis/data-ecosystem-portal/blob/master/src/actions.js#L125

ZakirG: so dont worry too much about reading it =P

PR #680, opened 2020-04-28
Add value attribute to indexing GUI buttons

ZakirG:
Deployed in https://zakir.planx-pla.net/indexing

### Improvements
- Added value attributes to the download buttons in the indexing GUI that contain links to the target files


ZakirG:
Screen Shot 2020-04-28 at 1 22 42 PM
 ^ this is what the change looks like. just an extra "value" attribute on the button set to the file link that the button produces

PR #682, opened 2020-05-06
Feat/explorerlink

paulineribeyre:
Screen Shot 2020-05-06 at 5 49 34 PM

Reused code written for the NDE dataset browser (see https://github.com/uc-cdis/data-ecosystem-portal and https://niaiddata.org/datasets), except a bit more general to allow configuring multiple fields as links

### New Features
- Ability to display clickable buttons in an Explorer table ("explorerConfig.table.linkFields" configuration variable)

### DIFF
--- a/docs/multi_tab_explorer.md
+++ b/docs/multi_tab_explorer.md
@@ -36,8 +36,10 @@ An example of this new `explorerConfig` is (some contents are omitted for concis
         "enabled": true,
         "fields": [
           "project_id",
+          "url",
           ...
-        ]
+        ],
+        "linkFields": [ "url" ]
       },
       "guppyConfig": {
         "dataType": "case",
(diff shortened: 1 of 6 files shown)


paulineribeyre:
> the button size feels unproportional if there is not too much texts

@mfshao true, maybe less padding would help

> maybe better to change the CSS that is rendering table cells so the content is vertically centered

for the example in the PR description it would be fine, but with more text like here https://niaiddata.org/datasets i don't think it'd look good

mfshao:
> > the button size feels unproportional if there is not too much texts
> 
> 
> 
> @mfshao true, maybe less padding would help
> 
> 
> 
> > maybe better to change the CSS that is rendering table cells so the content is vertically centered
> 
> 
> 
> for the example in the PR description it would be fine, but with more text like here https://niaiddata.org/datasets i don't think it'd look good
> 
> 
> 
> it looks a bit odd when not every row has a button, but reducing the size of the button can help

I don’t remember if we tried vertical center the text for NDE. Maybe it would work?
For the button a better way would be to give it a maximum height. And if the cell height is less it just shrinks?

PR #684, opened 2020-05-11
fix: use date + time instead of just date

mfshao:
Fixed an error caused by parsing Date incorrectly, which leads to a blank homepage

When creating a Date with no time values, the computed date will be relative to browser's time zone. Which means `new Date("2020-05-10")` will yield `Sat May 09 2020 19:00:00 GMT-0500 (Central Daylight Time)`, in here `2020-05-10` is be treated as a UTC time.

But if a time is giving in date/time pair, the computed date woill NOT do time zone conversion under the table, so `new Date("2020-05-10T00:00:00")` yields `Sun May 10 2020 00:00:00 GMT-0500 (Central Daylight Time)`

Ref: https://www.w3schools.com/js/js_date_formats.asp
https://stackoverflow.com/a/38050824

### Bug Fixes
- Fixed an error caused by parsing Date incorrectly, which leads to a blank homepage


mfshao:
deployed in https://qa-covid19.planx-pla.net/ for testing

PR #685, opened 2020-05-12
fix: use only filtered data for maps

mfshao:
Deployed in https://qa-covid19.planx-pla.net for testing

### Bug Fixes
- Use only `open-JHU` data to calculate `selectedDate`

### Improvements
- Pass filtered data to child map components in props

### DIFF
--- a/src/Covid19Dashboard/index.jsx
+++ b/src/Covid19Dashboard/index.jsx
@@ -66,16 +66,7 @@ class Covid19Dashboard extends React.Component {
   getTotalCounts() {
     // find latest date we have in the data
     let selectedDate = new Date();
-    if (this.props.rawData.length > 0) {
-      selectedDate = new Date(Math.max.apply(
-        null, this.props.rawData[0].date.map((date) => {
-          if (date.includes('T')) {
-            return new Date(date);
-          }
-          return new Date(date.concat('T00:00:00'));
-        }),
-      ));
-    }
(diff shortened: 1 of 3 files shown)


src/Covid19Dashboard/index.jsx
```diff
@@ -215,7 +209,7 @@ class Covid19Dashboard extends React.Component {
                 />
               </div>
               <div className='covid19-dashboard_visualizations'>
-                <WorldMapChart {...this.props} />
+                <WorldMapChart rawMapData={this.props.rawData.filter(location => (location.project_id === 'open-JHU' && (location.confirmed.length || location.deaths.length)))} />
```
paulineribeyre, line 212: what do you need `&& (location.confirmed.length || location.deaths.length)` for? 😄

paulineribeyre: nvm, i see why. probably should also be checked for the illinois map, but whatever

mfshao: this logic https://github.com/uc-cdis/data-portal/pull/685/files#diff-891735d0760d4f7e2be89d9467b3ba0fL266 was filtering out no data locations. I'm just moving this into here

mfshao: > nvm, i see why. probably should also be checked for the illinois map, but whatever

i'll leave this task to you :smirk:

PR #686, opened 2020-05-14
(PXP-5597) Export to Seven Bridges

mpingram:
Adds Export to Seven Bridges button to explorer, which allows user to export a PFB to Seven Bridges' analysis platform. This feature is configured with the `export-to-seven-bridges` button type and the `sevenBridgesExportURL` parameter in the portal config. 

### New Features
- Added Export to Seven Bridges button, allowing user to export a PFB to Seven Bridges.

### Dependency updates
- Adds new button type `export-to-seven-bridges` to portal config `dataExplorerConfig.buttons.` Example:
```
      {
        "enabled": true,
        "type": "export-to-seven-bridges",
        "title": "Export All to Seven Bridges",
        "rightIcon": "external-link"
      },
``` 
- Adds new parameter `dataExplorerConfig.sevenBridgesExportURL` to portal config, which specifies the URL belonging to Seven Bridges that portal should redirect to. Portal appends the signed url to this URL with a query param (`&PFB=<signed url>`). Example:
```
 "sevenBridgesExportURL": "https://windmill.sbgenomics.com/import/windmill"
```

### DIFF
--- a/src/GuppyDataExplorer/ExplorerButtonGroup/index.jsx
+++ b/src/GuppyDataExplorer/ExplorerButtonGroup/index.jsx
@@ -24,7 +24,7 @@ class ExplorerButtonGroup extends React.Component {
       toasterHeadline: '',
       toasterError: null,
       toasterErrorText: 'There was an error exporting your cohort.',
-      exportingToCloud: false,
+      exportingToTerra: false,
       exportingToSevenBridges: false,
       // for export to PFB
       exportPFBStatus: null,
@@ -44,11 +44,11 @@ class ExplorerButtonGroup extends React.Component {
     if (nextProps.job && nextProps.job.status === 'Completed' && this.props.job.status !== 'Completed') {
       this.fetchJobResult()
         .then((res) => {
-          if (this.state.exportingToCloud) {
+          if (this.state.exportingToTerra) {
             this.setState({
               exportPFBURL: `${res.data.output}`.split('\n'),
               toasterOpen: false,
-              exportingToCloud: false,
+              exportingToTerra: false,
             }, () => {
               this.sendPFBToTerra();
             });
@@ -56,7 +56,7 @@ class ExplorerButtonGroup extends React.Component {
             this.setState({
               exportPFBURL: `${res.data.output}`.split('\n'),
               toasterOpen: false,
-              exportingToCloud: false,
+              exportingToSevenBridges: false,
             }, () => {
               this.sendPFBToSevenBridges();
             });
@@ -317,7 +317,7 @@ class ExplorerButtonGroup extends React.Component {
   // ==========================================
 
   exportToTerra = () => {
-    this.setState({ exportingToCloud: true }, () => {
+    this.setState({ exportingToTerra: true }, () => {
       this.exportToPFB();
     });
   };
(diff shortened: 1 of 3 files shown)


src/GuppyDataExplorer/index.jsx
```diff
@@ -91,6 +91,7 @@ class Explorer extends React.Component {
               dropdowns: explorerConfig[this.state.tab].dropdowns,
               terraExportURL: explorerConfig[this.state.tab].terraExportURL,
               terraTemplate: explorerConfig[this.state.tab].terraTemplate,
+              sevenBridgesExportURL: explorerConfig[this.state.tab].sevenBridgesExportURL,
```
mfshao, line 94: i'm start think manybe we should check if these optional value exists before enabling corresponding buttons, but might be out of the scope of your ticket 🤔

mpingram: That makes sense -- I'll see if there's a simple way I can add that. What do you think of adding a check when portal reads the config, so portal fails to build if the config has the terra / sbg export buttons but is missing the `sevenBridgesExportURL` / `terraExportURL` parameters

src/GuppyDataExplorer/ExplorerButtonGroup/index.jsx
```diff
@@ -449,7 +473,7 @@ class ExplorerButtonGroup extends React.Component {
     if (buttonConfig.type === 'export-to-pfb') {
       return !this.state.exportingToCloud;
     }
-    if (buttonConfig.type === 'export') {
+    if (buttonConfig.type === 'export' || buttonConfig.type === 'export-to-seven-bridges') {
       // if exportingToCloud is true or the PFB job is running, the button is not enabled.
       return !(this.state.exportingToCloud || this.isPFBRunning());
```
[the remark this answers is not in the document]
mpingram: Good catch, thank you!

src/GuppyDataExplorer/ExplorerButtonGroup/index.jsx
```diff
@@ -49,7 +50,15 @@ class ExplorerButtonGroup extends React.Component {
...
               exportingToCloud: false,
             }, () => {
-              this.sendPFBToCloud();
+              this.sendPFBToTerra();
+            });
+          } else if (this.state.exportingToSevenBridges) {
+            this.setState({
+              exportPFBURL: `${res.data.output}`.split('\n'),
+              toasterOpen: false,
+              exportingToCloud: false,
```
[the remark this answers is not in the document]
mpingram: Yes, another good catch. Thank you Mingfei!

src/GuppyDataExplorer/index.jsx
mfshao: fails to build sounds a little bit too harsh to me. Personly I think throwing an error in the console would be enough. Even better if we can disable the button

mpingram: Gotchu. My reasoning with fail to build was that we would catch configuration errors during our PR checks. I'll plan on adding an error check to this code, logging an error to the console, and disabling the button.

mfshao: yeah i'm totally with you on that. We have the sanity-check.js which can bail out if sth is not right. But we also have places only throws errors. I guees both are acceptable, just the first option sounds more work on your plate to me 😛

mpingram: Haha that sounds good. Thanks again for the R!

src/GuppyDataExplorer/ExplorerButtonGroup/index.jsx
```diff
@@ -447,11 +471,30 @@ class ExplorerButtonGroup extends React.Component {
       return this.state.manifestEntryCount > 0;
     }
     if (buttonConfig.type === 'export-to-pfb') {
-      return !this.state.exportingToCloud;
+      // disable the pfb export button if any other pfb export jobs are running
+      return !(this.state.exportingToTerra || this.state.exportingToSevenBridges);
```
mfshao, line 475: shoudl we also check for `isPFBRunning` here? too many button types 🤯

mpingram: Haha yeah, I find this code hard to follow. I'll be honest that I don't fully understand it, but the original code didn't check for `isPFBRunning` and as far as I can tell the `isPending` check for export-to-pfb uses `isPFBRunning() == true` to decide if the button is pending (https://github.com/uc-cdis/data-portal/pull/686/files/67fe2b01dd812ce6a929815eebb95f403beaf83d#diff-a6734b941e5a6867b94a16ee82e40e96L471). So I thought it might have been intentional, so that the button is pending but not disabled when isPFBRunning() == true and we're not exporting to Terra (or seven bridges)

mfshao: yeah it was not in there in the first place, i'm ok with that 👍

PR #687, opened 2020-05-15
fix(dev-server): allow more space to run webpack-dev-server

bobaekang:
This PR allows `webpack-dev-server` to launch successfully without V8 crashing due to JavaScript heap out of memory error.

An alternative to this is adding `NODE_OPTIONS="--max-old-space-size=4096"` to the command when spinning up the dev server. While these two options are functionally equivalent and address the same issue, modifying the build script can lead to a more consistent developer experience in my view.

### Bug Fixes
* Set `--max-old-space-size=4096` in `/runWebpack.sh` when launching `webpack-dev-server`

### DIFF
--- a/runWebpack.sh
+++ b/runWebpack.sh
@@ -157,6 +157,7 @@ export REACT_APP_DISABLE_SOCKET=true
 # run webpack to generate a static bundle.js
 #
 if [[ "$NODE_ENV" == "dev" || "$NODE_ENV" == "auto" ]]; then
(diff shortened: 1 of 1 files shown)


ocshawn:
Sorry in the delay in getting to this PR, is this still an issue?

bobaekang:
@ocshawn Thank you for getting back to me. TBH, I forgot about this PR and the issue is probably not relevant at this point :sweat_smile:. I'll close the PR.

PR #690, opened 2020-06-04
COV-274 Public data explorer

mfshao:
This PR fulfills https://occ-data.atlassian.net/browse/COV-274

Added an optional flag `explorerPublic` in portal config `featureFlags`, e.g.:
>   "featureFlags": {
>     "explorer": true,
>     "explorerPublic": true,
>   }

If set to `true`, the data explorer page would be treated as a public component and can be accessed without login.
Data explorer page would be public accessiable if 1. tiered access level is set to `libre` OR 2. this new `explorerPublic` flag is set to `true`.
Tiered access logics (`libre`, `regular` or `private`) that are set for the env would still apply even if the data explorer page is public.
Backend service should have Guppy 0.4.2+ or 2020.05+ so it can handle user without JWT.

Deployed in https://mingfei.planx-pla.net/

### Improvements
- Added an optional flag `explorerPublic` in portal config `featureFlags`, if set to `true`, data explorer page would be public accessible 

### Dependency updates
- Backend service should have Guppy 0.4.2+ or 2020.05+ so it can handle user without JWT

### DIFF
--- a/docs/portal_config.md
+++ b/docs/portal_config.md
@@ -153,7 +153,8 @@ Below is an example, with inline comments describing what each JSON block config
   },
   "requiredCerts": [], // optional; do users need to take a quiz or agree to something before they can access the site?
   "featureFlags": { // optional; will hide certain parts of the site if needed
-    "explorer": true // required; indicates the flag and whether to hide it or not
+    "explorer": true, // required; indicates the flag and whether to hide it or not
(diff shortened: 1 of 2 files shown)


mfshao:
> wow, this was a way faster review than i expected 😆
> but i do think this deserves an update to the documentation with what you explained in the PR description

Ok i updated the `portal_config.md` doc with this new flag.
I'd like to rewrite that doc soon (I'll try to find some time for that) to reorg it so that each section of portal config get its own chapter, and mabe a small summary table indicating which option is for what, instead of puting everything together in a big code chunk with inline comments (I kind of hate this 😞 ).

PR #693, opened 2020-06-11
Replace dev.planx-pla.net with qa.planx-pla.net

paulineribeyre:
Fix travis failures: Replace dev.planx-pla.net with qa.planx-pla.net because the dev namespace will be shut down every night, while the QA one will not


paulineribeyre:
yeah, even keeping all the configs in /data dir doesn't really make sense imo. Everyone is using gitops.json now. But that's a discussion for later :p Just unblocking travis for now

PR #694, opened 2020-06-11
Feat/indexing gui improvements

ZakirG:
This PR addresses https://ctds-planx.atlassian.net/browse/PXP-6068

### New Features
- Adds a "Download Manifest" button to the success panel of the indexing flow at /indexing.
- Adds a "Download Logs" button to the success panel of the generate manifest flow at /indexing if logs are available.


src/Indexing/Indexing.jsx
```diff
@@ -423,6 +427,23 @@ class Indexing extends React.Component {
...
+    const downloadManifestSuccessButtons = [];
+    if (this.state.downloadManifestLogsLink) {
+      downloadManifestSuccessButtons.push({
+        caption: 'Download Logs',
+        icon: 'download',
+        value: this.state.downloadManifestLogsLink,
+        fn: () => this.downloadJobOutput(this.state.downloadManifestLogsLink),
+      });
+    }
+    downloadManifestSuccessButtons.push({
```
mfshao, line 439: should we check for `if (this.state.downloadManifestManifestLink)` here?

src/Indexing/Indexing.jsx
```diff
@@ -228,11 +228,13 @@ class Indexing extends React.Component {
         thisPointer.retrieveJobOutput(uid).then((resp) => {
           if (resp.data && resp.data.output) {
             const logsLink = resp.data.output.split(' ')[0];
+            const manifestLink = resp.data.output.split(' ')[1];
             thisPointer.setState({
               indexingFilesStatus: 'success',
               indexingFilesStatusLastUpdated: thisPointer.getCurrentTime(),
               indexingFilesPopupMessage: 'Done',
               indexingFilesLogsLink: logsLink,
+              indexingFilesManifestLink: manifestLink,
```
mfshao, line 237: does this state got set anywhere in the code?

mpingram:
Is there a way to download the logs if the job fails? It seems like the download logs button is only available if the indexing job succeeds(?). I think the most common use case for checking logs would be if the job fails and I want to know why.

src/Indexing/Indexing.jsx
ZakirG: I don't think so. The reason I added the check for logs is because Giang hasn't written code for the logs output for this generate manifest flow yet. That download manifest button already should not be displayed at all if the job was unsuccessful.

src/Indexing/Indexing.jsx
ZakirG: It is not in the constructor no -- it's only set here on this line you commented. Do you want it to be initialized to the empty string in the constructor, is that why you ask?

ZakirG:
> Is there a way to download the logs if the job fails? It seems like the download logs button is only available if the indexing job succeeds(?). I think the most common use case for checking logs would be if the job fails and I want to know why.

Ah, I agree with you. But no -- the way that sower job is written, there is no logs output if the job fails, just a traceback. And so if the job fails, there will be no output generated. The logs are an output of the job, and my GUI considers the job to have failed if there is no output.

src/Indexing/Indexing.jsx
mfshao: cool!

src/Indexing/Indexing.jsx
```diff
@@ -266,9 +268,11 @@ class Indexing extends React.Component {
         thisPointer.retrieveJobOutput(uid).then(
           (resp) => {
             if (resp.data && resp.data.output) {
-              const logsLink = resp.data.output.split(' ')[0];
+              const manifestLink = resp.data.output.split(' ')[0];
+              const logsLink = resp.data.output.split(' ')[1];
```
mfshao, line 272: i'm a bit of paranoid here. could it be possible that the sower job only returns 1 link? so this `[1]` would go out of bounds?

ZakirG: Yes, this is the current situation in qa-dcp. The sower job is only returning one link. But this code works correctly ~

PR #695, opened 2020-06-11
Feat/pxp 4796 404

scintillating7:
Portal now can redirect to a `NotFound` page for meaningless URLs.

Portal will check if the entered URL is:
1. some predefined meaningful URL (`/DD`, `/explorer`, `/_root`, etc...)
2. a program that exists in sheepdogs DB via peregrine query
3. a project that exists in sheepdogs DB and user has access to via peregrine query

If `ArboristUI` has been enabled, portal will further check if the entered URL is:
1. a root URL and user has sheepdog program admin
2. a valid program URL and user has sheepdog project admin

If none of above conditions is met, portal redirects user to `/not-found`

### Improvements
- Checks user permissions on project before routing to `ProjectSubmission` component. If that fails (either project is invalid or permissions aren't valid), direct to NotFound.
- Checks if file with a given GUID exists before routing to `CoreMetadataPage` component. If that fails, direct to NotFound.
- Deal with trailing slash and duplicated slashes in URL when routing
- Use exact matched for most of router paths

### Note
~This feature requires both config variables `useArboristUI` and `showArboristAuthzOnProfile` to be set to `true`. Otherwise authorization is not fetched (see [here](https://github.com/uc-cdis/data-portal/blob/774c18df9f412df863e9eab6005f415703ae8a2f/src/actions.js#L484)). For local/dev development,~

~1. adding those variables to your `[user].planx-pla.net/portal/gitops.json`~
~2. logging in~

~will enable correct routing to projects that you do have access to. I believe it's outside the scope of this PR to address whether we want these arborist variables to be the gateway for authorization.~

(Updated by @mfshao)
The redirect to NotFound logic will check against the auth premissions in the `authMapping` returned from Arborist but it will also check to see if a program/project exists in sheepdog via querying peregrine. The reason why we cannot rely complete on `authMapping` was described in https://github.com/uc-cdis/data-portal/blob/master/TECHDEBT.md


src/components/NotFound.less
```diff
@@ -0,0 +1,6 @@
+.error-placeholder__error-msg {
+    width: 100%;
+    height: 100%;
+    text-align: center;
+    margin-bottom: 40px;
```
mfshao, line 5: i think you could remove this `margin-bottom` and add `min-height: calc(100vh - 242px);` to put the footer close to the bottom of the page

src/authMappingUtils.js
```diff
@@ -76,3 +77,17 @@ export const userHasMethodOnAnyProject = (method, userAuthMapping = {}) => {
   const hasMethod = actionArrays.some(x => x.some(actionHasMethod));
   return hasMethod;
 };
+
+export const shouldDisplaySubmissionUIComponents = (project, userAuthMapping) => {
+  if (
+    !config.useArboristUI
```
mfshao, line 83: i would prefer not to import `config` into here, but rather move this `!config.useArboristUI` condiction out to `index.jsx`

`index.jsx` is already been importing a bunch of configs through `localconf` and `useArboristUI` can be add to them

src/authMappingUtils.js
```diff
@@ -76,3 +76,17 @@ export const userHasMethodOnAnyProject = (method, userAuthMapping = {}) => {
...
   return hasMethod;
 };
+
+export const shouldDisplaySubmissionUIComponents = (project, userAuthMapping, useArboristUI) => {
+  if (
+    !useArboristUI
+    || (isRootUrl(project) && userHasSheepdogProgramAdmin(userAuthMapping))
+    || (isProgramUrl(project) && userHasSheepdogProjectAdmin(userAuthMapping))
+    || userHasMethodOnProject('create', project, userAuthMapping)
+    || userHasMethodOnProject('update', project, userAuthMapping)
```
mfshao, line 86: lets also add `userHasMethodOnProject('read', project, userAuthMapping)` in here. If user has `read` premission then the user shuold see the nodes but connot submit

src/Submission/ProjectSubmission.jsx
```diff
@@ -29,33 +30,28 @@ const ProjectSubmission = (props) => {
...
+      props.project,
+      props.userAuthMapping,
+      useArboristUI,
+    ) ?
+      <div className='project-submission'>
+        <h2 className='project-submission__title'>{props.project}</h2>
+        {
+          <Link className='project-submission__link' to={`/${props.project}/search`}>browse nodes</Link>
+        }
+        {
```
mfshao, line 45: here we need to check for `userHasMethodOnProject('create', project, userAuthMapping) || userHasMethodOnProject('update', project, userAuthMapping)` in order to determine whether we should render the submission component

src/authMappingUtils.js
```diff
@@ -77,16 +77,35 @@ export const userHasMethodOnAnyProject = (method, userAuthMapping = {}) => {
...
+    || Object.values(projectList).includes(project)
+    );
   }
-
-  return false;
+  // if not, fall back on simple check against returned values from peregrine
+  return (
+    isRootUrl(project)
+    || Object.values(projectList).includes(project)
+    || programList.includes(project)
```
scintillating7, line 109: I think we need to be careful if the user navigates to just a program name.

mfshao:
deployed in https://mingfei.planx-pla.net/ with arborist UI on. Feel free to test:
https://mingfei.planx-pla.net/_root

src/Submission/ProjectSubmission.jsx
```diff
@@ -9,55 +9,105 @@ import './ProjectSubmission.less';
...
+    programList,
+    userAuthMapping,
+  ) => {
+    if (useArboristUI) {
+      // check against arborist policies
+      return ((isRootUrl(project) && userHasSheepdogProgramAdmin(userAuthMapping))
+        || (isProgramUrl(project)
+        && programList.includes(project)
+        && userHasSheepdogProjectAdmin(userAuthMapping))
+        || Object.values(projectList).includes(project));
```
mpingram, line 40: Making sure I understand, when we get `projectList` from Peregrine, it only includes projects that a user has access to, right? So a user can only view the `commons.url/program-project` page if they have access to the program and project. (Or are a sheepdog project/program admin)

mfshao: Yes! I believe the `projectList` only contains the projects that user has access to. But seems like the `programList` would always contains all the program names in the sheepdog DB. I think this is the current peregrine behavior?

But if we turned the arborist UI on then if a user is not project admin then he/she cannot see any program pages (they are meaningless to them anyways i think)

mpingram: Update, I think I get it now -- I'm a user, I can view the `commons.url/program-project` page as long as program/project exists, and if I'm a sheepdog project admin, I can view `commons.url/program-newproject` even if `program/newproject` doesn't exist yet?

mfshao: actually i should say this:
- If arborist UI is ON, user will NOT be redirect to `/not-found` if 
1. the URL is `_root` and user has sheepdog PROGRAM admin (in this case user can see _root submission page)
2. the URL is `commons.url/program` and the program exists in sheepdog DB and user has sheepdog PROJECT admin (in this case user can see program submission page)
3. the URL is `commons.url/program-project` and the project exists in sheepdog DB and user has access to that project (in this case user can see project submission page)

mfshao: Note the above discussion only applies to whether a user will be redirected to `/not-found`. It does NOT include whether user can see submission UI components (the components inside red box)
Screen Shot 2020-06-15 at 4 26 54 PM

The controlling logic of that UI component is at https://github.com/uc-cdis/data-portal/blob/580922bfed7b84ff867b08836155ba82d3141a74/src/Submission/ProjectSubmission.jsx#L78

mpingram: Gotchu, that all makes sense -- thanks for the explanation!

mfshao:
made more changes: 
1. Checks if file with a given GUID exists before routing to CoreMetadataPage component. If that fails, direct to NotFound.
- `/files/<a_valid_guid>` will be routed to `CoreMetadataPage`
- `/files/<an_invalid_guid>` will be routed to `NotFound`
2. Deal with trailing slash and duplicated slashes in URL when routing
- `/some///path/` --> `/some/path` 
3. Use exact matched for most of router paths
- `/explorer/bla/bla`  will be routed to `NotFound`
- `/explorer`  will be routed to `GuppyDataExplorer`

Therefore, I would like to dismiss stale reviews and can @mpingram or @ZakirG take another look for these newly added logics? Thanks 😄

src/components/NotFound.jsx
```diff
@@ -0,0 +1,20 @@
...
+import NotFoundSVG from '../img/not-found.svg';
+
+class NotFound extends React.Component {
+  render() {
+    return (
+      <div className='error-placeholder__error-msg'>
+        <h1>{'Oops! We can\'t seem to find the page you\'re looking for...'}</h1>
+        <p>
+        Either URL is not valid, or you do not have access.
+        Please contact administrator for more information.
```
ZakirG, line 12: Could this be "Please contact your administrator for more information" or "Please contact the common's administrator for more information" instead?

src/components/NotFound.jsx
```diff
@@ -0,0 +1,20 @@
...
+import './NotFound.less';
+import NotFoundSVG from '../img/not-found.svg';
+
+class NotFound extends React.Component {
+  render() {
+    return (
+      <div className='error-placeholder__error-msg'>
+        <h1>{'Oops! We can\'t seem to find the page you\'re looking for...'}</h1>
+        <p>
+        Either URL is not valid, or you do not have access.
```
ZakirG, line 11: Could you add "this"? So that it says "Either this URL is not valid, or you do not have access."

mfshao:
lgtm analysis failed with no logs... it has succeeded once but didn't updated PR status 🤷‍♂️ 

QA has sign off, will merge after jenkins turns green

PR #697, opened 2020-06-17
Fix/indexing gui timeout

ZakirG:
### Bug Fixes
- Increase indexing GUI timeout


mfshao:
I think I'm late to the party but if this is a parameter that may need to be adjusted from time to time, could we consider to mount it from manifest as an env var instead of hard code it in portal?

We have sth similar for `WORKSPACE_TIMEOUT_IN_MINUTES` i think

PR #699, opened 2020-06-21
Feat/peregrine request chunk

ZakirG:
This PR splits the query for homepage chart nodes into multiple queries when more than 15 nodes are requested. The chunked query results are combined and rendered. You can run my unit tests with 
`npm test src/Index/utils.test.js`

Deployed in https://zakir.planx-pla.net/

Fulfills https://ctds-planx.atlassian.net/browse/PXP-6235

### Improvements
- Homepage chart data request is now chunked to respect Postgres query limits

### Deployment Changes
- New optional config field in the index section of gitops.json called homepageChartNodeChunkSize. Sets the value for paritioning requests to Peregrine for the homepage chart. Defaults to 15 if not provided.

### DIFF
--- a/src/Index/utils.js
+++ b/src/Index/utils.js
@@ -21,21 +21,72 @@ const updateRedux = async projectNodeCounts => getReduxStore().then(
 
 // Chunk into groups of 15
 // Chunking is necessary to avoid hitting Postgres limits
-export const getChunkedPeregrineRequestURLs = (nodesToRequest) => {
-  let requestURLs = [];
+export const getChunkedPeregrineRequestUrls = (nodesToRequest) => {
+  let requestUrls = [];
 
   let k = 0;
   let chunkSize = 15;
   let n = nodesToRequest.length;
+
+  if(n == 0) {
+    return [`${datasetUrl}?nodes=`];
+  }
   
   while(k < n) {
-    let nodeChunk = nodesToRequest.slice(k, k + chunkSize);
-    let chunkRequestURL = `${datasetUrl}?nodes=${nodeChunk.join(',')}`;
-    requestURLs.push(chunkRequestURL);
+    let listRightBound = Math.min(k + chunkSize, n)
+    let nodeChunk = nodesToRequest.slice(k, listRightBound);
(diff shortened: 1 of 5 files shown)


src/Index/utils.js
```diff
@@ -19,6 +19,66 @@ const updateRedux = async projectNodeCounts => getReduxStore().then(
   },
 );
 
+// Chunk into groups of 15
+// Chunking is necessary to avoid hitting Postgres limits
+export const getChunkedPeregrineRequestUrls = (nodesToRequest) => {
+  const requestUrls = [];
+
+  let k = 0;
+  const chunkSize = 15;
```
mfshao, line 28: this `chunkSize` should be configuable but optional from config file, just like `lineLimit`. This 15 can be the default value if the it is omit

If an env have much more projects we may need to decrease this number further

src/Index/utils.js
```diff
@@ -32,40 +92,41 @@ export const loadHomepageChartDataFromDatasets = async (callback) => {
...
-      }
-      break;
-    default:
-      break;
     }
-  })
-    .catch((err) => {
-      console.log(err);
-    });
+    return;
```
mfshao, line 121: also print something for other errors?

src/Index/utils.js
```diff
@@ -19,6 +19,66 @@ const updateRedux = async projectNodeCounts => getReduxStore().then(
...
+const makePeregrineRequestForNode = async url => fetchWithCreds({
+  path: url,
+}).then((res) => {
+  if (res.status === 200) {
+    return [res.data, 200];
+  }
+  return [null, res.status];
+})
+  .catch((err) => {
+    console.log(err);
```
mfshao, line 55:
```suggestion
    console.error(err);
```

src/Index/utils.js
ZakirG: Are you sure that's desirable? The original functionality did not print something for the other errors. If a query comes back with status code 500, I don't think a console print stating "The homepage chart node query came back with status code 500" would be desirable, or consistent with our usual error handling style in this repo.

mfshao: i don't have any strong feeling about it. but i think it would be more helpful to know what went wrong, plus we are already printing some error messages in other functions

PR #700, opened 2020-07-01
Fix issue loading dictionary viewer for null props

paulineribeyre:
fix `Cannot read property 'length' of null` error at line 61 `if (cursor < str.length)` 
example https://github.com/uc-cdis/covid19_datadictionary/commit/afb325a2889d2753fe245fd3edc913f7c3de7864#
`null` should not be a value in the dictionary (we're removing it), but the dd viewer should not crash anyway

### Bug Fixes
- Fix issue loading dictionary viewer for null props


paulineribeyre:
other services should be fine, sheepdog will accept the value "null" but no value will be stored in the DB, that's expected (the mechanism for removing values is to submit "null")

PR #702, opened 2020-07-06
fix: data upload auth check

mfshao:
Fixed a bug cause `submission` page not correctly displaying `Map My Files` component if 1. `useArboristUI` is on AND 2.  service in `file_upload` has been changed from `fence` to `*` (which is intruduced in Indexd 2.12.0 or 2020.07)

Related PR: https://github.com/uc-cdis/indexd/pull/275 (see `Deployment changes`)

### Bug Fixes
- Fixed a bug cause `submission` page not correctly displaying `Map My Files` component

### Deployment changes
- Portal sould be updated to include this fix if `Indexd` is above `2.12.0` or `2020.07` and user want to use Arborist UI

### DIFF
--- a/src/authMappingUtils.js
+++ b/src/authMappingUtils.js
@@ -53,8 +53,8 @@ export const listifyMethodsFromMapping = (actions) => {
 
 
 export const userHasDataUpload = (userAuthMapping = {}) => {
-  // data_upload policy is resource data_file, method file_upload, service fence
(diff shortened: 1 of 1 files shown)


vpsx:
One small pedantic thing but I think the PR description should say service in the `file_upload` permission in the useryaml. (data_file is the resource)

PR #703, opened 2020-07-07
Dictionary categories refactoring

paulineribeyre:
Ticket COV-377

- **Current behavior for unknown categories:**
Screen Shot 2020-07-06 at 3 54 23 PM

For a category such as `genomic_data_file`, we can "guess" match it with `data_file`, which would result in using the same icon and color for `data_file` and `genomic_data_file`:

- **Dictionary viewer table view:**
Screen Shot 2020-07-06 at 3 50 55 PM

- **Dictionary viewer graph view:**
Screen Shot 2020-07-06 at 3 51 09 PM

- **Project submission page:**
Screen Shot 2020-07-06 at 3 51 25 PM

Alternatively (or in addition to this), we could keep the default icon (or display no icon) and dynamically generate (colorblind-friendly) new colors for new categories, instead of hardcoding which categories are expected and assign colors. I think that would be cool for a future improvement. We can even make it configurable

### Improvements
- Handle icons and colors for new node categories in dictionary viewer and in project submission page by matching them with existing category suffixes
- The project summary table now counts all file nodes instead of only the "data_file" category (e.g. "index_file", "metadata_file") (same behavior as Peregrine/Pidgin)
- Refactor the multiple "getCategoryColor" functions into a single one

### DIFF
--- a/src/DataDictionary/NodeCategories/helper.js
+++ b/src/DataDictionary/NodeCategories/helper.js
@@ -1,4 +1,4 @@
-import * as d3 from 'd3-scale';
+import { schemeCategory20 } from 'd3-scale';
 
 import IconAdministrative from './icons/icon_administrative.svg';
 import IconAnalysis from './icons/icon_analysis.svg';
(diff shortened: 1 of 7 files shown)


src/DataDictionary/NodeCategories/helper.js
```diff
@@ -1,3 +1,5 @@
+import * as d3 from 'd3-scale';
```
mfshao, line 1: can we just import `schemeCategory20` rather than the whole pack since we only need that

PR #708, opened 2020-07-20
feat: precommit

mfshao:
Add precommit hooks and sec baseline

### Improvements
- Add precommit hooks and sec baseline


.pre-commit-config.yaml
```diff
@@ -0,0 +1,6 @@
+repos:
+    -   repo: git@github.com:Yelp/detect-secrets
```
paulineribeyre, line 2: to match other repos you can add this https://github.com/uc-cdis/fence/blob/master/.pre-commit-config.yaml#L7-L13
but lgtm

PR #711, opened 2020-07-22
Fixing Safari charts

atharvar28:
Handles fixing the Safari charts

use of Moment Library 
- the source data for `props.label` in tooltip was in format yyyy-mm-dd. But Safari's date class didn't support but chrome's date class had support for that format. Safari displayed the date as ("NaN", "NaN") which was used as the label on the tooltip
- Moment is a date parsing library for JS and it supports both Safari and Chrome browser. It allows us to use any date format which is needed

### DIFF
--- a/src/Covid19Dashboard/PlotChart/index.jsx
+++ b/src/Covid19Dashboard/PlotChart/index.jsx
@@ -11,7 +11,8 @@ import './PlotChart.less';
 class CustomizedAxisTick extends React.Component {
   render() {
     const { x, y, payload } = this.props; // eslint-disable-line react/prop-types
-    const val = payload.value.replace(/-/g, '/'); // eslint-disable-line react/prop-types
+    console.log(x);
+    const val = payload.value; // eslint-disable-line react/prop-types
     const formattedDate = `${moment(val).month() + 1}/${moment(val).date()}`;
     return (
       <g transform={`translate(${x},${y})`}>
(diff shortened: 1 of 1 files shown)


src/Covid19Dashboard/PlotChart/index.jsx
```diff
@@ -23,8 +23,8 @@ class CustomizedAxisTick extends React.Component {
 
 function getDates(startDate, endDate, days) {
   const dates = [];
-  let currentDate = new Date(startDate);
-  const endingDate = new Date(endDate);
+  let currentDate = new Date(startDate).replace(/-/g, '/');
+  const endingDate = new Date(endDate).replace(/-/g, '/');
```
mfshao, line 27: since we are importing `moment`, should we use it throughout this code to parsing dates from strings?

basically we would format our date strings in to ISO 8601 formats and call `moment()` to parse it https://momentjs.com/docs/#/parsing/string/ this way we could avoid calling native `new Date()` and got unexpected results in the future

atharvar28: valid point 👍  @mfshao

src/Covid19Dashboard/PlotChart/index.jsx
```diff
@@ -7,12 +8,11 @@ import {
 import { numberWithCommas } from '../dataUtils.js';
 import './PlotChart.less';
 
-
 class CustomizedAxisTick extends React.Component {
   render() {
     const { x, y, payload } = this.props; // eslint-disable-line react/prop-types
-    const val = payload.value; // eslint-disable-line react/prop-types
-    const formattedDate = `${new Date(val).getUTCMonth() + 1}/${new Date(val).getUTCDate()}`;
+    const val = payload.value.replace(/-/g, '/'); // eslint-disable-line react/prop-types
```
mfshao, line 14: i think if you are using `moment` then you don't need all these `.replace(/-/g, '/')`

src/Covid19Dashboard/PlotChart/index.jsx
```diff
@@ -23,17 +23,17 @@ class CustomizedAxisTick extends React.Component {
 
 function getDates(startDate, endDate, days) {
   const dates = [];
-  let currentDate = new Date(startDate);
-  const endingDate = new Date(endDate);
+  let currentDate = moment(startDate).replace(/-/g, '/');
+  const endingDate = moment(endDate).replace(/-/g, '/');
```
mfshao, line 27: same as above

src/Covid19Dashboard/PlotChart/index.jsx
```diff
@@ -23,17 +23,17 @@ class CustomizedAxisTick extends React.Component {
...
   const addDaysToDate = (date) => {
-    const newDate = new Date(date.valueOf());
-    newDate.setDate(newDate.getUTCDate() + days);
+    const newDate = moment(date).valueOf();
+    newDate.set(newDate.date() + days);
     return newDate;
   };
   while (currentDate <= endingDate) {
```
mfshao, line 33: i suspect if this will work if `currentDate` and `endingDate` are both `moment` objects, if not, this may ends up in an infinite loop

maybe shuold consider https://momentjs.com/docs/#/query/is-same-or-before/

atharvar28:
creating a new PR for the changes

atharvar28:
creating a new PR for the change

PR #714, opened 2020-07-29
PXP-6236 Fix/query redir

mfshao:
This PR fulfills https://ctds-planx.atlassian.net/browse/PXP-6236

Fixed a bug causing login redirect to ditch query params in URL

Verify at: https://mingfei.planx-pla.net/jnkns-jenkins/search?node_type=app_checkup

### Bug Fixes
- Fixed a bug causing login redirect to ditch query params in URL


src/Login/ProtectedContent.jsx
```diff
@@ -279,9 +279,16 @@ class ProtectedContent extends React.Component {
     window.scrollTo(0, 0);
     const pageFullWidthClassModifier = isPageFullScreen(this.props.location.pathname) ? 'protected-content--full-screen' : '';
     if (this.state.redirectTo) {
+      let fromURL = '/';
+      if (this.state.from && this.state.from.pathname) {
+        fromURL = this.state.from.pathname;
+        if (this.state.from.search && this.state.from.search !== '') {
```
ZakirG, line 285: Should we not account for a non-search URL with query parameters like this?
`https://qa-covid19.planx-pla.net/open-JHU/explorer?filter={...}`

mfshao: oh sure! I think this implementation would takes care of that as well. It is called `location.search` but actually it is the the URL query string, not just `?search=`

https://github.com/ReactTraining/history/blob/master/docs/api-reference.md#locationsearch

ZakirG: oh!! haha funny coincidence

PR #722, opened 2020-08-18
PXP-6516 Fix/layout adjust

mfshao:
This PR fulfills https://ctds-planx.atlassian.net/browse/PXP-6516

Adjust layout of the submission form so each field is clearly separated

### Bug Fixes
- Fixed some warning messages about incorrect prop types

### Improvements
- Updatde some UI components from Ant Design
- Adjusted layout of the submission form so each field is clearly separated

### Dependency updates
- Removed `material-ui`
- Added `antd`

### DIFF
--- a/src/Submission/TextInput.less
+++ b/src/Submission/TextInput.less
@@ -6,6 +6,7 @@
 
 .text-input__label {
   margin: 3px;
+  font-weight: bold;
   display: inline-block;
 }
 
@@ -16,7 +17,10 @@
 }
 
 .text-input__input-description {
-  font-size: 1rem;
   display: inline-block;
   width: 750px;
+}
+
+.ant-form-item {
+  margin-bottom: 0;
 }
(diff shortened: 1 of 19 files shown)


src/components/tables/ProjectTable.test.jsx
```diff
@@ -15,11 +14,9 @@ test('Project tables renders', () => {
 
   // Material-UI components require the Mui theme ...
```
mpingram, line 15:
```suggestion
```

src/Submission/TextInput.less
```diff
@@ -16,7 +17,10 @@
...
 
 .text-input__input-description {
-  font-size: 1rem;
   display: inline-block;
   width: 750px;
-}
\ No newline at end of file
+}
+
+.ant-form-item {
```
mpingram, line 24: Possible to put this css on a wrapper div instead of `ant-form-item`? This could cause confusing behavior down the line if an ant form somewhere else in the website looks different because of this css.

mfshao: yeah good call! updated

PR #723, opened 2020-08-26
NIAID POC calls Requestor

paulineribeyre:


src/StudyViewer/StudyDetails.jsx
```diff
@@ -16,11 +18,27 @@ const onDownload = () => {
...
+      resource_path: '/programs/TODO',
+      resource_name: this.props.data.name,
+    };
+    fetchWithCreds({
+      path: `${requestorPath}request`,
+      method: 'POST',
+      body: JSON.stringify(body),
+    }).then(
+      ({ data }) => {
+        if (data && data.redirect_url) {
```
mfshao, line 34: do we need to do anything in case of this fails?

paulineribeyre: probably but i'm counting on you to handle it in the real code 😆

mfshao: yeah that's fine, i mean do they (niaid devs) need that when they are playing with this mock up?

paulineribeyre: i don't think so, i told them it's WIP

PR #732, opened 2020-09-09
Fix/nct

mfshao:
Addressing
https://ctds-planx.atlassian.net/browse/PXP-6730
https://ctds-planx.atlassian.net/browse/PXP-6731
https://ctds-planx.atlassian.net/browse/PXP-6733
https://ctds-planx.atlassian.net/browse/PXP-6734
https://ctds-planx.atlassian.net/browse/PXP-6735
https://ctds-planx.atlassian.net/browse/PXP-6736
https://ctds-planx.atlassian.net/browse/PXP-6737

PR #734, opened 2020-09-14
Fix/508

mfshao:
This is to add some useful changes into master as described in https://github.com/uc-cdis/data-portal/pull/732

Related tickets https://ctds-planx.atlassian.net/browse/PXP-6733 and https://ctds-planx.atlassian.net/browse/PXP-6737

### Improvements
- Fix page hierarchy for homepage UI components
- Add alt text for images in footer and header

### DIFF
--- a/src/components/layout/NavBar.jsx
+++ b/src/components/layout/NavBar.jsx
@@ -109,7 +109,7 @@ class NavBar extends Component {
 
     // added for backward compatibility
     // should always add homepageHref to components in portal config in the future
-    const homepageHref = config.homepageHref || components.homepageHref;
+    const homepageHref = components.homepageHref || config.homepageHref;
 
     return (
       <div className='nav-bar'>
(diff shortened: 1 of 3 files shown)


src/components/layout/NavBar.jsx
```diff
@@ -107,17 +107,21 @@ class NavBar extends Component {
         return this.canUserSeeComponent(item.name) ? navButton : null;
       });
 
+    // added for backward compatibility
+    // should always add homepageHref to components in portal config in the future
+    const homepageHref = config.homepageHref || components.homepageHref;
```
paulineribeyre, line 112: if `config.homepageHref` is deprecated maybe we should use `const homepageHref = components.homepageHref || config.homepageHref;`

paulineribeyre (approved):

src/components/layout/NavBar.jsx
mfshao: ohh good call

PR #735, opened 2020-09-18
fix(workspaceIndex): missing no-access routes

frickjack:
[PXP-6684](https://ctds-planx.atlassian.net/browse/PXP-6684)

* fix missing routes in `workspaceIndex.jsx`

### Improvements

* fix missing routes in `workspaceIndex.jsx`

### DIFF
--- a/src/workspaceIndex.jsx
+++ b/src/workspaceIndex.jsx
@@ -22,7 +22,7 @@ import theme from './theme';
 import getReduxStore from './reduxStore';
 import ReduxLogin, { fetchLogin } from './Login/ReduxLogin';
 import { ReduxNavBar, ReduxTopBar, ReduxFooter } from './Layout/reduxer';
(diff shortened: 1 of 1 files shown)


src/workspaceIndex.jsx
```diff
@@ -123,6 +124,16 @@ async function init() {
...
                     }
                   />
+                  <Route
+                    exact
+                    path='/lw-workspace/'
+                    component={ErrorWorkspacePlaceholder}
+                  />
+                  <Route
+                    exact
+                    path='/no-workspace-access'
```
mfshao, line 134: why we don't import these paths from localconf? we have `workspaceUrl` and `workspaceErrorUrl`

frickjack: ok :+1:

mfshao (approved):
👍

PR #737, opened 2020-09-21
Feat/covid bundle

mfshao:
Separate Covid19 dashboard into another bundle to reduce bundle size.

Supply `GEN3_BUNDLE=covid19` as env var to enable covid bundle

There is another ticket to remove Arranger https://ctds-planx.atlassian.net/browse/PXP-6409

### New Features
- Separate Covid19 dashboard into another bundle to reduce bundle size

### Deployment changes
- Add `GEN3_BUNDLE=covid19` as portal env var to enable covid bundle

### DIFF
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -104,16 +104,23 @@ if (process.env.NODE_ENV !== 'dev' && process.env.NODE_ENV !== 'auto') {
 
 const entry = {
   "bundle": ['babel-polyfill', './src/index.jsx'],
-  "workspaceBundle": ['babel-polyfill', './src/workspaceIndex.jsx']
+  "workspaceBundle": ['babel-polyfill', './src/workspaceIndex.jsx'],
+  "covid19Bundle": ['babel-polyfill', './src/covid19Index.jsx']
 };
 
 if (process.env.GEN3_BUNDLE === 'workspace') {
   // Just build the workspace bundle as bundle.js
   entry.bundle = entry.workspaceBundle;
   delete entry.workspaceBundle;
+  delete entry.covid19Bundle;
+} else if (process.env.GEN3_BUNDLE === 'covid19') {
+  entry.bundle = entry.covid19Bundle;
+  delete entry.workspaceBundle;
+  delete entry.covid19Bundle;
 } else if (process.env.GEN3_BUNDLE === 'commons') {
   // optimize webpack build
   delete entry.workspaceBundle;
+  delete entry.covid19Bundle;
 }
 // else - by default build all entry points
 //    note that runWebpack ensures GEN3_BUNDLE is set,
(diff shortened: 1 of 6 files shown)


frickjack (approved):
Hey Mingfei,

I'm approving this, but it would be cool to try to figure out a way that the index page could take some props or something that was some kind of route override.  So an abstract index could have default route table:
```
{ 
    "path1": { exact: true|false, component: <jsx-something /> },
   "path2": ...,
}
```
, and actual instances of index-page provide overrides somehow:
```
class CovidIndexPage {
    ...
    routeOverrides = { "path3": { exact, component: <jsx-whatever/>}}
}
```
where the base class render does something like:
```
render() {
     const myRoutes = { ...this.defaultRoutes, ...this.overrideRoutes };
     ...
     for (const it of myRoutes.entries()) {
     }
```

I'm not sure what the best way to do that kind of thing in React is ...

webpack.config.js
```diff
@@ -104,16 +104,23 @@ if (process.env.NODE_ENV !== 'dev' && process.env.NODE_ENV !== 'auto') {
 
 const entry = {
   "bundle": ['babel-polyfill', './src/index.jsx'],
-  "workspaceBundle": ['babel-polyfill', './src/workspaceIndex.jsx']
+  "workspaceBundle": ['babel-polyfill', './src/workspaceIndex.jsx'],
+  "covid19Bundle": ['babel-polyfill', './src/covid19Index.jsx']
 };
 
 if (process.env.GEN3_BUNDLE === 'workspace') {
```
paulineribeyre, line 111: how would we enable multiple bundles?

mfshao: if you don't specify `GEN3_BUNDLE` value then it will build all bundles, is that what you mean?

paulineribeyre: i see, what i was thinking is that we could have `GEN3_BUNDLE=covid19,workspace`

but then i'm not sure how this `entry.bundle` stuff would work 😅

mfshao: oh cool. I don't think we have the need to deploy multiple bundles right now so right now if you spin up a portal pod you would have to choose one flavor only

paulineribeyre: that's fair, then follow up question, why don't we do
```
if (process.env.GEN3_BUNDLE === 'workspace') {
  entry.bundle = ['babel-polyfill', './src/workspaceIndex.jsx'];
}
```
and default to only building `./src/index.jsx`, instead of deleting entries which looks a bit messy?

mfshao: I think for `Dockerfile` we don't have `GEN3_BUNDLE` set so it will build all bundles to make sure nothing stops us from building an image. So we can either `delete` from a full `entry` array, or adding into it for that case

PR #745, opened 2020-09-30
Chore/csrf

frickjack:
* do not rely on cookie for csrf token, but keep it as a fallback

### Improvements

* do not rely on cookie for csrf token, but keep it as a fallback


mfshao (approved):
👍

PR #746, opened 2020-09-30
PXP-6919 Fix study viewer redirection

paulineribeyre:
Study viewer: revert "?request_access" logic that is causing the redirection to be blocked as a popup by the browser, and replace "Request Access" button with "Login to Request Access" button when the user is logged out

### Improvements
- Study Viewer: replace "Request Access" button with "Login to Request Access" button when the user is logged out and fix blocked redirections


mfshao (approved):
thanks

src/StudyViewer/StudyDetails.jsx
```diff
@@ -201,9 +182,9 @@ class StudyDetails extends React.Component {
                }}
              />
            </Modal>
-           {(displayRequestAccessButton && !this.props.data.accessRequested) ?
+           {(displayRequestAccessButton && !userHasLoggedIn && !this.props.data.accessRequested) ?
```
mfshao, line 185: probably can remove `displayRequestAccessButton` here?

PR #749, opened 2020-10-01
PXP-6920 Study viewer improvements

paulineribeyre:
Jira Ticket: [PXP-6920](https://ctds-planx.atlassian.net/browse/PXP-6920)

### Improvements
- Study viewer improvements: configurable buttons in study viewer configuration; buttons can be hidden on the list view or on the single item view; request access redirection modal improvements

### DIFF
--- a/src/StudyViewer/StudyDetails.jsx
+++ b/src/StudyViewer/StudyDetails.jsx
@@ -56,7 +56,7 @@ class StudyDetails extends React.Component {
       // it means we haven't finished check yet
       // next is to check if user has access to the resource
       if (!this.isDataAccessible(this.props.data.accessibleValidationValue)) {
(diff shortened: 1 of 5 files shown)


src/StudyViewer/StudyDetails.jsx
```diff
@@ -239,7 +273,7 @@ class StudyDetails extends React.Component {
                }}
              />
            </Modal>
-           {(!userHasLoggedIn && !this.props.data.accessRequested) ?
+           {this.requestAccessButtonVisible ?
```
mfshao, line 276: i think only check this one is not sufficient, the alert will show after the user has logged in

it should be `(this.requestAccessButtonVisible && !userHasLoggedIn && !this.props.data.accessRequested)`

mfshao (approved):
👍

PR #757, opened 2020-11-05
(PXP-7133) Fix/login filter

mpingram:
Jira Ticket: [PXP-7133](https://ctds-planx.atlassian.net/browse/PXP-7133)

### Bug Fixes
- Fixes issue where login IDP option selector is extremely slow for large numbers of options (e.g. pandemic response commons)

### DIFF
--- a/src/Login/Login.jsx
+++ b/src/Login/Login.jsx
@@ -132,13 +132,15 @@ class Login extends React.Component {
               {p.desc}
               <div className='login-page__entry-login'>
                 {
-                  // If there are multiple URLs, display a dropdown next
-                  // to the login button
-                  // We use createFilter here with `ignoreAccents: false` to increase performance when dealing
-                  // with large numbers of IDPs (Incommon logins can have 3k+ options!). 
-                  // The `stringify` option to createFilter here ensures that react-select only searches over 
-                  // the login options' names (e.g. "The University of Chicago") and not the actual option values, 
-                  // which are URLs.
+                  // If there are multiple URLs, display a dropdown next to
+                  // the login button We use createFilter here with
+                  // `ignoreAccents: false` to increase performance when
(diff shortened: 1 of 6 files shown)


src/Login/Login.jsx
```diff
@@ -139,7 +139,7 @@ class Login extends React.Component {
                       isClearable
                       isSearchable
                       options={loginOptions[i]}
-                      filterOption={createFilter({ ignoreAccents: true, ignoreCase: true })}
+                      filterOption={createFilter({ ignoreAccents: false, matchFrom: 'any', stringify: option => `${option.label}` })}
```
paulineribeyre, line 142: you said `ignoreAccents: true` was slowing down the search, i assume `ignoreCase: true` was too?

what do `matchFrom: 'any'` and `stringify` do? i can't find docs :(

mpingram: I can't find docs on `stringify` either 😢 but I found it on a gh issue for searching over just labels and not values: https://github.com/JedWatson/react-select/issues/3403

I think this is the closest to docs on the createFilter function: https://react-select.com/advanced#custom-filter-logic

ignoreCase: true is actually the default to my understanding, I think it's also slowing down search but I found it more confusing to do case-sensitive search 🤷 If perf is an issue we can start there and then stick react-windowed back in if we need to.

paulineribeyre: i realize i'm being annoying but we might want to add a comment in the code to explain what these parameters do, since it's not documented (🤦‍♀️) and i agree the search should not be case sensitive

mpingram: Not at all, that's a great point! Lemme stick a comment in there

paulineribeyre (approved):
thanks for fixing!!

PR #758, opened 2020-11-06
PXP-7063 Configurable study viewer items order

paulineribeyre:
Jira Ticket: [PXP-7063](https://ctds-planx.atlassian.net/browse/PXP-7063)

### New Features
- Configurable study viewer items order

### DIFF
--- a/docs/study_viewer.md
+++ b/docs/study_viewer.md
@@ -53,6 +53,7 @@ Example configuration:
         "docDataType": "openAccessFile", // ES index of the open access documents, optional
         "openMode": "open-first", // optional, configure how the study viewer list do with each collapsible card on initial loading, see details in notes
         "openFirstRowAccessor": "", // optional, only works if `openMode` is `open-first`
+        "defaultOrderBy": [field name (eg "title", or from listItemConfig.blockFields/tableFields)>, <"asc" (default) or "desc">] // optional, overrides `openFirstRowAccessor`
         "buttons": [
             {
                 // configuration common to all buttons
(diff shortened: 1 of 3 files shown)


docs/study_viewer.md
```diff
@@ -53,6 +53,7 @@ Example configuration:
         "docDataType": "openAccessFile", // ES index of the open access documents, optional
         "openMode": "open-first", // optional, configure how the study viewer list do with each collapsible card on initial loading, see details in notes
         "openFirstRowAccessor": "", // optional, only works if `openMode` is `open-first`
+        "defaultOrderBy": [field name (eg "title", or from listItemConfig.blockFields/tableFields)>, <"asc" (default) or "desc">] // optional, overrides `openFirstRowAccessor`
```
mfshao, line 56:
```suggestion
        "defaultOrderBy": [<field name (eg "title", or from listItemConfig.blockFields/tableFields)>, <"asc" (default) or "desc">] // optional, overrides `openFirstRowAccessor`
```

also, why is this config an array of strings rather than `{ field: <field_name>, order: <"asc" (default) or "desc">  }` ?

paulineribeyre: to stop people from trying to configure more than one :) i guess this format could create confusion too, but i thought it wasn't too bad

src/StudyViewer/StudyViewer.jsx
```diff
@@ -79,6 +84,28 @@ class StudyViewer extends React.Component {
       });
     }
 
+    // sort items - order = [<field name>, <"asc" (default) or "desc">]
+    const field = order[0];
```
mfshao, line 88: i think you will need to check if `order` is specified here? since `studyViewerConfig.defaultOrderBy` can be `undefined`?

mfshao (approved):
👍

PR #761, opened 2020-11-16
Configurable workspace description

paulineribeyre:
Jira Ticket: COV-146

### Improvements
- Update workspace text from "Launching workspace..." to "Launching Workspace, this process may take several minutes"
- Add optional "workspacePageDescription" configuration

### DIFF
--- a/src/Workspace/index.jsx
+++ b/src/Workspace/index.jsx
@@ -1,5 +1,7 @@
 import React from 'react';
+import parse from 'html-react-parser';
 import Button from '@gen3/ui-component/dist/components/Button';
+
 import {
   workspaceUrl,
   wtsPath,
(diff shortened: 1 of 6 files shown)


src/Workspace/index.jsx
```diff
@@ -266,6 +269,11 @@ class Workspace extends React.Component {
             this.state.notebookStatus !== 'Running' &&
             this.state.notebookStatus !== 'Stopped' ?
               <div>
+                {workspacePageDescription ?
+                  <p className='workspace__description'>
+                    {parse(workspacePageDescription)}
```
mfshao, line 274: i'm in favor of give this thing a default description

mfshao (approved):

src/Workspace/index.jsx
paulineribeyre: why not but that's not my call to make 🤔

PR #762, opened 2020-11-17
fix: check user access base on both method and service

mfshao:
Portal's Arborist UI feature should check for both `service` and `method`
User should be able to submit data if has `{ service: *, method: *}` for a project
Also wrapped buttons in project submission table with `Link` so user can use right click to open it in a new tab

### Bug Fixes
- Portal's Arborist UI feature should check for both `service` and `method`
- User should be able to submit data if has `{ service: *, method: *}` for a project

### Improvements
- Make buttons in project submission table be able to get right click

### DIFF
--- a/src/Submission/ProjectSubmission.jsx
+++ b/src/Submission/ProjectSubmission.jsx
@@ -81,9 +81,8 @@ class ProjectSubmission extends React.Component {
         !useArboristUI
         || (isRootUrl(project) && userHasSheepdogProgramAdmin(userAuthMapping))
         || (isProgramUrl(project) && userHasSheepdogProjectAdmin(userAuthMapping))
-        || userHasMethodForServiceOnProject('create', '*', project, userAuthMapping)
-        || userHasMethodForServiceOnProject('update', '*', project, userAuthMapping)
-        || userHasMethodForServiceOnProject('*', '*', project, userAuthMapping)
+        || userHasMethodForServiceOnProject('create', 'sheepdog', project, userAuthMapping)
+        || userHasMethodForServiceOnProject('update', 'sheepdog', project, userAuthMapping)
       ) {
         return (
           <React.Fragment>
--- a/src/Submission/ProjectSubmission.jsx
+++ b/src/Submission/ProjectSubmission.jsx
@@ -8,7 +8,7 @@ import SubmitForm from './SubmitForm';
 import Spinner from '../components/Spinner';
 import './ProjectSubmission.less';
 import { useArboristUI } from '../configs';
-import { userHasMethodOnProject, isRootUrl, isProgramUrl, userHasSheepdogProgramAdmin, userHasSheepdogProjectAdmin } from '../authMappingUtils';
+import { userHasMethodForServiceOnProject, isRootUrl, isProgramUrl, userHasSheepdogProgramAdmin, userHasSheepdogProjectAdmin } from '../authMappingUtils';
 
 class ProjectSubmission extends React.Component {
   componentDidMount() {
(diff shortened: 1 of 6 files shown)


src/authMappingUtils.js
```diff
@@ -59,15 +59,17 @@ export const userHasDataUpload = (userAuthMapping = {}) => {
   return resource !== undefined && resource.some(actionIsFileUpload);
 };
 
-export const userHasMethodOnResource = (method, resourcePath, userAuthMapping = {}) => {
+export const userHasMethodForServiceOnResource =
+(method, service, resourcePath, userAuthMapping = {}) => {
   const actions = userAuthMapping[resourcePath];
-  return actions !== undefined && actions.some(x => x.method === method);
+  return actions !== undefined && actions.some(x => (x.service === service && x.method === method));
```
paulineribeyre, line 65:
```suggestion
  return actions !== undefined && actions.some(x => ((x.service === service || x.service === '*') && (x.method === method) || x.method ==='*')));
```

src/Submission/ProjectSubmission.jsx
```diff
@@ -81,8 +81,9 @@ class ProjectSubmission extends React.Component {
         !useArboristUI
         || (isRootUrl(project) && userHasSheepdogProgramAdmin(userAuthMapping))
         || (isProgramUrl(project) && userHasSheepdogProjectAdmin(userAuthMapping))
-        || userHasMethodOnProject('create', project, userAuthMapping)
-        || userHasMethodOnProject('update', project, userAuthMapping)
+        || userHasMethodForServiceOnProject('create', '*', project, userAuthMapping)
+        || userHasMethodForServiceOnProject('update', '*', project, userAuthMapping)
+        || userHasMethodForServiceOnProject('*', '*', project, userAuthMapping)
```
paulineribeyre, line 86: you shouldn't need this line if `userHasMethodForServiceOnResource` is fixed to handle `*` properly

paulineribeyre (changes requested):

src/Submission/ProjectSubmission.jsx
mfshao: Oh this is actually a question i wanted to ask, I'm thinking do we need to check for `{service: sheepdog, method: *}`?

Right now Portal checks for `/services/sheepdog/submission/project` and `/services/sheepdog/submission/program` to see if user has sheepdog admin, could we use `{service: sheepdog, method: *}` instead?

src/Submission/ProjectSubmission.jsx
```diff
@@ -81,8 +81,9 @@ class ProjectSubmission extends React.Component {
         !useArboristUI
         || (isRootUrl(project) && userHasSheepdogProgramAdmin(userAuthMapping))
         || (isProgramUrl(project) && userHasSheepdogProjectAdmin(userAuthMapping))
-        || userHasMethodOnProject('create', project, userAuthMapping)
-        || userHasMethodOnProject('update', project, userAuthMapping)
+        || userHasMethodForServiceOnProject('create', '*', project, userAuthMapping)
+        || userHasMethodForServiceOnProject('update', '*', project, userAuthMapping)
```
vpsx, line 85: I think this will now require the user to have literal `service: '*'` in the relevant policy role... I'm not sure this is what is intended?

src/Submission/ProjectSubmission.jsx
paulineribeyre: > do we need to check for `{service: sheepdog, method: *}`?

ugh, yeah maybe 🤔 and for the download button it should be `fence`, etc

> could we use `{service: sheepdog, method: *}` instead?

if by "sheepdog admin" you mean program/project creation, then `/services/sheepdog/submission/project` and `/services/sheepdog/submission/program` are right. But if you mean "access to all the data", then `{service: sheepdog, method: *}`

src/Submission/ProjectSubmission.jsx
mfshao: right, so my understanding is if a user is able to submit to sheepdog then he/she must have either of the following for a project:
1. `{service: *, method: create/update}`
2. `{service: *, method: *}`
3. `{service: sheepdog, method: *}` (?)
4. `{service: sheepdog, method: create/update}` (??? - do we even have this)

Does this hold? 🤔

vpsx: Yup yup, * is supposed to match anything. So I think if you add the `|| x.service === '*'` in userHasMethodForServiceOnResource (and same for method), then in these checks you can just have 
```
        || userHasMethodForServiceOnProject('create', 'sheepdog', project, userAuthMapping)
        || userHasMethodForServiceOnProject('update', 'sheepdog', project, userAuthMapping)
```

(Not sure what you mean by do we even have this--as in, in the useryamls?)

mfshao: Yeah I know we can have {service: sheepdog, method: create/update}, just not sure if we have this in any user.yaml now

vpsx (approved):
Thank you!!

paulineribeyre (approved):

PR #766, opened 2020-11-24
chore: bump guppy and gen3-ui-component version

ZakirG:


frickjack (approved):
lgtm - but please bump up to nodejs 14:
hey Zakir - would you go ahead and bump the Dockerfile up to nodejs 14?
   https://github.com/uc-cdis/data-portal/blob/master/Dockerfile#L18
cc @Mingfei

PR #767, opened 2020-11-24
PXP-7062 Fix/dap tag

mfshao:
Jira Ticket: [PXP-7062](https://ctds-planx.atlassian.net/browse/PXP-7062)

Implement DAP virtual pageview for NCT as described in [this official doc](https://github.com/digital-analytics-program/gov-wide-code/blob/master/documentation/GSA%20DAP%204.1%20-%20DAP%20Code%20Capabilities%20Summary%20and%20Reference.pdf)

### Bug Fixes
- Fixed a bug causing a blank chart element being added into chart carousel even though there is no chart configured

### Improvements
- Implement DAP virtual pageview for NCT


webpack.config.js
```diff
@@ -84,9 +91,29 @@ const plugins = [
...
+      const rv = {};
+      if (scriptSrcURLs.length > 0) {
+        scriptSrcURLs.forEach((url) => {
+          rv[(new URL(url)).origin] = true;
+        });
+      }
+      if (Object.keys(rv).length > 0) {
+        return Object.keys(rv).join(' ');
+      }
+      return '';
```
paulineribeyre, line 115: `return Object.keys(rv).join(' ');` would work fine without the `if` and `return '';` statements

(same comment for line 99)

src/nctIndex.jsx
```diff
@@ -94,6 +95,7 @@ async function init() {
           <BrowserRouter basename={basename}>
             <div className='main-page'>
               {GA.init(gaTracking, dev, gaDebug) && <RouteTracker />}
+              {injectDAPTag && <DAPRouteTracker />}
```
paulineribeyre, line 98: as mentioned on slack, i would make the URLs configurable and add the `DAPRouteTracker` in all index pages instead of only the NCT one so that it works for all Commons

src/components/DAPAnalytics.jsx
```diff
@@ -0,0 +1,54 @@
...
+
+const scriptExists = url => document.querySelectorAll(`script[src="${url}"]`).length > 0;
+
+const logPageChange = (pathname, search = '') => {
+  const page = pathname + search;
+  const { location } = window;
+  // check if DAP script has been loaded
+  if (scriptExists('https://dap.digitalgov.gov/Universal-Federated-Analytics-Min.js?agency=NIH&subagency=NIAID')) {
+    // apply virtual pageview using DAP function
+    window.gas('send', 'pageview', `${location.origin}${page}`);
```
paulineribeyre, line 15: in the documentation it doesn't look like they are adding the origin?

Screen Shot 2020-12-01 at 2 05 44 PM

src/index.ejs
```diff
@@ -8,7 +8,7 @@
...
+    <meta http-equiv="Content-Security-Policy" content="default-src 'self' https://login.bionimbus.org https://wayf.incommonfederation.org; child-src blob:; img-src 'self' https://opendata.datacommons.io https://static.planx-pla.net data: https://*.s3.amazonaws.com https://www.google-analytics.com; script-src 'self' 'unsafe-eval' https://www.google-analytics.com <%= htmlWebpackPlugin.options.script_src %>; worker-src 'self' blob:; style-src 'self' 'unsafe-inline' https://fonts.googleapis.com; object-src 'none'; font-src 'self' data: https://fonts.googleapis.com https://fonts.gstatic.com; connect-src 'self' https://login.bionimbus.org https://*.s3.amazonaws.com https://wayf.incommonfederation.org https://*.mapbox.com https://opendata.datacommons.io https://static.planx-pla.net <%= htmlWebpackPlugin.options.connect_src %>; frame-src <%= htmlWebpackPlugin.options.connect_src %> 'self' https://auspice.planx-pla.net https://auspice.pandemicresponsecommons.org;;">
```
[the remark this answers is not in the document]
mfshao: i have doubts with it 😕 since `htmlWebpackPlugin` won't do anthing for `/dev.html`, so you have to end up with hard coding it in there. And I don't think that is what we want

webpack.config.js
mfshao: oops yeah

src/components/DAPAnalytics.jsx
mfshao: cool, their manual didn't say it has to be without the origin but we can stick to their instructions

src/index.ejs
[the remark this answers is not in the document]
mfshao: oh so you want add `htmlWebpackPlugin` support into `dev.html` as well?

src/components/DAPAnalytics.jsx
paulineribeyre: i really don't know which one's right, just bringing it to your attention 😄

src/index.ejs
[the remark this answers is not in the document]
paulineribeyre: i guess we don't need to fire pagechange events from local dev
i was also just asking you here, i don't know the answer 😆

src/components/DAPAnalytics.jsx
mfshao: yeah you are right actually, at least for GA the path should not have origin, so I think that also holds for DAP

src/index.ejs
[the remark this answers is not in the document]
mfshao: it does make the life of debugging this thing easier, but that is all i think 🤷‍♂️

webpack.config.js
```diff
@@ -84,9 +91,29 @@ const plugins = [
           rv[(new URL(url)).origin] = true;
         });
       }
-      return Object.keys(rv).join(' ');
+      if (connectSrcURLs.length > 0) {
+        connectSrcURLs.forEach((url) => {
```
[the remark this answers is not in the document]
mfshao: i will make the DAP url configurable but I also think it makes sense to keep these `connectSrcURLs` and `scriptSrcURLs` for future

src/components/DAPAnalytics.jsx
```diff
@@ -0,0 +1,53 @@
...
+import { Route } from 'react-router-dom';
+
+const scriptExists = url => document.querySelectorAll(`script[src="${url}"]`).length > 0;
+
+const logPageChange = (pathname, search = '') => {
+  const page = pathname + search;
+  // check if DAP script has been loaded
+  if (scriptExists('https://dap.digitalgov.gov/Universal-Federated-Analytics-Min.js?agency=NIH&subagency=NIAID')) {
+    // apply virtual pageview using DAP function
+    window.gas('send', 'pageview', `${page}`);
```
paulineribeyre, line 14:
```suggestion
    window.gas('send', 'pageview', page);
```

paulineribeyre (approved):

PR #779, opened 2021-01-05
(PXP-6032):fix/table-column-sorting: keep track if sorting is done and sort after field mapping if it has not already been done

ocshawn:
Jira Ticket: [PXP-6032](https://ctds-planx.atlassian.net/browse/PXP-6032)

### Bug Fixes
keep track if sorting is done using the field attribute and sort after field mapping if it has not already been done

### DIFF
--- a/src/GuppyDataExplorer/ExplorerVisualization/index.jsx
+++ b/src/GuppyDataExplorer/ExplorerVisualization/index.jsx
@@ -97,8 +97,9 @@ class ExplorerVisualization extends React.Component {
 
   render() {
     const chartData = this.getData(this.props.aggsData, this.props.chartConfig, this.props.filter);
-    const tableColumns = (this.props.tableConfig.fields && this.props.tableConfig.fields.length > 0)
-      ? this.props.tableConfig.fields : this.props.allFields;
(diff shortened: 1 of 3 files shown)


paulineribeyre (commented):
lgtm, defer to @mfshao

src/GuppyDataExplorer/ExplorerVisualization/index.jsx
```diff
@@ -229,6 +230,7 @@ class ExplorerVisualization extends React.Component {
               className='guppy-explorer-visualization__table'
               tableConfig={{
                 fields: tableColumns,
+                ordered: tableColumnsOrdered,
```
mfshao, line 233: should update `TableConfigType` about this under https://github.com/uc-cdis/data-portal/blob/08f94f1fd08bfab0dd99055def4689eb7a6c943a/src/GuppyDataExplorer/configTypeDef.js

ocshawn: i have added thanks for the heads up

mfshao (approved):

PR #782, opened 2021-01-08
Fix PRC charts displaying undefined

paulineribeyre:
Jira Ticket: COV-595

bug introduced in https://github.com/uc-cdis/data-portal/pull/773

### Bug Fixes
- Covid-19 dashboard: fix US charts not loading because of an undefined variable

### DIFF
--- a/src/Covid19Dashboard/WorldMapChart/index.jsx
+++ b/src/Covid19Dashboard/WorldMapChart/index.jsx
@@ -165,7 +165,7 @@ class WorldMapChart extends React.Component {
         FIPS: `${feature.properties.FIPS || feature.properties.fips}`,
         values: {
           'confirmed cases': confirmed,
-          deaths: deaths,
+          deaths,
         },
       };
       if (recovered) {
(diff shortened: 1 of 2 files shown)


src/Covid19Dashboard/WorldMapChart/index.jsx
```diff
@@ -192,7 +192,8 @@ class WorldMapChart extends React.Component {
 
       // density map data contains fips and iso3;
       // choropleth map data contains FIPS and iso_a3.
-      const fips = `${feature.properties.FIPS || feature.properties.fips}`;
+      let fips = feature.properties.FIPS || feature.properties.fips;
+      fips = fips ? `${fips}` : fips; // convert to string if not undefined
```
ocshawn, line 196: this will evaluate to false if vale of fips is 0, is this acceptable?

ocshawn: can use 
`typeof fips !== 'undefined'`

paulineribeyre: fips 0 doesn't exist afaik but i'll update it to be safe

ocshawn (approved):
looks good

PR #790, opened 2021-01-21
fix: optional format arg

mfshao:
Make `format` to be optional in query strings issued by Guppy for backward compatibility, so newer Portal can work with old Guppy if don't need the export to CSV/TSV feature deployed

Companion PR: https://github.com/uc-cdis/guppy/pull/108

Deployed at https://mingfei.planx-pla.net/explorer (Guppy server there is version `2020.12` - old)
Also deployed at https://qa-midrc.planx-pla.net/explorer (Guppy server there is version `0.10.0` - new)

### Bug Fixes
- Make `format` to be optional in query strings issued by Guppy for backward compatibility 

### Dependency updates
- Guppy to 0.10.1

### DIFF
--- a/src/GuppyDataExplorer/ExplorerButtonGroup/index.jsx
+++ b/src/GuppyDataExplorer/ExplorerButtonGroup/index.jsx
@@ -352,7 +352,7 @@ class ExplorerButtonGroup extends React.Component {
     }
     this.props.downloadRawData(queryArgObj).then((res) => {
       if (res) {
-        const blob = new Blob([isJsonFormat ? JSON.stringify(res, null, 2) : res], { type: `text/${isJsonFormat ? 'json' : fileFormat.toLowerCase()}` });
+        const blob = new Blob([isJsonFormat ? JSON.stringify(res, null, 2) : res], { type: `text/${isJsonFormat ? 'json' : fileTypeKey}` });
         FileSaver.saveAs(blob, filename);
       } else {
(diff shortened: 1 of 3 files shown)


src/GuppyDataExplorer/ExplorerButtonGroup/index.jsx
```diff
@@ -344,9 +344,13 @@ class ExplorerButtonGroup extends React.Component {
...
-    this.props.downloadRawData({ format: fileTypeKey }).then((res) => {
+    const fileTypeKey = fileFormat.toLowerCase();
+    const isJsonFormat = fileTypeKey === 'json' || fileTypeKey === 'data';
+    const queryArgObj = {};
+    if (fileTypeKey !== 'data') {
+      queryArgObj.format = fileTypeKey;
+    }
+    this.props.downloadRawData(queryArgObj).then((res) => {
       if (res) {
         const blob = new Blob([isJsonFormat ? JSON.stringify(res, null, 2) : res], { type: `text/${isJsonFormat ? 'json' : fileFormat.toLowerCase()}` });
```
cterrazas2, line 355: Since you are using the `toLowerCase()` for the `fileTypeKey` you could just use it for the blob too instead of calling twice.

mfshao: what do you mean by "calling twice"?

cterrazas2 (approved):

PR #793, opened 2021-01-28
(PXP-7433): feat/explorationTableFormatting: added toggle to show hide empty columns in tables

ocshawn:
Jira Ticket: [PXP-7433](https://ctds-planx.atlassian.net/browse/PXP-7433)

### New Features
added toggle to show hide empty columns in tables


src/GuppyDataExplorer/ExplorerTable/index.jsx
```diff
@@ -224,6 +226,45 @@ class ExplorerTable extends React.Component {
       });
     });
   };
+  /**
+   * Sets empty coumns visability to state showEmptyColumns
+   * @param colList: column configs for root table
+   * @returns column configs for root table with
+   * empty coumns visability set to state showEmptyColumns
+   */
+  markEmptyColums = (colList) => {
```
ocshawn, line 235: should this just be inline rather then its own separate function? would like to know other peoples thoughts

src/GuppyDataExplorer/ExplorerTable/index.jsx
```diff
@@ -224,6 +226,45 @@ class ExplorerTable extends React.Component {
       });
     });
   };
+  /**
+   * Sets empty coumns visability to state showEmptyColumns
+   * @param colList: column configs for root table
+   * @returns column configs for root table with
+   * empty coumns visability set to state showEmptyColumns
```
ocshawn, line 233: spelling to fix: coumns

mfshao (changes requested):
There are some typos, but my main concern is about the logic of the `markEmptyColums` function. Please see my comments below

And it would be good if you can deploy this to somewhere and mention that in PR comment so others can test

src/GuppyDataExplorer/ExplorerTable/index.jsx
```diff
@@ -224,6 +226,45 @@ class ExplorerTable extends React.Component {
...
+  markEmptyColums = (colList) => {
+    if (!this.props.rawData || this.props.rawData.length === 0) {
+      return colList;
+    }
+    // Go through each column and see if it has any current data
+    return colList.map((obj) => {
+      const tempObj = obj;
+      let columnIsEmpty = true;
+      // see if any item has data in current column
+      for (let i = 0; i < this.props.rawData.length; i += 1) {
```
mfshao, line 244: minor point, you might be able to simplify this loop code with using some helper functions from `lodash` like `findIndex` https://lodash.com/docs/4.17.15#findIndex

ocshawn: simplified using every

ocshawn:
Changes are visible in QA https://qa-covid19.planx-pla.net/explorer

src/GuppyDataExplorer/ExplorerTable/index.jsx
```diff
@@ -298,7 +339,19 @@ class ExplorerTable extends React.Component {
...
-          : <p className='explorer-table__description'>{explorerTableCaption}</p> }
+          : <div className='explorer-table__description'>
+            <span>
+              {explorerTableCaption}
+            </span>
+            <label className={`explorer-table__hide_empty_column_toggle ${this.props.rawData ? '' : 'ant-switch-disabled'}`}>
+                Show Empty Columns
+              <Switch
+                onChange={this.hideEmptyColumnToggle}
+                disabled={!this.props.rawData}
```
mfshao, line 350: should this get disabled when the table gets locked as well?

mfshao (approved):
generally lgtm, with a small comment

src/GuppyDataExplorer/ExplorerTable/index.jsx
ocshawn: will there be anything in this.props.rawData when the table is locked? i can look into

mfshao: there is a props for that, see https://github.com/uc-cdis/data-portal/blob/2924128c9d9c145fda07e0eb20c02011366d64b5/src/GuppyDataExplorer/ExplorerTable/index.jsx#L358

the table gets locked if the tiered access level is `regular` and current user doesn't have access to all data

ocshawn: if you look further up in the code line 341 this code is not shown if its locked
`{(this.props.isLocked) ? <React.Fragment />
          : this code`

mfshao: oops good point!

PR #803, opened 2021-02-10
(PXP-7402) Discovery page

mpingram:
Jira Ticket: [PXP-7402](https://ctds-planx.atlassian.net/browse/PXP-7402)

Adds a beta version of the Discovery page, which shows a list of the studies available on the Gen3 commons. Also adds support for Typescript (compiling w/ babel, tests, and linting). This PR looks a lot bigger than it really is due to test data + Typescript setup -- the files to start with are:
  - `src/Discovery/index.tsx` <- this connects the Discovery page to the metadata-service backend and handles access checking
  - `src/Discovery/Discovery.tsx` <- this is the entire Discovery page -- I kept it in one file on purpose, as the Discovery page is still a beta feature and we're going to heavily modify its appearance in the future (adding a grid view, etc).
  - `src/Discovery/Discovery.test.tsx`

~~Running at https://mpingram.planx-pla.net/discovery.~~ Working on deploying to my dev env but running into problems with submitting data to MDS. For the moment, please test with these instructions:
1. Check out this branch of `cdis-manifest`: https://github.com/uc-cdis/cdis-manifest/tree/feat/bdc-staging-discovery-page.
2. Run portal locally against https://staging.gen3.biodatacatalyst.nhlbi.nih.gov/:
`HOSTNAME=staging.gen3.biodatacatalyst.nhlbi.nih.gov NODE_ENV=auto bash ./runWebpack.sh`

#### Configuration ####
- Enabled by `"featureFlags": { "discovery": true }` in the portal config.
- Requires `"useArboristUI": true` in the portal config.
- Configured by `"discoveryConfig": <DiscoveryConfig>` in the portal config. DiscoveryConfig schema can be found in src/Discovery/DiscoveryConfig.d.ts.

### New Features
- Adds a beta version of the Discovery page, which shows a list of the studies available on the Gen3 commons.

### Improvements
- Adds support for Typescript to the portal codebase.


.eslintrc.js
```diff
@@ -1,137 +1,172 @@
 module.exports = {
-  //"extends": "eslint:recommended",
-  "extends": "airbnb",
-  "root":true,
-  "env": {
-      "browser": true,
-      "es6": true,
-      "jest": true
+  // "extends": "eslint:recommended",
```
mpingram, line 2: Sorry about this 🤦  My editor autoformatted this file while I was editing it.

package.json
```diff
@@ -29,6 +29,8 @@
     "d3-zoom": "^1.7.3",
     "echarts": "^4.2.1",
     "echarts-for-react": "^2.0.14",
+    "enzyme": "^3.2.0",
```
ocshawn, line 32: Is enzyme needed in production?

mpingram: Great q, my linter was complaining at me that enzyme was in `devDepdendencies` and not dependencies so I moved it here to keep it happy. I was actually surprised to learn this but it turns out that when we're using webpack, it doesn't matter if a dep is in `devDependencies` or `dependencies`, it only gets included in production if a file imports it. So I believe this has no impact on the production bundle as long as no non-test files import enzyme.

src/Discovery/Discovery.tsx
```diff
@@ -0,0 +1,552 @@
...
+            })}
+          </div>
+        ))}
+      </Space>
+    </Modal>
+  </div>);
+};
+
+Discovery.defaultProps = {
+  params: { studyUID: undefined },
```
ZakirG, line 549: I'm not 100% sure, but I think that `null` would be a better initializer here. Null is used to define something programmatically empty. Undefined is meant to say that the reference is not existing. If I would make a property intentionally empty but still present, I would use null rather than undefined. Paraphrased from https://stackoverflow.com/questions/6604749/what-reason-is-there-to-use-null-instead-of-undefined-in-javascript?rq=1

src/index.jsx
```diff
@@ -45,6 +45,7 @@ import isEnabled from './helpers/featureFlags';
 import sessionMonitor from './SessionMonitor';
 import Workspace from './Workspace';
 import ResourceBrowser from './ResourceBrowser';
+import DiscoveryBeta from './Discovery';
```
ZakirG, line 48: Is there a strategic value to including "Beta" in the name of this component?

ZakirG:
It looks like this PR adds enhancements to the portal config like the `discoveryConfig` block. Can you update `docs/portal_config.md` accordingly?

src/index.jsx
mpingram: Woops, no I forgot to update that.

src/Discovery/DiscoveryConfig.d.ts
```diff
@@ -0,0 +1,92 @@
+/* eslint-disable camelcase */
```
mfshao, line 1: is there a reason to not use camelcase for these configuration fields?

src/Discovery/index.tsx
```diff
@@ -0,0 +1,75 @@
...
+import { DiscoveryConfig } from './DiscoveryConfig';
+import { userHasMethodForServiceOnResource } from '../authMappingUtils';
+import { hostname, discoveryConfig, useArboristUI } from '../localconf';
+
+const loadStudiesFromMDS = async (): Promise<any[]> => {
+  // Why `_guid_type='discovery_metadata'? We need to distinguish the discovery page studies in MDS
+  // from other data in MDS. So all MDS records with `_guid_type='discovery_metadata'` should be
+  // the full list of studies for this commons.
+  const GUID_TYPE = 'discovery_metadata';
+  const LIMIT = 10000; // required or else mds defaults to returning 10 records
```
mfshao, line 14: what if we have more than 10000 records in MDS? Does MDS support pagination queries?

src/Discovery/Discovery.tsx
```diff
@@ -0,0 +1,552 @@
...
+
+  return (<div className='discovery-container'>
+    { (config.features.page_title && config.features.page_title.enabled) &&
+      <h1 className='discovery-page-title'>{config.features.page_title.text || 'Discovery'}</h1>
+    }
+    <div className='discovery-header'>
+      <div className='discovery-header__stats-container'>
+        {
+          config.aggregations.map(aggregation => (
+            <div className='discovery-header__stat' key={aggregation.name}>
```
ZakirG, line 331: Optional suggestion: I think these stat boxes would benefit from vertical line separators on their right side, following the convention of the Data Explorer's stat counts.

src/Discovery/Discovery.tsx
```diff
@@ -0,0 +1,552 @@
...
+            { fieldGroup.fields.map((field) => {
+              // display nothing if selected study doesn't have this field
+              // and this field isn't configured to show a default value
+              if (!modalData[field.field] && !field.include_if_not_available) {
+                return null;
+              }
+              return (
+                <div key={field.name} className='discovery-modal__attribute'>
+                  { field.include_name !== false &&
+                        <span className='discovery-modal__attribute-name'>{field.name}</span>
```
mfshao, line 530: minor: personally i'd include `:` after `field.name` so iot will show as `{field.name}: {value}`

src/Discovery/Discovery.tsx
```diff
@@ -0,0 +1,552 @@
...
+    </div>
+    <Modal
+      className='discovery-modal'
+      visible={modalVisible}
+      onOk={() => setModalVisible(false)}
+      onCancel={() => setModalVisible(false)}
+      width='80vw'
+      footer={false}
+    >
+      <Space direction='vertical' size='large'>
```
mfshao, line 491: suggest to add `width: 100%` to style this vertical space, otherwise it won't fill up the modal container if the data is short

src/Discovery/Discovery.tsx
```diff
@@ -0,0 +1,552 @@
...
+        { config.study_page_fields.header &&
+          <Space align='baseline'>
+            <h3 className='discovery-modal__header-text'>{modalData[config.study_page_fields.header.field]}</h3>
+            <a href={`/discovery/${modalData[config.minimal_field_mapping.uid]}/`}><LinkOutlined /> Permalink</a>
```
mfshao, line 495: Suggestion: wrap this link with `copy` so clicking on it will copy the premalink to clipboard might be a better user experience

Portal is already importing `clipboard-plus` for some other pages, might be able to use that

src/Discovery/DiscoveryConfig.d.ts
mpingram: Great point, no there isn't, plus it's different from the portal config. I'll update the config to use camelcase!

src/Discovery/Discovery.tsx
mpingram: That's a really good idea!! If it's ok with you I'm going to add this to a list of improvements for me to make in the future -- specifically changing this from a link to a button that copies the link.

src/Discovery/Discovery.tsx
mpingram: This is a good idea for readability, I'm adding it.

src/Discovery/Discovery.tsx
mpingram: I think you're right! I'm adding it.
Screen Shot 2021-02-12 at 9 13 00 AM

src/Discovery/index.tsx
mpingram: Thank you for calling this out -- MDS does support pagination, AND the max limit is actually 2000 records. I'll change the logic to this:
1. Get LIMIT=1000 records from mds. If request fails throw an error and show an empty page until user refreshes (no retry logic for right now)
2. While the number of records returned is 1000, query again with offset=1000 LIMIT=1000. If request fails throw an error and show an empty page until user refreshes (again no retry logic)

src/Discovery/Discovery.tsx
```diff
@@ -327,15 +327,18 @@ const Discovery: React.FunctionComponent<DiscoveryBetaProps> = (props: Discovery
...
         {
-          config.aggregations.map(aggregation => (
-            <div className='discovery-header__stat' key={aggregation.name}>
-              <div className='discovery-header__stat-number'>
-                {renderAggregation(aggregation, visibleResources)}
-              </div>
-              <div className='discovery-header__stat-label'>
-                {aggregation.name}
+          config.aggregations.map((aggregation, i) => (
+            <>
```
ZakirG: oo fancy

ZakirG (approved):
Works really well. Good job !!!!!!!!

src/Discovery/index.tsx
```diff
@@ -0,0 +1,94 @@
...
+      // eslint-disable-next-line no-await-in-loop
+      const res = await fetch(url);
+      if (res.status !== 200) {
+        throw new Error(`Request for study data at ${url} failed. Response: ${JSON.stringify(res, null, 2)}`);
+      }
+      // eslint-disable-next-line no-await-in-loop
+      const jsonResponse = await res.json();
+      const studies = Object.values(jsonResponse).map(entry => entry[STUDY_DATA_FIELD]);
+      allStudies = allStudies.concat(studies);
+      const noMoreStudiesToLoad = studies.length < LIMIT;
```
mfshao, line 37: Just want to make sure a potential edge case is covered in here: what if the MDS has exactly 1000 records? Would the 2nd request error out or finishes with 200 but with no data being returned?

mpingram: Great q, let me test that!

mpingram: I just tested this by setting LIMIT=137 which results in 2 queries, one with LIMIT=137 OFFSET=0 which returns all 137 studies, and then a second query which returns no studies (`{}`). The empty object doesn't cause any problems, it results in a empty list of studies which gets concat'ed with `allStudies` so the page works like normal. Thanks for catching this edge case!

mfshao: yay awesome! thanks for checking

mfshao (approved):
lgtm

PR #804, opened 2021-02-12
(PXP-7429): fix/workspaceLogout: added backup if api call fails to refresh user, …

ocshawn:
Jira Ticket: [PXP-7429](https://ctds-planx.atlassian.net/browse/PXP-7429)

### Bug Fixes
- added backup if api call fails to refresh user
- added listeners to iframe when on workspace page


src/SessionMonitor/index.js
```diff
@@ -80,9 +103,19 @@ export class SessionMonitor {
...
       store.dispatch(fetchUser).then((response) => {
+        this.fetchUserAttemptCounter = 0;
         if (response.type === 'UPDATE_POPUP') {
           this.popupShown = true;
         }
+      }).catch(() => {
+        this.fetchUserAttemptCounter += 1;
+        // try again after 5 seconds only retry 9 times
+        // stop after 9 times as this will be triggered again after updateSessionTime
+        if (this.fetchUserAttemptCounter < 10) {
```
mfshao, line 114: what are these 9 retries for?

src/SessionMonitor/index.js
```diff
@@ -70,6 +88,11 @@ export class SessionMonitor {
       return Promise.resolve(0);
     }
 
+    // If on workspace start workspace specific event listeners
+    if (this.isUserOnPage('workspace')) {
+      this.workspaceStart();
```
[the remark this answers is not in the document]
ocshawn: sorry as soon as i finished making this pull request i thought of moving it to the component, please see latest commit

src/SessionMonitor/index.js
ocshawn: good point, don't need the retries as the code automatically retries every few minutes

src/SessionMonitor/index.js
```diff
@@ -83,6 +83,9 @@ export class SessionMonitor {
         if (response.type === 'UPDATE_POPUP') {
           this.popupShown = true;
         }
+      }).catch(() => {
+        // if API failed check if user is still logged in
+        this.notifyUserIfTheyAreNotLoggedIn();
```
mfshao, line 88: why logging out user at here? the logic was if it can reach `refreshSession()` then it means the user is still logged in and actively using the portal

mfshao (commented):
please address the travis issue before this PR can get merged

src/SessionMonitor/index.js
ocshawn: the bug had to do with the user being logged out, but not alerted that they had been logged out, leading to other errors. this code checks if they are still logged in and if not alerts them, it does not log them out. I am still not 100% on how all of this code works if you do not think this check is needed i can remove

mfshao: I think the bug was actually buried deeper in the logic flow: the `notifyUserIfTheyAreNotLoggedIn` function will make a call to the submission api and when it returns a 401 the portal assumes the user is logged out. But this qill result in as the current behavior that for some times the popup showed up early while in other occassions some are shown up later.

This is mainly due to the fact that the access token in cookie is not updated until the pervious one has expired. After this fix https://github.com/uc-cdis/fence/pull/874 is merged, both session token and access token will be updated.

For that case, I probably will test and change this logic flow back to let portal actively logout users. So this piece of code will probably be removed in my work. I'm okay will merging as is because I think it won't break things, just a FYI

The changes to the workspace code LGTM

mfshao (approved):

src/SessionMonitor/index.js
ocshawn: sounds good, thanks for looking over, im glad someone else was also looking into this

src/Workspace/index.jsx
```diff
@@ -61,22 +62,25 @@ class Workspace extends React.Component {
...
+    e.target.focus();
+
+    // add event listeners for sessionMonitor timeout
+    const iframeContent = e.target.contentDocument;
+    if (iframeContent) {
+      iframeContent.addEventListener('mousedown', () => sessionMonitor.updateUserActivity(), false);
+      iframeContent.addEventListener('keypress', () => sessionMonitor.updateUserActivity(), false);
+    }
+  }
+
```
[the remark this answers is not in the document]
pieterlukasse: See also this for what I think should have been used: https://developer.mozilla.org/en-US/docs/Web/API/Window/postMessage . And this PR which proposes a solution using "message" events: https://github.com/uc-cdis/data-portal/pull/1002

PR #806, opened 2021-02-18
PXP-4219 Fix/session

mfshao:
Jira Ticket: [PXP-4219](https://ctds-planx.atlassian.net/browse/PXP-4219)

Improved token refresh and session monitor logic, see comments in `SessionMonitor/index.js`

TL;DR: Portal will now tracking user's activity interval and actively logging out inactive users, rather than passively rely on waiting for Fence's token expiration.

To support this, Fence must be able to update access token in cookie before the current one expires as well, see https://github.com/uc-cdis/fence/pull/874

Also removed a good chunk of old codes in portal (mostly relevent to the old `submissionApiOauth`, which is before Fence)

Deployed in: https://qa-mickey.planx-pla.net/

### Improvements
- Updated session monitor and logout inactive user logic, fixing various bugs (401 requests in middle of session, wrongly displayed logout popups, etc...)

### Dependency updates
- Should be used with Fence [4.26.1](https://github.com/uc-cdis/fence/releases/tag/4.26.1)


.eslintrc.js
```diff
@@ -168,5 +168,4 @@ module.exports = {
       typescript: {}, // this loads <rootdir>/tsconfig.json to eslint
     },
   },
-  ignorePatterns: ['src/params.js'],
```
mfshao, line 171: have to put this into `.eslintignore` since my vscode eslint doesn't like this somehow and it breaks my local eslint 🤷‍♂️

src/Login/ProtectedContent.jsx
```diff
@@ -196,50 +177,7 @@ class ProtectedContent extends React.Component {
           // The oauth dance below is only relevant for legacy commons - pre jwt
```
ZakirG, line 177: I think this comment may need to be removed now

src/Login/ProtectedContent.jsx
```diff
@@ -168,15 +153,11 @@ class ProtectedContent extends React.Component {
...
-    const nowMs = Date.now();
     const newState = Object.assign({}, initialState);
 
     if (!newState.authenticated) {
       return Promise.resolve(newState);
     }
-    if (nowMs - lastTokenRefreshMs < 41000) {
-      return Promise.resolve(newState);
-    }
     return store.dispatch(fetchProjects())
```
ZakirG, line 161: The if-blocks in this area appear to be unnecessary -- the conditions all result in `return Promise.resolve(newState);`

ZakirG: Specifically this snippet:
```
 const projects = store.getState().submission.projects;
        if (projects) {
          // user already has a valid token
          return Promise.resolve(newState);
        } else if (info.status !== 403 || info.status !== 401) {
          // do not authenticate unless we have a 403 or 401
          // should only check 401 after we fix fence to return correct
          // error code for all cases
          // there may be no tables at startup time,
          // or some other weirdness ...
          // The oauth dance below is only relevant for legacy commons - pre jwt
          return Promise.resolve(newState);
        }
        return Promise.resolve(newState);
```

ZakirG (approved):

src/Login/ProtectedContent.jsx
mfshao: i think the original logic of marking user as not authed and redirecting to `/login` is still valid, so I updated with that

PR #807, opened 2021-02-18
feat: data explorer query state in url

ZakirG:
Jira Ticket: https://ctds-planx.atlassian.net/browse/PXP-7517

This PR supports loading and unloading Data Explorer state to and from a URL. The feature currently supports only single-select and range filters. This feature is disabled by default; it can be enabled with the `storeExplorerStateInURL` featureFlag in the portal config.

Companion PRs:
https://github.com/uc-cdis/guppy/pull/112
https://github.com/uc-cdis/gen3-ui-component/pull/98

### New Features
- Data Explorer state can now be loaded and unloaded to and from the page URL. This feature is disabled by default; it can be enabled with the `storeExplorerStateInURL` featureFlag in the portal config.


src/GuppyDataExplorer/GuppyDataExplorer.jsx
```diff
@@ -13,29 +13,98 @@ import {
...
+import isEnabled from '../helpers/featureFlags';
 
 class GuppyDataExplorer extends React.Component {
   constructor(props) {
     super(props);
+    let initialState = {};
+    let initialFilter = {};
+
+    // aborting URL decode
+    if (isEnabled('storeExplorerStateInURL')) {
```
mfshao, line 25: let's call this feature flag `explorerStoreFilterInURL` to indicate this is a data explorer feature (since it starts with `expolrer...`)

src/GuppyDataExplorer/GuppyDataExplorer.jsx
```diff
@@ -13,29 +13,98 @@ import {
...
+      const stateFromURL = getQueryParameter('filters');
+      if (stateFromURL) {
+        const decodedFilter = atob(stateFromURL);
```
mfshao, line 28: suggest to also surround this with `try/catch` block

for example, if you delete a few chars from the base64 encoded filter string, paste it into address bar and enter, i think the page will crash because of this `atob()` will throw errors

src/GuppyDataExplorer/GuppyDataExplorer.jsx
```diff
@@ -13,29 +13,98 @@ import {
...
+      encodableExplorerStateForURL: { filter: initialFilter },
     };
   }
 
+  getEncodedQueryStateFromExplorer = () => {
+    const stateObj = this.state.encodableExplorerStateForURL;
+    if (this.isExplorerStatePristine(stateObj)) {
+      return '';
+    }
+    const encoded = encodeURIComponent(btoa(JSON.stringify(stateObj)));
```
mfshao, line 50: It would be nice if you could put some comments into the changes in this code, like briefly explain why are we using `btoa()` and `atob()` in here

We both know we are doing base64 encoding/decoding ere, but just to share the knowledges

src/GuppyDataExplorer/GuppyDataExplorer.jsx
```diff
@@ -13,29 +13,98 @@ import {
...
+    }
+    window.history.pushState('', '', `?filters=${encodedState}`);
+  }
+
   handleReceiveNewAggsData = (newAggsData) => {
     this.setState({ aggsData: newAggsData });
   };
 
+  handleFilterChangeForQueryStateUrl = (eventIn) => {
+    let event = Object.assign({}, eventIn);
```
mfshao, line 68: minor: can also consider using the spread operator...

src/GuppyDataExplorer/GuppyDataExplorer.jsx
```diff
@@ -13,29 +13,98 @@ import {
...
+        delete event[field];
+      }
+    }
+    newExplorerState.filter = event;
+    this.setState({ encodableExplorerStateForURL: newExplorerState });
+  }
+
+  isExplorerStatePristine = (explorerState) => {
+    const values = Object.values(explorerState);
+    for (let i = 0; i < values.length; i += 1) {
```
mfshao, line 84: minor: might use a `for...of` loop

src/GuppyDataExplorer/utils.js
```diff
@@ -73,3 +73,26 @@ export const labelToPlural = (label, titleCase = false) => {
...
+export const getQueryParameter = (queryKey) => {
+  // Stable method to get query parameter for given key
+  const decoded = decodeURIComponent(window.location.search)
+    .replace('?', '')
+    .split('&')
+    .map(param => param.split('='))
+    .reduce((values, [key, value]) => {
+      values[key] = value; // eslint-disable-line no-param-reassign
+      return values;
+    }, {});
```
mfshao, line 86: probably can simplify this logic with the ES6 `URLSearchParams` API https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams

Dockerfile
```diff
@@ -21,7 +21,8 @@ RUN apt-get update && apt-get install -y --no-install-recommends \
     && rm -rf /var/lib/apt/lists/* \
     && ln -sf /dev/stdout /var/log/nginx/access.log \
     && ln -sf /dev/stderr /var/log/nginx/error.log \
-    && npm install -g npm@7.6.3
+    && npm install -g npm@7.8.0 \
+    && npm config set maxsockets 5
```
mfshao, line 25: can you also add this into Travis script? just to keep consistancy and prevent any more drama from happening in the future

src/GuppyDataExplorer/GuppyDataExplorer.jsx
ZakirG: When I use a for...of loop like `for (const value of values) {`, i get this eslint error:

```
/Users/zakirgowani/Desktop/Projects/data-portal/src/GuppyDataExplorer/GuppyDataExplorer.jsx
  91:5  error  iterators/generators require regenerator-runtime, which is too heavyweight for this guide to allow them. Separately, loops should be avoided in favor of array iterations  no-restricted-syntax
```

so i'm gonna leave it as is ~

src/GuppyDataExplorer/GuppyDataExplorer.jsx
```diff
@@ -13,29 +13,107 @@ import {
...
+    let initialFilter = {};
+
+    if (isEnabled('explorerStoreFilterInURL')) {
+      const stateFromURL = getQueryParameter('filters');
+      if (stateFromURL) {
+        try {
+          const decodedFilter = atob(stateFromURL);
+          const isValidJSON = IsValidJSONString(decodedFilter);
+          if (isValidJSON) {
+            initialState = JSON.parse(decodedFilter);
```
mfshao, line 31: is this needed? can you do `JSON.parse(decodedFilter).filter` instead since you already has validated the JSON?

mfshao (approved):
very cool

mfshao:
BTW I think the export to workspace is not working in your dev env but that is probably not related to this PR

package.json
```diff
@@ -33,7 +33,7 @@
     "enzyme-adapter-react-16": "^1.6.0",
     "file-saver": "^1.3.3",
     "fuse.js": "^3.3.0",
-    "graphiql": "^0.13.0",
+    "graphiql": "~0.13.0",
```
mfshao, line 36: latest `0.17.0` of this guy seems breaking webpack and babel 🤦 , pinning it to `0.13.x` for now

PR #814, opened 2021-03-01
(COV-553): feat/updateMapRanges: update map ranges

ocshawn:
Jira Ticket: [COV-553](https://occ-data.atlassian.net/browse/COV-553)

### New Features
- update map ranges
- change component logic so it does not run unnecessarily

### DIFF
--- a/src/Covid19Dashboard/IllinoisMapChart/index.jsx
+++ b/src/Covid19Dashboard/IllinoisMapChart/index.jsx
@@ -67,6 +67,52 @@ class IllinoisMapChart extends React.Component {
       },
       hoverInfo: null,
     };
+    this.mapData = {
+      modeledCountyGeoJson: null,
+      colors: {},
+      colorsAsList: null,
+    };
+  }
+  componentDidUpdate() {
+    if (!(this.mapData.colorsAsList === null
+          && Object.keys(this.props.jsonByLevel.county).length > 0)) {
+      return;
+    }
+    if (Object.keys(this.props.jsonByLevel.country).length && !this.choroCountyGeoJson) {
+      this.choroCountyGeoJson = addDataToGeoJsonBase(
+        this.props.jsonByLevel.county,
+      );
+    }
+    this.mapData.modeledCountyGeoJson = filterCountyGeoJson(this.props.modeledFipsList);
+
+    // Finds max value in data set then finds its base 10 exponent
+    const maxValExponent = Math.log10(this.mapData.modeledCountyGeoJson.features
+      .map(obj => obj.properties.confirmed)
+      .sort((a, b) => b - a)[0]);
+
+    // Math to calculate Index range for map
+    const colorRangeMath = (base) => {
+      // applies maxValExponent to base then rounds down to greatest place
+      const tempNum = Math.ceil(base ** maxValExponent);
+      const roundingDigits = 10 ** (tempNum.toString().length - 1);
+
+      return Math.floor(tempNum / roundingDigits) * roundingDigits;
+    };
+
+    this.mapData.colors = {
+      0: '#FFF',
+      [colorRangeMath(2)]: '#F7F787',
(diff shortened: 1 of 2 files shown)


ocshawn:
changes are in QA and can be tested at https://qa-covid19.planx-pla.net/

src/Covid19Dashboard/IllinoisMapChart/index.jsx
```diff
@@ -67,6 +67,52 @@ class IllinoisMapChart extends React.Component {
...
+
+      return Math.floor(tempNum / roundingDigits) * roundingDigits;
+    };
+
+    this.mapData.colors = {
+      0: '#FFF',
+      [colorRangeMath(2)]: '#F7F787',
+      [colorRangeMath(3)]: '#EED322',
+      [colorRangeMath(4)]: '#E6B71E',
+      [colorRangeMath(5)]: '#DA9C20',
```
paulineribeyre, line 107: why do we use arrays here?

src/Covid19Dashboard/IllinoisMapChart/index.jsx
```diff
@@ -67,6 +67,52 @@ class IllinoisMapChart extends React.Component {
...
+      this.choroCountyGeoJson = addDataToGeoJsonBase(
+        this.props.jsonByLevel.county,
+      );
+    }
+    this.mapData.modeledCountyGeoJson = filterCountyGeoJson(this.props.modeledFipsList);
+
+    // Finds max value in data set then finds its base 10 exponent
+    const maxValExponent = Math.log10(this.mapData.modeledCountyGeoJson.features
+      .map(obj => obj.properties.confirmed)
+      .sort((a, b) => b - a)[0]);
```
paulineribeyre, line 91: can we use the `max` function instead?

src/Covid19Dashboard/WorldMapChart/index.jsx
```diff
@@ -125,6 +125,90 @@ class WorldMapChart extends React.Component {
...
+
+    // Finds max value in data set
+    const maxCountryVal = this.props.geoJson.features
+      .map(obj => obj.properties.confirmed)
+      .sort((a, b) => {
+        if (typeof a === 'string') {
+          return 1;
+        }
+        if (typeof b === 'string') {
+          return -1;
```
paulineribeyre, line 166: is this because of "<5" strings? if so, please add a comment to explain what you're doing. And the IL map code should be able to handle this too

src/Covid19Dashboard/IllinoisMapChart/index.jsx
ocshawn: it was an artifact of a previous version of the code, ill remove them

ocshawn: the array is used to evaluate the function so you can use it as the ID, is there a cleaner way to do this?

paulineribeyre: oh does it not work without the array? 🤯

ocshawn: yep does not work without the brackets, uses the brackets as a resolver in this case, kind of odd, otherwise you get typeOf errors

src/Covid19Dashboard/IllinoisMapChart/index.jsx
ocshawn: sort vs max speed test https://jsben.ch/hoZGp looks to depend on the browser, chrome has optimized max

src/Covid19Dashboard/IllinoisMapChart/index.jsx
paulineribeyre: interesting

paulineribeyre (approved):

PR #823, opened 2021-03-16
(PXP-7399): fix/highlightActiveNavItem: change the way activeTab is set

ocshawn:
Jira Ticket: [PXP-7399](https://ctds-planx.atlassian.net/browse/PXP-7399)

### Bug Fixes
change the way activeTab is set so that it always highlights which page the user is on

### DIFF
--- a/src/Layout/reducers.js
+++ b/src/Layout/reducers.js
@@ -1,13 +1,10 @@
 const bar = (state = {}, action) => {
   switch (action.type) {
-  case 'ACTIVE_INIT': {
-    return { ...state, active: window.location.pathname };
-  }
   case 'ACTIVE_CHANGED': {
     return { ...state, active: action.data };
   }
   default:
-    return state;
+    return {...state, active: window.location.pathname};
   }
 };
 
--- a/src/Layout/reducers.js
+++ b/src/Layout/reducers.js
@@ -4,7 +4,7 @@ const bar = (state = {}, action) => {
     return { ...state, active: action.data };
   }
   default:
-    return {...state, active: window.location.pathname};
+    return { ...state, active: window.location.pathname };
   }
 };
 
--- a/src/Layout/reducers.js
+++ b/src/Layout/reducers.js
@@ -1,14 +1,13 @@
 const bar = (state = {}, action) => {
   switch (action.type) {
+  case 'ACTIVE_INIT': {
+    return { ...state, active: window.location.pathname };
+  }
   case 'ACTIVE_CHANGED': {
     return { ...state, active: action.data };
   }
   default:
-    // Keep active in sync with URL for navigation updates outside of NavBar
-    if (state.active === window.location.pathname) {
-      return state;
-    }
-    return { ...state, active: window.location.pathname };
+    return state;
   }
 };
 
--- a/src/Layout/reducers.js
+++ b/src/Layout/reducers.js
@@ -4,6 +4,10 @@ const bar = (state = {}, action) => {
     return { ...state, active: action.data };
   }
   default:
+    // Keep active in sync with URL for navigation updates outside of NavBar
+    if (state.active === window.location.pathname) {
+      return state;
+    }
     return { ...state, active: window.location.pathname };
   }
 };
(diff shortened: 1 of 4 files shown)


ocshawn:
changes are in QA https://qa-covid19.planx-pla.net/

src/Layout/reducers.js
```diff
@@ -1,13 +1,14 @@
...
-  case 'ACTIVE_INIT': {
-    return { ...state, active: window.location.pathname };
-  }
   case 'ACTIVE_CHANGED': {
     return { ...state, active: action.data };
   }
   default:
-    return state;
+    // Keep active in sync with URL for navigation updates outside of NavBar
+    if (state.active === window.location.pathname) {
```
williamhaley, line 8: I think this violates one of the three principles of redux; that a reducer be a pure function (looking at the diff it seems like we were already doing that, but maybe now's a good time to re-think it)

https://redux.js.org/understanding/thinking-in-redux/three-principles

Ideally, a reducer run multiple times with the same parameters yields the same results. I think the action(s) that trigger navigation changes or the components that render changes related to the current path should probably be where we compare the current path.

Is the problem that we have some code where we're navigating without using react-router? And we can't reliably determine what the current nav item is?

ocshawn: the problem is that the NavBar does not update when changes are made outside of it. 
there are pages that use `this.props.history.push('/workspace');` that NavBar then does not update from 

i am new to react and react router so any suggestions on how to properly solve this i am open to

mfshao: I didn't test but maybe we can add a `ComponentDidUpdate` call inside `NavBar`, check pathname in there and dispatch action if needed. 🤔

mfshao: I didn't test but maybe we can add a `ComponentDidUpdate` call inside `NavBar`, check pathname in there and dispatch action if needed. 🤔 

But I would hightly suggest to investigate the possibility for a more elegant fix. The current logic flow of `NavBar` which contains a lost of customized `props`, `states` and `actions` is really concerning for me. To my knowledge this kind of work should be able to handeled by `react-router` completely. But this `NavBar` code is 3 years old, whoe knows what happened back then 🤷‍♂️ 

Here is a quick demo I put together using `NavLink` https://codesandbox.io/s/react-router-navlink-meeg2 , it can hightlight the active UI component no matter if a link is clicked, a button is clicked, or directly update URL in address bar. If we can refactor our `NavBar` component using something similar to this, I think it will be a much better solution.

williamhaley: 💯 ^

ocshawn: will redo this all, thanks for the input, have wasted too much time trying to "fix" it

src/components/layout/NavButton.jsx
```diff
@@ -1,35 +1,28 @@
...
-  isActive,
-  tabIndex,
 }) => (
-  <div
-    role='button'
-    tabIndex={tabIndex}
-    className={isActive ? 'body-typo button-active nav-button' : 'body-typo nav-button'}
-    onClick={onActiveTab}
-    onKeyPress={onActiveTab}
+  <NavLink
```
williamhaley, line 11: ✨ 🚀 Excellent 👍

williamhaley (approved):
This looks great! Nice to see some legacy code get an update to a more maintainable solution.

mfshao:
Did you test if it works in local dev mode (running portal on webpack dev server and access using `/dev.html`)?

ocshawn:
yes i tested it locally, and it is also in qa and can be tested there

src/components/layout/NavBar.jsx
```diff
@@ -74,83 +65,51 @@ class NavBar extends Component {
...
-                >
-                  <img
-                    className='nav-bar__logo-img'
-                    src='/src/img/logo.png'
-                    alt={components.homepageAltText || 'Gen3 portal logo'}
-                  />
-                </Link>
-              )
-              }
+              <NavLink exact to={homepageHref}>
```
mfshao, line 93: would this work if the `homepageHref` is an external URL? I remember the `NavLink` can only route you on client side 🤔

mfshao (commented):
This is really awesome 👍 ! Like the way we address this issue and get rid of some tech debts in there. Just a small comment

src/components/layout/NavBar.jsx
ocshawn: will look into what happens, good catch

mfshao (approved):
pretty cool 👍 thank you

PR #824, opened 2021-03-19
PXP-7594 PXP-7598 Fix/pentests

mfshao:
Jira Ticket: [PXP-7594](https://ctds-planx.atlassian.net/browse/PXP-7594) and [PXP-7598](https://ctds-planx.atlassian.net/browse/PXP-7598)

Deployed and tested in: https://mingfei.planx-pla.net/

Try hit: https://mingfei.planx-pla.net/package.json

### Improvements
- Fixed Pentest Info Disclosure: Support Packages & Config Files vulnerability
- Fixed Pentest Outdated Nginx Version vulnerability
- Use NPM 7 in dockerfile
- Use Ubuntu 20.04 as base image in dockerfile
- Use Nginx 1.18 in dockerfile


Dockerfile
```diff
@@ -1,7 +1,7 @@
 # To run: docker run -d --name=dataportal -p 80:80 quay.io/cdis/data-portal
 # To check running container: docker exec -it dataportal /bin/bash
 
-FROM quay.io/cdis/ubuntu:18.04
+FROM quay.io/cdis/ubuntu:20.04
```
williamhaley, line 4: Just curious, do we have a cdis `node` docker image?

nginx.conf
```diff
@@ -50,6 +51,30 @@ server {
       deny all;
     }
 
+    # Block all access to package and config files
+    # Note if WAF is deployed this should already be handled by WAF
+    location ^~ /package.json {
```
williamhaley, line 56: Do we have apps that may be exposing `node_modules` or these sorts of files, or is this here to mimic the WAF and be extra cautious?

williamhaley (approved):
lgtm! Just a couple questions

Dockerfile
mfshao: I think we only have one `12.10.0-alpine` node image

nginx.conf
mfshao: I don't think these are exposed on purpose

And for WAF, I think by default our WAF is configure for only audit and log violation rather than blocking them https://github.com/uc-cdis/cloud-automation/blob/master/gen3/lib/manifestDefaults/modsec/crs-setup.conf#L96-L97 🤔 , so we better just block these from within Portal so the vulnerabilities are fixed whether WAF exists or not

mfshao: and I don't know why WAF is put in that mode by default, need to ask more ppl

mfshao:
will merge after https://github.com/uc-cdis/data-portal/pull/825, since that one is more urgent

PR #825, opened 2021-03-22
fix: single level router

mfshao:
React router `Switch` component can only work with 1st level components only https://stackoverflow.com/q/63073469

If put something nested using `<React.Fragment>` under it then it may fail on resolving routes

This shows up in some envs (e.g: bloodpac) as if you navigate to a project page, you only get a header and footer 🤦  

Verify this bug at: https://data.bloodpac.org/Exhale-Fluxion (notice there is no `main-content`)

Deployed and tested: https://qa-bloodpac.planx-pla.net/DEV-test

This fix should be picked into `2021.03` and `2021.04` release 

### Bug Fixes
- Fixed a bug causing page routes cannot be correctly resolved by react router

### DIFF
--- a/src/index.jsx
+++ b/src/index.jsx
@@ -384,7 +384,7 @@ async function init() {
                       />)
                     }
                   />
-                  {isEnabled('discovery') && <React.Fragment>
+                  {isEnabled('discovery') &&
                     <Route
                       exact
                       path='/discovery'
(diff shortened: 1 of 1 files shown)


mfshao:
This bug is only visible if an env is running Portal `2.45.0+` or `2021.03+` AND has Discovery page enabled

mpingram (approved):
Woah 👀 👀 thank you for catching this Mingfei, this is a huge bug 🤦

PR #830, opened 2021-03-28
Fix/npm7 build

mfshao:
Update NPM config to address intermittent quay.io build failures.

It appears NPM 7 would have open more concurrent file descriptors and/or sockets during `npm i`. Quay.io builds after updating to NPM 7 frequently fails on `EMFILE` error. This behavior is not observed if downgrade to NPM 6, and I also cannot reproduce this issue locally with `docker build`. It appears that this issue only occurs frequently when building images on Quay.io.

~There is no option for NPM to config how many files it can open concurrently, but they have a config to control the max number of socket it can use. In NPM 7, the default value of this `maxsockets` field has changed from `50` to `infinity`, but changing it back to `50` didn't solve the build failure issue. I have to further decrease it to `5` to stabilize the Quay building process. 🤦~

Update: it looks like NPM has introduced a regression in version `7.7.0`, and they are working on it. But until that fix has been released and tested by us, let's juts pin NPM to `7.6.3` for now. 

Also updated NodeJS to 14 in Travis to match our local dev env and dockerfile

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,6 +14,9 @@ cache:
 services:
     - docker
 
+before_install:
+    - npm install -g npm@7.6.3
+
 install:
     - npm install --force
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,11 +1,11 @@
 env:
   - REACT_APP_DISABLE_SOCKET=true
-  
+
 language: node_js
 git:
   depth: 1
 node_js:
-  - "12"
+  - "14"
 
 cache:
   directories:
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -1,11 +1,11 @@
 env:
   - REACT_APP_DISABLE_SOCKET=true
-  
+
 language: node_js
 git:
   depth: 1
 node_js:
-  - "12"
+  - "14"
```
williamhaley, line 8: I'm used to installing `node` via `nvm` or `nodenv`. If we install `npm@7.6.3` in our Docker image does that also mean the system node is set to `14`?

mfshao: For our docker file we pull down the latest stable 14.x nodejs from NodeSource, it would come with npm 6 by default (npm 7 ships with nodejs 15).

For Travis we are still using npm 6, but I can also update it to use npm 7

williamhaley (approved):
:shipit: lgtm

PR #833, opened 2021-04-01
(PXP-7760): Feat/remove empty field lists

ocshawn:
Jira Ticket: [PXP-7760](https://ctds-planx.atlassian.net/browse/PXP-7760)

### New Features
- add hideEmptyFilterSection flag and functionality to hide empty filter sections 
- add handling of edge case where tab will have no visible filters

### Dependency updates
- Update @gen3/ui-component dependency to version 0.7.0
- Update @gen3/guppy dependency to version 0.11.3


src/GuppyDataExplorer/ExplorerVisualization/index.jsx
```diff
@@ -19,6 +19,7 @@ import {
 import { checkForAnySelectedUnaccessibleField } from '../GuppyDataExplorerHelper';
 import './ExplorerVisualization.css';
 import { labelToPlural } from '../utils';
+import { explorerHideEmpty } from './../../configs';
```
paulineribeyre, line 22: i think the preference might be to import from `localconf`

docs/portal_config.md
```diff
@@ -160,6 +160,7 @@ Below is an example, with inline comments describing what each JSON block config
     "explorer": true, // required; indicates the flag and whether to hide it or not
     "explorerPublic": true // optional; If set to true, the data explorer page would be treated as a public component and can be accessed without login. Data explorer page would be public accessible if 1. tiered access level is set to libre OR 2. this explorerPublic flag is set to true.
     "discovery": true, // optional; whether to enable the Discovery page. If true, `discoveryConfig` must be present as well.
+    "explorerHideEmpty": false, // optional, when filtering data hide:  FilterSection when they are empty.
```
mfshao, line 163: I would prefer this field be added to `explorerConfig.filters`, so each explorer tabs can control their explorer filers individually

And rename it to `hideEmptyFilterSection` if it got moved

ocshawn: was going to tie this into other features, but probably best to separate it (so i will rename), I don't think it would be good UI to have items hidden in one tab but not another so im going to leave it separate from the individual tabs for now

mfshao: if you want to keep it in here, then I suggest to rename it to `explorerHideEmptyFilterSection` so it is clear that this is a explorer feature flag

paulineribeyre (commented):
lgtm but @mfshao should review the name change

src/GuppyDataExplorer/ExplorerVisualization/index.jsx
```diff
@@ -216,6 +216,7 @@ class ExplorerVisualization extends React.Component {
                   ...heatMapGuppyConfig,
                 }}
                 filterConfig={heatMapFilterConfig}
+                hideEmptyFilterSection={hideEmptyFilterSection}
```
mfshao, line 219: you don't really need this, this filter is for DAT and it is not visible anyways

mfshao (commented):
see my comment in https://github.com/uc-cdis/data-portal/pull/833#discussion_r605853120 for the name

src/GuppyDataExplorer/ExplorerVisualization/index.jsx
ocshawn: thanks ill remove that one

docs/portal_config.md
ocshawn: updated, thanks for the suggestion

mfshao (approved):

PR #836, opened 2021-04-15
feat: encode query state in url

ZakirG:


mfshao:
no longer needed, original changes already merged in https://github.com/uc-cdis/data-portal/pull/807

PR #838, opened 2021-04-21
Feat/ecosystem user interface

ZakirG:
Jira Ticket: [PXP-7955](https://ctds-planx.atlassian.net/browse/PXP-7955)

Deployed here: https://zakir.planx-pla.net/discovery

Note: This branch preserves Craig's original call to individual MDS's in a file called `deprecated.tsx`. That code is what runs in the page; my code stubs do run fine but the call to the aggregate MDS is currently commented out so as to show the larger dataset on the site.

This PR addresses the following subtasks:
-> https://ctds-planx.atlassian.net/browse/PXP-8032
-> https://ctds-planx.atlassian.net/browse/PXP-8024
-> https://ctds-planx.atlassian.net/browse/PXP-8058

And has in-progress code stubs for these subtasks (the actual responses from the aggregate MDS are mocked with test data for now):
-> https://ctds-planx.atlassian.net/browse/PXP-8021
-> https://ctds-planx.atlassian.net/browse/PXP-8020
-> https://ctds-planx.atlassian.net/browse/PXP-8023
-> https://ctds-planx.atlassian.net/browse/PXP-8038

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -17,6 +17,7 @@ module.exports = {
     },
   },
   rules: {
+    'no-underscore-dangle': "off",
     indent: [
       'error',
       2,
--- a/data/config/default.json
+++ b/data/config/default.json
@@ -649,7 +649,8 @@
       },
       "search": {
         "searchBar": {
-          "enabled": true
+          "enabled": true,
+          "placeholder": "Search through studies by attribute..."
         }
       },
       "authorization": {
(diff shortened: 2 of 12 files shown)


mfshao:
I think we should keep the `default.json` unchanged, and add another local json config file if you need

ZakirG:
> I think we should keep the `default.json` unchanged, and add another local json config file if you need

That sounds reasonable, and I can make a note not to merge default.json changes to data-portal master. However, this PR is not to data-portal master, it is to Craig's branch, which makes default.json changes.

craigrbarnes:
I will do the PR now. I will add a task to have Ecosystem Browser support a local config file.

PR #839, opened 2021-04-21
(HP-97) Support Export to Workspace from the Discovery page

mpingram:
Jira Ticket: [HP-97](https://ctds-planx.atlassian.net/browse/HP-97)

Add support for exporting data files from the Discovery page to Workspaces.
In addition, adds two small features to the Discovery page: allows links to be displayed in the table and the modal, and configures whether the Discovery page is behind a login or not.

### New Features
- Add beta support for exporting data files from the Discovery page to Workspaces. This implementation is likely to change in the future, and the configuration may not be supported in future versions of portal.
- Added `requireLogin` option to Discovery page -- configures whether the Discovery page is public or requires user to log in to view.

### Improvements
- Filtering by access level is done in the table to save space.


src/Discovery/Discovery.tsx
```diff
@@ -316,13 +316,51 @@ const Discovery: React.FunctionComponent<DiscoveryBetaProps> = (props: Discovery
...
+    });
+    // download the manifest
+    const MANIFEST_FILENAME = 'manifest.json';
+    const blob = new Blob([JSON.stringify(manifest, null, 2)], { type: 'text/json' });
+    FileSaver.saveAs(blob, MANIFEST_FILENAME);
   };
 
   const visibleResources = filterByTags(
-    filterByAccessLevel(searchFilteredResources, accessLevel, accessibleFieldName),
+    searchFilteredResources,
```
ZakirG, line 363: Why this change?

docs/portal_config.md
```diff
@@ -388,7 +388,17 @@ Below is an example, with inline comments describing what each JSON block config
     "dropdowns": {} // optional; dropdown groupings for buttons
   },
   "discoveryConfig": { // config for Discovery page. Required if 'featureFlags.discovery' is true. See src/Discovery/DiscoveryConfig.d.ts for Typescript schema.
+    "requireLogin": false, // optional, defaults to false. If true, requires user to sign in before seeing the Discovery page
```
mfshao, line 391: the `Resource Browser` component is calling this field `public`, we should do that, we already have `indexPublic`, `explorerPublic` and `resourceBrowserPublic` done in 3 different ways, we should avoid adding 4th way if possible

docs/portal_config.md
```diff
@@ -388,7 +388,17 @@ Below is an example, with inline comments describing what each JSON block config
...
     "features": {
+      "exportToWorkspaceBETA": { // configures the export to workspace feature. If enabled, the Discovery page data must contain a field which is a list of GUIDs for each study. See `manifestFieldName`
```
mfshao, line 393: are we going to call it `BETA`?

docs/portal_config.md
```diff
@@ -388,7 +388,17 @@ Below is an example, with inline comments describing what each JSON block config
...
+          "enable": boolean
+          "enableDownloadManifest": boolean // enables a button which allows user to download a manifest file for gen3 client
+          "documentationLinks": {
+              "gen3Client": string // link to documentation about the gen3 client. Used for button tooltips
+              "gen3Workspaces": string // link to documentation about gen3 workspaces. Used for button tooltips.
```
mfshao, line 398: are these links subject to change from commons to commons?

The way those logic got wired up in the code implies this whole `documentationLinks` is required, but if they are not that volatile, why not just hard code them as what we have done for the `/submission` page?

src/Discovery/Discovery.tsx
```diff
@@ -316,13 +316,51 @@ const Discovery: React.FunctionComponent<DiscoveryBetaProps> = (props: Discovery
     setSearchFilteredResources(results);
   };
 
-  const handleAccessLevelChange = (ev) => {
-    const value = ev.target.value as AccessLevel;
-    setAccessLevel(value);
+  const handleExportToWorkspaceClick = async () => {
```
mfshao, line 319: I assume this may take a while if the manifest is large? Do we have something in design to inform user that the operation is pending? like a disabled button or button with spinner?

mfshao (commented):
is this thing deployed somewhere?

docs/portal_config.md
mpingram: Haha great point, I'll update this to be `public: boolean` instead of `requireLogin: boolean`

docs/portal_config.md
mpingram: I was trying to convey that this config is likely to be replaced / updated in a newer version of portal. For HEAL/BDCat rn we're working on a way to configure multiple different export options from the Disco page, I'm imagining that will mean that the config will look different.

docs/portal_config.md
mpingram: You're right, I think this is probably not worth the complexity. I'll change these from configurable values to constants declared in the file.

src/Discovery/Discovery.tsx
mpingram: I can put a spinner on the button! I don't think this operation will take too long but it's a good thing to do anyway.

src/Discovery/Discovery.tsx
mpingram: I switched up the way the table filters by access -- now it leverages builtin filtering in the antd table. So we don't need the `filterByAccessLevel` function anymore so I removed it!

src/Discovery/Discovery.tsx
```diff
@@ -406,25 +450,82 @@ const Discovery: React.FunctionComponent<DiscoveryBetaProps> = (props: Discovery
...
+                content={(<span className='discovery-popover__text'>With the Manifest File, you can use the Gen3 Client
+                to download the data from the selected studies to your local computer.</span>)}
               >
-                <Radio.Button value={AccessLevel.BOTH}>All</Radio.Button>
-                <Radio.Button value={AccessLevel.UNACCESSIBLE}><LockFilled /></Radio.Button>
-                <Radio.Button value={AccessLevel.ACCESSIBLE}><UnlockOutlined /></Radio.Button>
-              </Radio.Group>
-            </div>
+                <Button
+                  onClick={handleDownloadManifestClick}
```
mfshao, line 472: probably should add spinner to this button as well

src/Discovery/Discovery.tsx
```diff
@@ -406,25 +450,82 @@ const Discovery: React.FunctionComponent<DiscoveryBetaProps> = (props: Discovery
...
+                  type='text'
```
mfshao, line 473: I'm curious if this is an arbitrary design choice? Personally I feel the look is a bit weird if this button has no border but the one next to it has

mfshao (approved):
very cool feature 👍

src/Discovery/Discovery.tsx
mpingram: The intention was to de-emphasize the button! We're actually planning a cleanup / style update of the Disco page for HEAL so I can take a second look at the button style then.

ZakirG (approved):

PR #841, opened 2021-04-26
fix: clear npm cache

mfshao:
Clear npm cache arbitraryly at the end of docker image building process, try to mitigate portal stuck at start up issues

### DIFF
--- a/package.json
+++ b/package.json
@@ -35,6 +35,8 @@
     "fuse.js": "^3.3.0",
     "graphiql": "~0.13.0",
     "graphql": "^14.1.1",
+    "graphql-language-service-interface": "2.8.2",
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -35,6 +35,8 @@
     "fuse.js": "^3.3.0",
     "graphiql": "~0.13.0",
     "graphql": "^14.1.1",
+    "graphql-language-service-interface": "2.8.2",
+    "graphql-language-service-parser": "1.9.0",
```
mfshao, line 39: have to pin this now because there is a regression in latest releases of these and cause webpack to compain to having suitable loaders

see issue 1832 in the `graphiql` repo

mpingram (approved):
LGTM!

PR #843, opened 2021-04-28
Fix/node mem

mfshao:
Setting `max-old-space-size` earlier, before getting into npm commands
Install `strace` for debugging

### DIFF
--- a/runWebpack.sh
+++ b/runWebpack.sh
@@ -140,6 +140,7 @@ bash custom/customize.sh
 # see https://nodejs.org/api/cli.html#cli_max_old_space_size_size_in_megabytes
 export NODE_OPTIONS='--max-old-space-size=3584'
 echo "INFO: NPM config"
+npm config set loglevel verbose
 npm config list
 
 # workspace bundle does not currently deploy sheepdog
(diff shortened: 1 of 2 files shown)


themarcelor (approved):
+1

runWebpack.sh
```diff
@@ -136,6 +136,9 @@ fi
 # on the current APP environment - ugh
 #
 bash custom/customize.sh
+
+# see https://nodejs.org/api/cli.html#cli_max_old_space_size_size_in_megabytes
+export NODE_OPTIONS='--max-old-space-size=3584'
```
jawadqur, line 141: Can we set it closer to 8g? Since the pod has a 8g limit? Or do we want to keep it at 4?

jawadqur (approved):

runWebpack.sh
mfshao: i think it is still 4G in our current deployment https://github.com/uc-cdis/cloud-automation/blob/master/kube/services/portal/portal-deploy.yaml#L79

mfshao:
no longer needed

PR #844, opened 2021-04-28
chore(debugging): Adding debugging lines to investigate data-portal bootstrapping issues

themarcelor:
Adding debugging lines to investigate data-portal bootstrapping issues

### Improvements
- Adding debugging lines to investigate data-portal bootstrapping issues


themarcelor:
Closing it as we don't need these debugging tools now.

PR #846, opened 2021-05-05
Add export-pfb-to-workspace to config doc

paulineribeyre:


mfshao (approved):
There is another markdown doc dedicated for this feature, do you want to mention that in there?

paulineribeyre:
@mfshao yeah i thought it's fine like this, just wanted to add this button type to the list so it's exhaustive

PR #849, opened 2021-05-12
PXP-7955: Ecosystem Browser UI

ZakirG:
Jira Ticket: [PXP-7955](https://ctds-planx.atlassian.net/browse/PXP-7955)

This PR adds an Ecosystem Browser to data-portal. Deployed here: https://zakir.planx-pla.net/discovery

### New Features
- Adds a new featureFlag, `discoveryUseAggMDS`, which is false by default. When set to true, it alters the behavior of the Discovery page to function as an Ecosystem Browser which allows users to browse data from multiple commons. That data is supplied by a new path in the metadata service, an "aggregate" path. See `docs/ecosystem_browser.md` for more information.

### Improvements
- Discovery page table improvements: Improves column width stability when applying filters on the Discovery page. Adds placeholder text to search box.
- Refactors Discovery page into modularized subcomponents: DIscoveryMDSSearch, DiscoveryListView, DiscoveryTagViewer, DiscoverySummary.
- Adds new configurable attribute to DiscoveryConfig studyColumn block called `hrefValueFromField`. If this attribute is present, the text in the column will be linked. The href value of the link will be the corresponding value of the fieldname in this attribute.

### Dependency updates
- New featureFlag: `discoveryUseAggMDS`. It is false by default. When set to true, it alters the behavior of the Discovery page to function as an Ecosystem Browser.


src/Discovery/aggMDSUtils.jsx
```diff
@@ -0,0 +1,124 @@
+import { discoveryConfig, hostname } from '../localconf';
+
+const aggMDSURL = `${hostname}mds/aggregate`;
+const aggMDSDataURL = `${aggMDSURL}/metadata`;
```
mfshao, line 4: minor: I'd prefer setting these 2 URLs in `localconf` and import them into here, just like we do for all other pages

src/Discovery/aggMDSUtils.jsx
```diff
@@ -0,0 +1,124 @@
...
+  const commons = Object.keys(metadataResponse);
+
+  let allStudies = [];
+  for (let j = 0; j < commons.length; j += 1) {
+    const commonsName = commons[j];
+    // Now retrieve field mappings for each commons
+    // eslint-disable-next-line no-await-in-loop
+    const fieldMapping = await retrieveFieldMapping(commonsName);
+    // eslint-disable-next-line no-await-in-loop
+    const commonsInfo = await retrieveCommonsInfo(commonsName);
```
mfshao, line 51: probably better if we just break this part into another for loop so we can use `await Promise.all(...)` for each of these so we don't do await in loop

src/Discovery/aggMDSUtils.jsx
```diff
@@ -0,0 +1,124 @@
...
+      x.commons = commonsName;
+      x.frontend_uid = `${commonsName}_${index}`;
+
+      // let shortNameKey = fieldMapping.short_name || 'short_name';
+      x.name = x[fieldMapping.short_name];
+      if (fieldMapping.short_name === 'authz' && Array.isArray(x.name)) {
+        x.name = x.name[0].split('/').slice(-1)[0];
+      }
+      if (fieldMapping.short_name === 'auth_resource_path') {
+        x.name = x.name.split('/').slice(-1)[0];
```
mfshao, line 71: why is this needed?

src/Discovery/aggMDSUtils.jsx
```diff
@@ -0,0 +1,124 @@
...
+      const keys = Object.keys(entry);
+      const studyId = keys[0];
+
+      const entryUnpacked = entry[studyId].gen3_discovery;
+      const x = { ...entryUnpacked };
+      x.study_id = studyId;
+      x.commons = commonsName;
+      x.frontend_uid = `${commonsName}_${index}`;
+
+      // let shortNameKey = fieldMapping.short_name || 'short_name';
```
mfshao, line 65: unused code?

src/Discovery/aggMDSUtils.jsx
```diff
@@ -0,0 +1,124 @@
...
+      }
+      if (fieldMapping.short_name === 'auth_resource_path') {
+        x.name = x.name.split('/').slice(-1)[0];
+      }
+      x.short_name = x.name;
+      x.full_name = x[fieldMapping.full_name];
+      x._subjects_count = x[fieldMapping._subjects_count];
+      x.study_description = x[fieldMapping.study_description];
+      x._unique_id = `${commonsName}_${x._unique_id}_${index}`;
+      x.commons_url = commonsInfo.commons_url;
```
mfshao, line 78: it seem that we have many MDS fields that has speical purposes, do we have a list or doc somewhere to say which of these fields are "mush-haves" and what is their functionailties?

mfshao (commented):
overall looks impressive, but I'd feel more comfortable if @mpingram can take another round of 👀

src/Discovery/DiscoveryMDSSearch.tsx
```diff
@@ -0,0 +1,34 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import {
+  Input,
+} from 'antd';
+import { SearchOutlined } from '@ant-design/icons';
+
+class DiscoveryMDSSearch extends React.Component {
+  render() {
```
williamhaley, line 9: Since this only has a render() should it be a functional component? Or do we generally avoid them for some reason?

src/Discovery/Discovery.css
```diff
@@ -49,21 +50,34 @@
...
 
+.discovery-header__stats-wrapper {
+  display: inline-block;
+}
+
+.discovery-header__stats-wrapper#Studies {
+  width: 65px;
+}
+
+.discovery-header__stats-wrapper#TotalSubjects {
```
williamhaley, line 61: The class before the id selector gives us some context, but it's moot due to the id. Do we need to bother with the BEM style selector? Really I'd rather not see us use ids, but either way seems redundant.

data/config/brh.json
```diff
@@ -0,0 +1,840 @@
+{
```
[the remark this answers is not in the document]
williamhaley: Yeah, that's a good point. It made sense to have this in GH while we were working on feature branches, but we probably don't want to merge this in to master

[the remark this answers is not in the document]
williamhaley: Although, I do see we have a precedent for other commons to have their config here https://github.com/uc-cdis/data-portal/tree/master/data/config

I'll try to figure out if this is appropriate or not. I feel like we had this convo before at some point, but I can't remember the outcome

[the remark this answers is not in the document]
ZakirG: it's true that the data/config JSON files in data-portal are often duplicates of configuration data from gitops-qa. but i find them convenient for local development -- to test the ecosystem browser locally, it's helpful to have a config file so that you can set `APP=brh` following the precedent set by other commons ~

src/Discovery/aggMDSUtils.jsx
ZakirG: For certain commons, studies don't have a "short name" field. In those cases, the auth resource path's suffix a good descriptor of the short name of the record. So when the auth resource path field is provided to the UI from the aggregate metadata service's field mapping endpoint, that line retrieves the suffix of the value provided

williamhaley (approved):
👍

src/Discovery/aggMDSUtils.jsx
ZakirG: we don't have a doc, and the field-translations appear to be a bit ad-hoc and case-by-case per commons. the actual field mappings are hardcoded in a config file in the Metadata Service, so adding a doc to this repository likely won't be helpful to any end-users nor commons administrators. but lmk if you feel differently ~

src/Discovery/Discovery.tsx
```diff
@@ -265,7 +228,8 @@ const Discovery: React.FunctionComponent<DiscoveryBetaProps> = (props: Discovery
       }],
       onFilter: (value, record) => record[accessibleFieldName] === value,
       ellipsis: false,
-      width: undefined,
+      width: '106px',
```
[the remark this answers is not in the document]
ZakirG: this is a good point that you make. but i'll push back and say -- there are plenty of hardcoded, non-configurable pixel widths in all of our CSS files. some hardcoding is necessary to present a standardized UI, including for responsive UIs. in this case, i hardcoded a column width for one column that is unlikely to shift so that other columns are allowed to shift freely (this is a limitation of the antd table library; some columns must have fixed widths https://ant.design/components/table/#components-table-demo-fixed-header ). if a use case arises that requires configurability of this value, making it configurable will be straightforward. but for now, i think this hardcoded value is appropriate in this context

mfshao (approved):

PR #850, opened 2021-05-14
remove bhc

mfshao:
Remove BHC from schema validation scripts


testSchema.sh
```diff
@@ -6,11 +6,11 @@ set -e
 declare -a testCases
 testCases=(
 bpa "data.bloodpac.org"
-bhc "data.braincommons.org"
-gtex "dcp.bionimbus.org"
+gtex "gen3.biodatacatalyst.nhlbi.nih.gov"
```
williamhaley, line 9: Just confirming that this `gtex` change was intentional 👀 while removing the `bhc` stuff

williamhaley (approved):
lgtm. One comment on a change that may be unrelated

testSchema.sh
mfshao: Yeah I feel we may need it. Because we've changed gtex to bdcat? If not I'll just change it back 😅

williamhaley: 👍

PR #854, opened 2021-05-21
(COV-766): style/prcResponsive: make map respond better to larger and smaller browser widths

ocshawn:
Jira Ticket: [COV-766](https://occ-data.atlassian.net/browse/COV-766)

### Improvements
- make PRC main map respond better to larger and smaller browser widths

### DIFF
--- a/src/Covid19Dashboard/Covid19Dashboard.less
+++ b/src/Covid19Dashboard/Covid19Dashboard.less
@@ -77,7 +77,6 @@
 .map-chart {
   width: 100%;
   flex-basis: 200%;
-  //max-width: 1080px; set for limiting world map to one set
   height: 55vw;
 }
 .map-chart-il {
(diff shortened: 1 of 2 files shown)


src/Covid19Dashboard/Covid19Dashboard.less
```diff
@@ -77,7 +77,30 @@
 .map-chart {
   width: 100%;
   flex-basis: 200%;
-  max-width: 1080px;
+  //max-width: 1080px; set for limiting world map to one set
+  height: 55vw;
```
williamhaley, line 81: I don't have any context for how this is implemented so feel free to ignore me 😄, but is this supposed to be `vh`, or is the height relative to the width? I wonder if this is a case where a grid based layout would help?

ocshawn: it is supposed to be vw, it is used it to keep the same relative aspect ratio for the map as the screen size changes. I dont think grid would be appropriate as i am not stacking elements, only have one row and need them to flex that is why i used flex box

ocshawn:
can view in qa https://qa-covid19.planx-pla.net/

mfshao (approved):

PR #855, opened 2021-05-28
(COV-843): feat/chartPopup: add popup to charts

ocshawn:
Jira Ticket: [COV-843](https://occ-data.atlassian.net/browse/COV-843)
### New Features
- add popup to charts on PRC homepage

### DIFF
--- a/src/Covid19Dashboard/ChartCarousel/index.jsx
+++ b/src/Covid19Dashboard/ChartCarousel/index.jsx
@@ -34,7 +34,7 @@ class ChartCarousel extends PureComponent {
   }
 
   onChartClick = (chartConfig) => {
-    if (!this.props.popupOnClick) {
+    if (!this.props.enabledPopupOnClick) {
       return;
     }
 
@@ -115,7 +115,7 @@ class ChartCarousel extends PureComponent {
         onMouseOut: () => this.onChartHover(null),
       };
       // If chart is clickable
-      if (this.props.popupOnClick) {
+      if (this.props.enabledPopupOnClick) {
         chartContainerDivProps.onClick = () => this.onChartClick(chartConfig);
         chartContainerDivProps.role = 'button';
         chartContainerDivProps.tabIndex = 0;
@@ -182,7 +182,7 @@ class ChartCarousel extends PureComponent {
                   {...this.props}
                   chartsConfig={[this.state.popupChart.config]}
                   isInPopup
-                  popupOnClick={false}
+                  enabledPopupOnClick={false}
                 />
               </Popup>
               : null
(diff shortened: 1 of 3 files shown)


ocshawn:
code in QA https://qa-covid19.planx-pla.net/

src/Covid19Dashboard/index.jsx
```diff
@@ -285,6 +285,7 @@ class Covid19Dashboard extends React.Component {
                         key={i}
                         chartsConfig={carouselConfig}
                         {...this.props}
+                        popupOnClick
```
paulineribeyre, line 288: did you mean for this to be parametrizable through the config or something?

paulineribeyre (approved):

src/Covid19Dashboard/ChartCarousel/index.jsx
```diff
@@ -140,6 +171,22 @@ class ChartCarousel extends PureComponent {
...
+            this.state.popupChart ?
+              <Popup
+                title={this.state.popupChart.title}
+                onClose={() => this.closePopupChart()}
+              >
+                <ChartCarousel
+                  {...this.props}
+                  chartsConfig={[this.state.popupChart.config]}
+                  isInPopup
+                  popupOnClick={false}
```
ZakirG, line 185: Why is the value of this property hardcoded to false here? Also -- I think the property could be named better to demonstrate what it toggles. For example, `popupOnClickEnabled` perhaps

src/Covid19Dashboard/index.jsx
ocshawn: it is hard-coding it to be true in in instance, it is not configurable 

would you prefer i write it out? to make it more obvious what it does

src/Covid19Dashboard/ChartCarousel/index.jsx
ocshawn: because i am pulling the properties from the parent with  {...this.props} i need to make sure that i don't enable popups inside of popups
And i need to include  the properties from the parent because sometimes other properties are used for some of the diverse graphs 

I will think more about using a better variable name

ZakirG (approved):

src/Covid19Dashboard/index.jsx
paulineribeyre: it makes sense now that i read your answer to Zakir's comment!

src/Covid19Dashboard/ChartCarousel/index.jsx
ocshawn: ended up going with `enablePopupOnClick`

PR #858, opened 2021-06-03
(COV-933): Feat/covid 19 map overlay standalone

ocshawn:
Jira Ticket: [COV-933](https://occ-data.atlassian.net/browse/COV-933)
### New Features
- add Gen3 Map Template view and instructions 
- add multiple filters to PRC IL map

- missing map json data for standalone


paulineribeyre (approved):

src/Covid19Dashboard/standalone/README.md
```diff
@@ -0,0 +1,176 @@
...
+.
+.
+.
+ℹ ｢wdm｣: Compiled successfully.
+
+```
+You can access the url shown above
+(the port number might be different) and should see something
+like:
+![Illinois map with population](img/Map1.png)
```
paulineribeyre, line 96: not sure why this is in a code block, it looks weird and the image isn't displayed

ocshawn: found the error, it is further above at line 57 will fix

PR #865, opened 2021-06-17
Fix/508 compliance patches

ZakirG:
Jira Tickets: [PXP-8000](https://ctds-planx.atlassian.net/browse/PXP-8000) and  [PXP-7996](https://ctds-planx.atlassian.net/browse/PXP-7996)

Changes deployed here: https://zakir.planx-pla.net

This pull request:
- Ensures all text-contrast ratios across the site meet the WCAG AA standard for accessibility. You can verify contrast ratios using this tool: https://webaim.org/resources/contrastchecker/
- Ensures modals are navigable using the keyboard alone. The npm focus-trap module is added to ensure that underlying content cannot gain focus when a modal is open.

### Improvements
- Text-contrast ratios across the site now meet the WCAG AA standard for accessibility.
- Modals are now navigable using the keyboard alone. Underlying content will not gain focus when a modal is open.


src/components/layout/TopIconButton.jsx
```diff
@@ -2,13 +2,13 @@ import React from 'react';
...
 
-const TopIconButton = ({ name, icon, onActiveTab = () => {}, isActive = false }) => (
+const TopIconButton = ({ name, icon, onActiveTab = () => {}, isActive = false, tabIndex = '0' }) => (
   <div
     className={isActive ? 'top-icon-button button-top-active body-typo' : 'top-icon-button body-typo'}
     onClick={onActiveTab}
     onKeyDown={onActiveTab}
     role='button'
-    tabIndex={0}
+    tabIndex={tabIndex}
```
mfshao, line 11: why this change (and all the changes above with setting this value to `-1`)?

ZakirG: This line you commented on -- it's propagating the new tabIndex property I added into the div. The new property I added sets this value to -1 because the tab order is going to be using the link rather than the button. Before my change, tabbing the top bar nav buttons had two tabs per button. Now there's only one tap per button. If you test out my new tab order, it works better :D

ZakirG: The "tab order" refers to the user experience while navigating the page using just tab, shift+tab, and enter.

mfshao: i wonder if we should default it to `-1`, but minor

mfshao:
looks cool, I don't know why the lock file has changed so much. Did you use `--legacy-peer-deps` with `npm i`?

mfshao (approved):
🚀

PR #866, opened 2021-06-22
fix: popup mask

mfshao:
### Bug Fixes
- fixed a bug causing popup mask to disappear for user logout popups


ZakirG (approved):
Works perfectly, thank you!

PR #868, opened 2021-06-23
Fix/508 compliance patches

ZakirG:
Jira Tickets: 
Keyboard navigability of expandible/collapsible items: 
[PXP-8001](https://ctds-planx.atlassian.net/browse/PXP-8001)

Expand/collapse state defined with aria attributes: 
[PXP-7999](https://ctds-planx.atlassian.net/browse/PXP-7999)

Roles for non-button-tag buttons properly defined: 
[PXP-7998](https://ctds-planx.atlassian.net/browse/PXP-7998)

Method to bypass blocks of repetitive content: 
[PXP-7995](https://ctds-planx.atlassian.net/browse/PXP-7995)

Focus outlines on clickable elements: 
[PXP-7994](https://ctds-planx.atlassian.net/browse/PXP-7994)

Fix all HTML validator errors:
 [PXP-7992](https://ctds-planx.atlassian.net/browse/PXP-7992)

Text resizability:
[PXP-7990](https://ctds-planx.atlassian.net/browse/PXP-7990)

This PR wraps up the 508 compliance task set, making the site more accessible.

### New Features
- The Data Explorer and Discovery page now have accessibility links at the top of the pages that allow for more accessible keyboard navigation, bypassing blocks of repetitive content to tab over to areas of interest.

### Improvements
- Text can now be resized using browser text size settings.
- aria-selected and aria-expanded roles are filled in and maintained across the site for improved accessibility.
- Focus outlines are now visible on all clickable elements.
- Resolved keyboard trap issue in the Query page code editor. Users can unfocus from a textarea by pressing the Esc key.
- Page titles are now correctly populated using the gitops.json `appName` attribute.


webpack.config.js
```diff
@@ -64,7 +53,7 @@ const plugins = [
     },
   }),
   new HtmlWebpackPlugin({
-    title,
+    title: configFile.components.appName,
```
mfshao, line 56: I think we need a default value for this (before it would be `Generic Data Commons`)?

src/Covid19Dashboard/ChartCarousel/ChartCarousel.less
```diff
@@ -45,7 +45,7 @@
 /** override some style for carousel's arrow and dots **/
 
 .slick-dots li button:before {
-  font-size: 20px;
+  font-size: 1.25rem;
```
mfshao, line 48: I didn't check all the px<-->rem conversions but the page LGTM, are you sticking to the baseline of `16px == 1rem`?

src/Discovery/Discovery.tsx
```diff
@@ -293,6 +348,13 @@ const Discovery: React.FunctionComponent<DiscoveryBetaProps> = (props: Discovery
     <div className='discovery-container'>
       { (config.features.pageTitle && config.features.pageTitle.enabled)
       && <h1 className='discovery-page-title'>{config.features.pageTitle.text || 'Discovery'}</h1>}
+      <div className='g3-accessibility-links' id='discovery-page-accessibility-links'>
```
mfshao, line 351: Do we have a design for these accessibility links? I was hope it can have a label or something say "hey this is a group of accessibility links and you can use them to navigate to different compoments", because it has been a bit of confusing when I first saw them

Maybe an help icon with tooltip?

mfshao: oh and why do we only have accessibility links for Discovery and Explorer but not other pages?

mfshao (commented):
looks nice, I do have some small questions

src/Covid19Dashboard/ChartCarousel/ChartCarousel.less
ZakirG: yes that's right!

src/Discovery/Discovery.tsx
ZakirG: that's a great idea, i'll add a tooltip. and -- the reason is that the 508 standard only requires them in areas where there are blocks of repeated content so that a keyboard navigator can bypass those blocks. those two pages have the most repeated content to skip through, others don't have much below the primary area (profile page api keys, query page)

ZakirG: there's no design, but i have a reference here: https://www.w3.org/TR/UNDERSTANDING-WCAG20/navigation-mechanisms-skip.html    the part where it says "On this page:"

package.json
```diff
@@ -58,6 +58,7 @@
     "pluralize": "^8.0.0",
     "prop-types": "^15.6.0",
     "query-string": "^6.3.0",
+    "rc-tooltip": "^3.7.3",
```
mfshao, line 61: any reason we cannot use antd Tooltip but need to install rc-tooltip?

ZakirG: i added rc-tooltip because gen3-ui-component uses rc-tooltip, i'm matching the popup style. also, the antd tooltip styling clashes with the navbar tooltips, which have a darker outline, arrow tip, white background, no box shadow. the antd tooltip has box shadow and looks much different. so in healpreprod, where navbar tooltips are enabled, the accessibility tooltip appearance clashes (it's right near the navbar icons)

mfshao (approved):
nice work! 👍

PR #869, opened 2021-06-29
Feat/heal bundle no home

haraprasadj:
HEAL bundle with home page defaulting to Discovery

The only files to be reviewed are:
src/healIndex.jsx 
webpack.config.js

If they are fine, I will merge those into the branch feat/heal-mvp which is currently deployed to HEAL preprod


haraprasadj:
created another one cleanly

PR #870, opened 2021-06-29
Feat/heal bundle no home

haraprasadj:
### New Features
HEAL bundle with no home page (defaulting to Discovery page)


src/healIndex.jsx
```diff
@@ -0,0 +1,450 @@
...
+                      )
+                    }
+                  />
+                  <Route
+                    exact
+                    path='/'
+                    component={
+                      props => (
+                        <ProtectedContent
+                          public={indexPublic}
```
mfshao, line 140: maybe you will want to modify the `public` policy so it is the same with discovery page

mfshao (approved):

mfshao (commented):
and I see you are just changing the component that home page is pointing to but not removing an other components. I'm not sure if that is the requirement for HEAL but I can defer to @mpingram

src/healIndex.jsx
haraprasadj: can I just remove this - `public={indexPublic}`?

haraprasadj:
> and I see you are just changing the component that home page is pointing to but not removing an other components. I'm not sure if that is the requirement for HEAL but I can defer to @mpingram

right, that is a bigger change, we should confer with @mpingram about that

PR #873, opened 2021-07-02
PXP-8419 fix: input target

mfshao:
Jira Ticket: [PXP-8419](https://ctds-planx.atlassian.net/browse/PXP-8419)

Submission form is crashing when receving input due to trying to asynchronously accessing the property 'target' on a 
released/nullified synthetic event, hence causing page to crash

This is to fix this bug

### Bug Fixes
- Fixed a bug causing submission page to crash when using form submission

### DIFF
--- a/src/Submission/SubmitForm.jsx
+++ b/src/Submission/SubmitForm.jsx
@@ -31,8 +31,8 @@ class SubmitForm extends Component {
   };
 
   onChange = (event) => {
+    const { target } = event;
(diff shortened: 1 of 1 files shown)


ZakirG (approved):
Reproduced the bug at preprod.gen3.biodatacatalyst.nhlbi.nih.gov; tested this PR locally with dev.html against preprod; verified the observed bug has been resolved.

PR #874, opened 2021-07-05
Add tickbox support for files in exploration page, so that manifest.json can be further filtered

eirianop:
### New Features
User has the ability to add tickboxes using gitops.json configuration, it works just like linkFields. When tickboxes are enabled for files, users can arbitrarily pick files to include in the manifest.json download. Other buttons were not impacted (although would be nice in the future).


eirianop:
The explorer page looks a little funny, and I would love some help getting "Select All" to show up properly above the table.

The reducers logic is complicated because I did not want to modify the ExplorerTable/index.jsx too much. Basically, I have considered two "modes:" selecting and deselecting. If selecting mode is enabled, no items are ticked except the ones the user wants in manifest.json. You enter deselecting mode by toggling "select all" and then deselecting items you do not want. This way, the state only contains the items the user has interacted with and can filter all the files that would normally be included in manifest.json accordingly.

Importantly, I have only implemented this for manifest.json. It should be straightforward to extend this to other button types. 

This feature has been tested manually by myself and Jun Huh.

eirianop:
🔨

eirianop:
I can see neither the Travis CI - Branch step nor the continuous-intnegration/jenkins/pr-merge step. If there are errors that need to be resolved please grant me access so that I can see the logs.

Thanks!

src/GuppyDataExplorer/ExplorerTable/index.jsx
```diff
@@ -147,6 +157,18 @@ class ExplorerTable extends React.Component {
               />
             )
             : null;
+        case this.props.tableConfig.tickBox.includes(field) && field:
+          return valueStr ?
+            <CheckBox
+              id={valueStr}
+              item={valueStr}
```
mfshao, line 164: this should be a json object

src/GuppyDataExplorer/ExplorerTable/index.jsx
```diff
@@ -353,6 +375,14 @@ class ExplorerTable extends React.Component {
             </div>
           )}
 
+      <div>
+        <CheckBox
```
mfshao, line 379: can you supply a dummy ID to this `CheckBox` since that is a required props for that component

src/GuppyDataExplorer/ExplorerTable/ExplorerTickBox/reducers.js
```diff
@@ -0,0 +1,95 @@
...
+        // does not matter what items were previously selected
+        return getSelectAllState();
+      }
+
+    case 'TOGGLE_TICKBOX':
+      if (action.key in state.filteredItems) {
+        newToggleValue = !state.filteredItems[action.key];
+      }
+      // check for cases equivalent to toggling SELECT ALL
+      Object.keys(state.filteredItems).forEach(function(key) {
```
mfshao, line 44: why don't just check if this `keys` array has a length 0 here?

src/GuppyDataExplorer/ExplorerTable/index.jsx
```diff
@@ -353,6 +375,14 @@ class ExplorerTable extends React.Component {
...
+          isEnabled={true}
+          isSelected={this.props.allSelected}
+          onChange={() => this.props.addToggleAll()}
+        /> Select All
```
mfshao, line 383: lets add some css here so the checkbox and label are on the same row

mfshao (commented):
Pretty impressive work, there are some eslint issues but I think you can fix all of them by running `npm run eslint`

Also please note when we were implementing this table in the first place the latest version of `react-table` is v6 so we have to add these kind of checkboxs manually. But since them they have released `react-table v7` which includes the support of a bunch of new features and one of them is `row-selection` https://react-table-v7.tanstack.com/docs/examples/row-selection

If we upgrade to v7 then hopefully it can makes this PR simpler

ocshawn:
cleaning up old prs please reopen if this still needs to be addressed

PR #880, opened 2021-07-08
Add commons_url to manifest if both exists

craigrbarnes:
Jira Ticket: [PXP-8418](https://ctds-planx.atlassian.net/browse/PXP-8418)

### New Features
If Discovery is in Ecosystem mode and the study item contains a commons_url field, add "common_url" to the download and export manifest.

### DIFF
--- a/src/Discovery/DiscoveryListView.tsx
+++ b/src/Discovery/DiscoveryListView.tsx
@@ -62,9 +62,13 @@ export const DiscoveryListView: React.FunctionComponent<DiscoveryListViewProps> 
     }
     // combine manifests from all selected studies
     const manifest = [];
-    props.selectedResources.forEach((study) => {
+    props.selectedResources.forEach((study: any) => {
       if (study[manifestFieldName]) {
-        manifest.push(...study[manifestFieldName]);
+        if ('commons_url' in study) { // HEAL addition to allow hostname based DRS in file access clients
+          manifest.push(...study[manifestFieldName].map((x) => ({ ...x, commons_url: study.commons_url })));
+        } else {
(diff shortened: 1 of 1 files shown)


src/Discovery/DiscoveryListView.tsx
```diff
@@ -62,9 +62,13 @@ export const DiscoveryListView: React.FunctionComponent<DiscoveryListViewProps>
     }
     // combine manifests from all selected studies
     const manifest = [];
-    props.selectedResources.forEach((study) => {
+    props.selectedResources.forEach((study: any) => {
       if (study[manifestFieldName]) {
-        manifest.push(...study[manifestFieldName]);
+        if ('commons_url' in study) { // HEAL addition to allow hostname based DRS in file access clients
```
mfshao, line 67: just to make sure wo want to hard code this and make it a convention for all MDS data rather than make it configurable?

mfshao (approved):

nmetokishlubsky (approved):

ZakirG:
Have you tested this code against some env? Which env has data that results in manifests with a `commons_url` field?

craigrbarnes:
> Have you tested this code against some env? Which env has data that results in manifests with a `commons_url` field?

I tested it locally and got the expected results.

src/Discovery/DiscoveryListView.tsx
craigrbarnes: I don't want to keep it this way. However, it's faster considering the timeframe of HEAL. Created https://ctds-planx.atlassian.net/browse/PXP-8438 to address this after the HEAL MVP.

PR #882, opened 2021-07-09
(PXP-8060): Support showing Access for multiple commons in the Ecosystem Browser

johnfrancismccann:
Jira Ticket: [PXP-8060](https://ctds-planx.atlassian.net/browse/PXP-8060)

When the discovery page is set to use the aggregate MDS, load namespaced authz mappings from the Workspace Token Service so that Access for multiple commons can be shown. This new functionality is dependent on the `/aggregate/authz` [endpoint in the Workspace Token Service](https://github.com/uc-cdis/workspace-token-service/pull/36).

Screen Shot 2021-07-09 at 11 26 07 AM

### New Features
- Support showing Access for multiple commons in the Ecosystem Browser


src/Discovery/aggMDSUtils.jsx
```diff
@@ -114,4 +114,15 @@ const loadStudiesFromAggMDS = async () => {
   return studies;
 };
 
-export default loadStudiesFromAggMDS;
+const loadAuthMappingsFromWTS = async () => {
+  const res = await fetch(`${hostname}wts/aggregate/authz`);
```
williamhaley, line 118: Do we even need `hostname` here? Could we do `/wts/...`? Not sure what the end URL looks like, but I'm assuming it's referring to the portal's own host, right?

src/Discovery/index.tsx
```diff
@@ -60,42 +60,49 @@ const DiscoveryWithMDSBackend: React.FC<{
...
+              __accessible: accessible,
+            };
+          });
+          setStudies(studiesWithAccessibleField);
+        } else {
+          setStudies(rawStudies);
+        }
+      }).catch((err) => {
+        // eslint-disable-next-line no-console
+        console.error('Error encountered while loading studies: ', err);
```
williamhaley, line 102: Minor point, but JS automatically adds a space after `,` separated values.
```suggestion
        console.error('Error encountered while loading studies:', err);
```

williamhaley (dismissed):
I think this all makes sense to me 👍 Just a couple minor/inconsequential questions

src/Discovery/aggMDSUtils.jsx
mfshao: usually we set a new variable for a path in `localconfs` and import into each individual components, rather than directly importing `hostname` everywhere

mfshao:
already merged through https://github.com/uc-cdis/data-portal/pull/1089, closing

PR #884, opened 2021-07-12
HP-269 Feat/button redirect

mfshao:
Jira Ticket: [HP-269](https://ctds-planx.atlassian.net/browse/HP-269)

`Download Manifest` and `Export to Workspace` button will change to `Login to Download Manifest` and `Login to Export to Workspace` if user has not logged in, and redirect user to login page upon click

Note the disable/enable logic of these 2 buttons has not changed, so if there is nothing to download/export, these two buttons will stay disabled

### New Features
- Discovery: `Download Manifest` and `Export to Workspace` button will change to `Login to Download Manifest` and `Login to Export to Workspace` if user has not logged in, and redirect user to login page upon click.


mpingram (approved):
LGTM

PR #888, opened 2021-07-12
HP-272 Fix/heal btn name

mfshao:
Jira Ticket: [HP-272](https://ctds-planx.atlassian.net/browse/HP-272)

### Improvements
- Discovery: make `Download Manifest` button name configurable


jawadqur (approved):
Is there a default value or do we have to supply it in config?

mfshao:
> Is there a default value or do we have to supply it in config?

yeah the the default is hard coded in there as `Download Manifest`

ZakirG (approved):

PR #890, opened 2021-07-14
HP-312 Fix/adv filter height

mfshao:
[HP-312](https://ctds-planx.atlassian.net/browse/HP-312)

Make the height of adv search filters updated better according to the table's height

### Bug Fixes
- Discovery: fix a bug causing adv filter panel to cutoff

### DIFF
--- a/src/Discovery/DiscoveryListView.tsx
+++ b/src/Discovery/DiscoveryListView.tsx
@@ -1,4 +1,4 @@
-import React, { useState } from 'react';
+import React, { useState, useEffect } from 'react';
 import { Table, Empty } from 'antd';
 import './Discovery.css';
 import { DiscoveryConfig } from './DiscoveryConfig';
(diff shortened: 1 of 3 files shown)


src/Discovery/DiscoveryListView.tsx
```diff
@@ -29,6 +39,10 @@ const DiscoveryListView: React.FunctionComponent<Props> = (props: Props) => {
       locale={{
         emptyText: <Empty image={Empty.PRESENTED_IMAGE_SIMPLE} description='No Studies' />,
       }}
+      onChange={() => {
+        // forcing calling useEffect to update adv search filter height
+        setOnHeightChange(!onHeightChange);
```
mfshao, line 44: this is going to cause the component to re-render when data in table has been updated from with the table (like by sorting or filtering on access level), just to trigger the `useEffect` hook

don't know if there is a better way to address it 🤦

ZakirG (approved):

PR #895, opened 2021-07-21
(COV-966): fix/GraphCarouselNavigation: fix carousel styling

ocshawn:
Jira Ticket: [COV-966](https://occ-data.atlassian.net/browse/COV-966)

### Bug Fixes
Fix broken carousel styling introduced in release 3.0.0

### DIFF
--- a/src/Covid19Dashboard/ChartCarousel/ChartCarousel.less
+++ b/src/Covid19Dashboard/ChartCarousel/ChartCarousel.less
@@ -42,32 +42,7 @@
   z-index: 10; // place it over the "mapbox" link...
 }
 
-/** override some style for carousel's arrow and dots **/
-
-/* dots */
-.ant-carousel .slick-dots li button {
-  position: relative;
-  background: transparent;
-  padding: 10px;
-  box-sizing: content-box;
-}
-.ant-carousel .slick-dots li.slick-active button {
-  background: transparent;
-}
-.slick-dots li button:before {
-  content: " ";
-  width: 100%;
-  height: 100%;
-  font-size: 1.25rem;
-  background-color: #000;
-  opacity: 0.3;
-  display: block;
-}
-.slick-dots li.slick-active button:before {
-  opacity: 1;
-}
-
-/* arrow */
+/** override some style for carousel's arrows **/
 
 .ant-carousel .slick-arrow {
   background: var(--g3-color__smoke);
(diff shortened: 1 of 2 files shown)


src/Index/page.less
```diff
@@ -39,20 +39,3 @@
   max-height: 40vh;
   object-fit: contain;
 }
-
-.index-page .ant-carousel .slick-dots-bottom {
```
mfshao, line 43: why removing these? these are intended for the index page that other env has

src/Covid19Dashboard/ChartCarousel/ChartCarousel.less
```diff
@@ -44,24 +44,52 @@
 
 /** override some style for carousel's arrow and dots **/
 
+/* dots */
+.ant-carousel .slick-dots li button {
```
mfshao, line 48: I'd prefer to add a parent class so these changes are only applied to the carousel in covid19dashboard

src/Index/page.less
ocshawn: which environment? (so i can test) as they seem to be not needed

src/Covid19Dashboard/ChartCarousel/ChartCarousel.less
ocshawn: i think these should be global so it works the same everywhere (ill look for more environments using it and keep testing so i make sure it works everywhere)

src/Index/page.less
ocshawn: im seeing the problem ill look into fixing it everywhere

mfshao: NCT has this at least, there was 1 env that has carousel but I can't remember now 🤷‍♂️ 

FYI there is [this commit](https://github.com/uc-cdis/data-portal/pull/803/commits/40f38376d56b6567f9024e920f761165a98c19d6) that was for the discovery page

Gloabally overwriting antd styles might be causing some problems on other pages, so I'd suggest you get in touch with @mpingram , or put a scope on style overwritings

src/Covid19Dashboard/ChartCarousel/ChartCarousel.less
ocshawn: i implemented the css this way so that the buttons would be more accessible with bigger git states while still looking the same

ocshawn:
rearranged the css so that it applies the dot styling to all Carousels on the index page as far as i see this currently only applies to covid-19 but i did a mock-up and tested on niaid as well

mfshao (approved):

PR #897, opened 2021-07-23
(COV-931): feat/totalVaccinated: add vaccine total to il page

ocshawn:
Jira Ticket: [COV-931](https://occ-data.atlassian.net/browse/COV-931)

### New Features
add vaccine total to IL page

### DIFF
--- a/src/Covid19Dashboard/index.jsx
+++ b/src/Covid19Dashboard/index.jsx
@@ -459,7 +459,7 @@ Covid19Dashboard.defaultProps = {
     country: {}, state: {}, county: {}, last_updated: '',
   },
   jhuJsonByTimeLatest: { il_county_list: {}, last_updated: '' },
-  vaccinesByCountyByDate: { il_county_list: {}, last_updated: '' },
+  vaccinesByCountyByDate: { il_county_list: {}, last_updated: '', total: null },
   selectedLocationData: null,
   top10ChartData: [],
   idphDailyChartData: [],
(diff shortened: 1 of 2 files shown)


src/Covid19Dashboard/index.jsx
```diff
@@ -444,6 +459,7 @@ Covid19Dashboard.defaultProps = {
     country: {}, state: {}, county: {}, last_updated: '',
   },
   jhuJsonByTimeLatest: { il_county_list: {}, last_updated: '' },
+  vaccinesByCountyByDate: { il_county_list: {}, last_updated: '' },
```
mfshao, line 462: nitpicking: is this really the shape of `vaccinesByCountyByDate` object?

if not, better change to reflct what will be the actual shape of that object

mfshao (approved):

src/Covid19Dashboard/index.jsx
ocshawn, line 462:
```suggestion
  vaccinesByCountyByDate: { il_county_list: {}, last_updated: '', total: null },
```

ocshawn: should i put the total? see below change or leave it off, otherwise the shape is the same

mfshao: yeah I think it make sense to me to have that `total` in the default object, since you were expecting that field in code

PR #898, opened 2021-07-26
Chore/fix ecosystem tag filtering

craigrbarnes:
Fixed to address Ecosystem browser  issues with metadata from Aggregate MDS.

### Bug Fixes
* duplicate tags are removed
* subject count fields are converted to integer if type is a string
* cleaned up added fields in loadStudiesFromAggMDSRequests (x.name = x.short_name;)

### DIFF
--- a/src/Discovery/aggMDSUtils.jsx
+++ b/src/Discovery/aggMDSUtils.jsx
@@ -72,13 +72,14 @@ const loadStudiesFromAggMDSRequests = async (offset, limit) => {
       x.study_id = studyId;
       x.commons = commonsName;
       x.frontend_uid = `${commonsName}_${index}`;
-      x.short_name = x.name;
       x._unique_id = `${commonsName}_${x._unique_id}_${index}`;
       x.commons_url = commonsInfo.commons_url;
       x.tags.push(Object({ category: 'Commons', name: commonsName }));
+      x.name = x.short_name; // TODO: this will need to be refactored
 
       // If the discoveryConfig has a tag with the same name as one of the fields on an entry,
       // add the value of that field as a tag.
+
       discoveryConfig.tagCategories.forEach((tag) => {
         if (tag.name in x) {
           if (typeof x[tag.name] === 'string') {
@@ -91,9 +92,7 @@ const loadStudiesFromAggMDSRequests = async (offset, limit) => {
           }
         }
       });
-
-      x.tag = getUniqueTags(x.tags);
-
+      x.tags = [...getUniqueTags(x.tags).entries()].map((e) => (e[1]));
       return x;
     });
 
--- a/src/Discovery/aggMDSUtils.jsx
+++ b/src/Discovery/aggMDSUtils.jsx
@@ -72,14 +72,13 @@ const loadStudiesFromAggMDSRequests = async (offset, limit) => {
       x.study_id = studyId;
       x.commons = commonsName;
       x.frontend_uid = `${commonsName}_${index}`;
+      x.name = x.short_name; // TODO: this will need to be refactored
       x._unique_id = `${commonsName}_${x._unique_id}_${index}`;
       x.commons_url = commonsInfo.commons_url;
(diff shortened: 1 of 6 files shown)


src/Discovery/aggMDSUtils.jsx
```diff
@@ -72,13 +72,14 @@ const loadStudiesFromAggMDSRequests = async (offset, limit) => {
       x.study_id = studyId;
       x.commons = commonsName;
       x.frontend_uid = `${commonsName}_${index}`;
-      x.short_name = x.name;
       x._unique_id = `${commonsName}_${x._unique_id}_${index}`;
       x.commons_url = commonsInfo.commons_url;
       x.tags.push(Object({ category: 'Commons', name: commonsName }));
+      x.name = x.short_name; // TODO: this will need to be refactored
```
mfshao, line 78: I think the assumtion in the old logic is `short_name` can be optional, and make it the same as `name` if omitted
Looks like that is no longer true?

src/Discovery/aggMDSUtils.jsx
```diff
@@ -91,9 +92,7 @@ const loadStudiesFromAggMDSRequests = async (offset, limit) => {
           }
         }
       });
-
-      x.tag = getUniqueTags(x.tags);
-
+      x.tags = [...getUniqueTags(x.tags).entries()].map((e) => (e[1]));
```
mfshao, line 95: I have to say I don't really get this: if `getUniqueTags()` still can have dups in return value, why don't change that function to make it work more appropriately?

And why the old code is updating `x.tag` while this is updating `x.tags`?

src/Discovery/aggMDSUtils.jsx
craigrbarnes: There is some issue with short_name as all of the metadata returns this field. This is actually a bigger issue in that we:
* allow Ecosystem Discovery to have its fields completely configurable
or
* Define a spec for minimal set of expected fields from the metadata services

src/Discovery/aggMDSUtils.jsx
craigrbarnes: Let me refactor and put it within getUniqueTags(). x.tag is never used anywhere. I assumed the intention was x.tags

src/Discovery/aggMDSUtils.jsx
mfshao: I actually would vote for the 2nd option, as we are already having a number of "predefined meaningful" fields, and will be likely to have more, if we can have a defined "minimal set of meaningful fields", it can both simplify portal config and data ingestion

src/Discovery/aggMDSUtils.jsx
craigrbarnes: I have refactored getUniqueTags()

src/Discovery/aggMDSUtils.jsx
```diff
@@ -16,18 +16,9 @@ const retrieveCommonsInfo = async (commonsName) => {
...
-  if (tagsSoFar.has(nextTag.name)) {
-    return tagsSoFar;
-  }
-  tagsSoFar.set(nextTag.name, {
-    category: nextTag.category,
-    name: nextTag.name,
-  });
-  return tagsSoFar;
-}, new Map());
+const getUniqueTags = ((tags) => tags.filter((v, i, a) => a.findIndex((t) => (t.category === v.category && t.name === v.name)) === i));
```
mfshao, line 21: nitpicking: you probably can use `.uniq()` or `.uniqWith()` from `lodash` if that might be easier

but overall lgtm

mfshao (approved):

src/Discovery/aggMDSUtils.jsx
craigrbarnes: Good to know. I did not realize lodash was in package.json until now. If you and Zakir are ok with it as is, I will not change it.

PR #900, opened 2021-07-28
PXP-8520 Feat/paymodels

mfshao:
Jira Ticket: [PXP-8520](https://ctds-planx.atlassian.net/browse/PXP-8520)

Display user pay model information in a banner on top of the workspace page

Only visible if the current user has a pay model setup according to Hatchery

Screen Shot 2021-07-28 at 12 04 01 PM

### Improvements
- Workspace: display user pay model information (if available) in a banner on top of the workspace page
- Workspace: external login options hint banner will only display if user has any unconnected options

### Deployment changes
- The pay model support depends on this PR: https://github.com/uc-cdis/hatchery/pull/38


src/Workspace/index.jsx
```diff
@@ -350,6 +378,23 @@ class Workspace extends React.Component {
...
+      clearInterval(this.state.payModelInterval);
+    }
+    try {
+      const payModelInterval = setInterval(async () => {
+        const data = await this.getWorkspacePayModels();
+        this.setState({
+          payModel: data,
+        });
+      }, 30000);
+      this.setState({ payModelInterval });
```
mfshao, line 392: `checkWorkspacePayModel` doesn't really need to be a periodical job for now

but it will be needed once we really stores and fetches the value of cost from DB

ocshawn (approved):
looks good to me, dont know if it should refresh every 30s, but i am not familiar enough with the usage

mfshao:
> looks good to me, dont know if it should refresh every 30s, but i am not familiar enough with the usage

Yes, the current information that got displayed on this banner don't need auto-refreash, but will be needed once we have the cost/remaining balance info in

PR #903, opened 2021-07-29
fix: for jupyter kernel termination

mfshao:
We have a mechanism now to kill the kernel of the jupyter container after a certain idle time has passed.

Since it only terminates the jupyter container, the fuse sidecar container will still be running, so the pod still exists and Portal "thinks" that pod is spinning up, which is incorrect

This PR is to fix that said issue by putting the workspace launching status to `Stopped` and alert user that they should kill the pod and re-launch

Screenshot:
Screen Shot 2021-07-29 at 5 00 25 PM

### Bug Fixes
- Workspace: fixed a bug causing workspace be putting into incorrect `Launching` state if some containers has already been terminated


src/Workspace/index.jsx
```diff
@@ -412,6 +423,15 @@ class Workspace extends React.Component {
                     {(this.state.workspaceStatus === 'Launching')
                       ? <Spinner text='Launching Workspace, this process may take several minutes' />
                       : null}
+                    {(this.state.workspaceStatus === 'Stopped')
+                      ? (
+                        <div className='spinner'>
```
ocshawn, line 428: why not use the Spinner component?

ocshawn: also why is this a spinner? or is this just what is already there?

mfshao: This is just a trick to take advantage of the styles of the `Spinner` component without really displaying 3 moving dots (unfortunately our `Spinner` component doesn't have a `Stopped` state)

The ultimate goal is to replace our own `Spinner` component with antd spinner component, but I ran into some styling issue so I just tabled it for now 😞

ocshawn (approved):
looks good

PR #904, opened 2021-08-09
Fix/test schema

mfshao:
- Remove PDC from testSchems (unblock Travis)
- Use `npm ci` and new npm dep resolver in both Travis and Docker


.travis.yml
```diff
@@ -19,7 +19,7 @@ before_install:
     - npm config set maxsockets 5
 
 install:
-    - npm install --force --legacy-peer-deps
```
williamhaley, line 22: Are these two actually equivalent? I'm ➕ for `npm ci`, but just wanted to make sure we're not impacting any legacy deps/issues

williamhaley (approved):

.travis.yml
mfshao: No they are not equivalent. Ideally we should use `npm ci` everywhere in Travis and Docker. But we used to have some issues with `npm ci` failing in Travis so we were forced to stay with `npm i --force`. It seems to be resolved when we bump up Node and npm versions in Travis

As for `--legacy-peer-deps`, we put in there when we first switch to npm 7 to make sure it won't breaks. But I was testing without using `--legacy-peer-deps` when doing local dev and it seems to be stable enough so I think we should be able to switch now

PR #906, opened 2021-08-12
Add UI components and polling for zip downloads

mcannalte:
Jira Ticket: https://ctds-planx.atlassian.net/browse/HP-365

### New Features
- A user can select one or more studies to request for download. This dispatches a sower job which is monitored until it either sends back an s3 presigned url, or fails with an error description.
### Deployment Changes
- Requires the configuration of the `batch-export` sower job: https://github.com/uc-cdis/sower-jobs/pull/29 in the environment's manifest. This job also has its own setup in cloud automation: https://github.com/uc-cdis/cloud-automation/pull/1708 
- 2 new fields in `gitops.json`: `discoveryconfig.features.exportToWorkspace.[enableDownloadZip (bool) downloadZipButtonText (string)]`


nmetokishlubsky (approved):

src/Discovery/DiscoveryActionBar.tsx
```diff
@@ -27,8 +29,117 @@ interface Props {
...
+  setDownloadInProgress: (boolean) => void,
+  setDownloadStatusMessage: (object) => void,
+  manifestFieldName: string,
+  maxDownloadSizeBytes: number,
+) => {
+  const studyIDs = selectedResources.map((study) => study.project_number);
+  let downloadSize = 0;
+  selectedResources.forEach(
+    (study) => {
+      study[manifestFieldName].forEach(
```
mfshao, line 53: You need to put some checks here, not all `selectedResources` will have manifest field

src/Discovery/DiscoveryActionBar.tsx
```diff
@@ -27,8 +29,117 @@ interface Props {
...
+        (file) => {
+          downloadSize += file.file_size;
```
mfshao, line 55: What if that entry don't have `file_size`? should we query indexd?

In fact, I was hope all this file size calculation and limitation thing would happen in the backend job, and Portal as a frontend component just get whatever the messgaes from the backend job, like "error the file size is too large"

I think that is consistent with your design doc's API design section

src/Discovery/DiscoveryActionBar.tsx
```diff
@@ -27,8 +29,117 @@ interface Props {
...
+        },
+      );
+    },
+  );
+
+  if (downloadSize > maxDownloadSizeBytes) {
```
mfshao, line 61: same as pervious comment

src/Discovery/DiscoveryActionBar.tsx
```diff
@@ -27,8 +29,117 @@ interface Props {
...
+    {
+      title: 'Your download is being prepared',
+      message: 'Please remain on this page while your download is being prepared.\n\n'
+                 + 'When your download is ready, it will begin automatically. You can close this window.',
+      active: true,
+    },
+  );
+
+  const triggerDownloadResponse = await fetch(
+    '/job/dispatch',
```
mfshao, line 88: should just import the `jobAPIPath` var from localconf

src/Discovery/DiscoveryActionBar.tsx
```diff
@@ -27,8 +29,117 @@ interface Props {
...
+      method: 'POST',
+      body: JSON.stringify({
+        action: 'batch-export',
+        input: {
+          study_ids: studyIDs,
+        },
+      }),
+    },
+  );
+  const initialialJobState = await triggerDownloadResponse.json();
```
mfshao, line 99:
```suggestion
  const initialJobState = await triggerDownloadResponse.json();
```

src/Discovery/DiscoveryActionBar.tsx
```diff
@@ -27,8 +29,117 @@ interface Props {
...
+  const downloadFailedMessage = {
+    title: 'Download failed',
+    message: 'There was a problem preparing your download.'
+               + 'Please consider using the gen3 client to download these files via a manifest.',
+    active: true,
+  };
+
+  if (initialialJobState === null) {
```
mfshao, line 107: You should consider rewriting this and above section using `fetchWithCreds().then().catch()` chain, for better error handling because you will have HTTP ststus code.

For example the failure of sower job dispatching is not necessarily that sower does not have this said job configured, it could also because user don't have permission to use sower according to user.yaml

If you need a reference, see: https://github.com/uc-cdis/data-portal/blob/master/src/Indexing/Indexing.jsx#L163

src/Discovery/DiscoveryActionBar.tsx
```diff
@@ -27,8 +29,117 @@ interface Props {
...
+            active: true,
+          },
+        );
+        setDownloadInProgress(false);
+        setTimeout(() => window.open(url), 2000);
+      } else {
+        setTimeout(pollForUpdate, 5000);
+      }
+    };
+    setTimeout(pollForUpdate, 5000);
```
mfshao, line 139: use `setTimeout()` for the final `window.open()` makes sense, but for `pollForUpdate`, why don't use `setInterval()` and use the interval as a React state?

src/Discovery/DiscoveryActionBar.tsx
```diff
@@ -129,6 +241,42 @@ const DiscoveryActionBar = (props: Props) => {
...
+                    }
+                  }}
+                  type='text'
+                  disabled={props.selectedResources.length === 0 || props.downloadInProgress === true}
+                  icon={<DownloadOutlined />}
+                >
+                  {(
+                    () => {
+                      if (props.user.username) {
+                        if (props.downloadInProgress === true) {
```
mfshao, line 268: minor: you can consider adding a pending state to this button in this case https://ant.design/components/button/#components-button-demo-loading

mfshao (commented):
Can you deploy this feature somewhere so ppl can test?

Also would suggest to move the large amount of text in your current PR comment to descriptions and for those individual sections (`New features`, `Improvements`, `Bug fixes` etc), you should only keep a few brief bullet points as summerization. So it can be picked up by the `gen3-release-helper` script better when releasing

Also you should mention this feature depends on the 2 other PRs as `deployment changes`

src/Discovery/DiscoveryActionBar.tsx
mcannalte: There is also a check for this in the backend job: https://github.com/uc-cdis/sower-jobs/blob/1a922a7698e2e44e0f537141e3adf0de9db6b7f1/export_job/export_zip_from_manifest_job.py#L64

I added this additional check here because it would take a while for the user to get the error back from the job response, by comparison to just adding this extra check. 

But yes, I'll add checks to make sure `__manifest` and `file_size` are present too

mfshao: If BE and FE are doing the same thing, I would perfer to relief that duty from our FE compoment

In this scenario, after submitting the job, the compont should just trun into a `pending` state, and wait for the response from BE to figure out what to present to the user

mfshao: And: if a study don't have this `file_size` info in the manifest, should it be remove from the ID list that sends to the job?

If not, will there be discrepancy? Like the FE thinks the user should be able to download the file according to the manifest data, and BE realizes the total size is actually out of limit?

src/Discovery/DiscoveryActionBar.tsx
```diff
@@ -27,8 +29,117 @@ interface Props {
...
+        (file) => {
+          downloadSize += file.file_size;
+        },
+      );
+    },
+  );
+
+  if (downloadSize > maxDownloadSizeBytes) {
+    const maxSizeMB = (maxDownloadSizeBytes / 1000000).toFixed(1);
+    const downloadSizeMb = (downloadSize / 1000000).toFixed(1);
```
mfshao, line 63: FYI: for these 2, might be better just import the `humanFileSize` function from [https://github.com/uc-cdis/data-portal/blob/44921cc423704d8edb8da2a1467200c234d5625b/src/utils.js#L4

src/Discovery/DiscoveryActionBar.tsx
```diff
@@ -188,6 +337,22 @@ const DiscoveryActionBar = (props: Props) => {
                 {(props.user.username) ? 'Open In Workspace' : 'Login to Open In Workspace'}
               </Button>
             </Popover>
+            {
+              props.downloadStatusMessage.active
+            && (
+              <Popup
```
mfshao, line 343: Just tried a quick run in QA, looks like after succeeded, the returned presigned URL text overflows the context area of this component

Suggest to replace it with https://ant.design/components/modal/ for better layout management, and you can also turn that presigned URL from plain text into HTML link

We are trying to swap out some our old UI components from Portal codes, `Popup` is one of them

src/Discovery/DiscoveryActionBar.tsx
```diff
@@ -27,8 +29,117 @@ interface Props {
...
+
+  if (downloadSize > maxDownloadSizeBytes) {
+    const maxSizeMB = (maxDownloadSizeBytes / 1000000).toFixed(1);
+    const downloadSizeMb = (downloadSize / 1000000).toFixed(1);
+    setDownloadStatusMessage(
+      {
+        title: 'Download limit exceeded',
+        message: `
+        The selected studies contain ${downloadSizeMb} MB of data, which exceeds the download limit of ${maxSizeMB} MB.
+        Please deselect some studies and try again, or use the gen3 client.
```
mfshao, line 69:
```suggestion
        Please deselect some studies and try again, or use the Gen3 Client.
```
For consistency through out of Portal

src/Discovery/DiscoveryActionBar.tsx
```diff
@@ -27,8 +29,97 @@ interface Props {
...
+  setDownloadStatusMessage({
+    title: 'Your download is being prepared',
+    message: 'Please remain on this page while your download is being prepared.\n\n'
+               + 'When your download is ready, it will begin automatically. You can close this window.',
+    active: true,
+    url: '',
+  });
+
+  const handleJobError = (err) => {
+    console.error(err);
```
mfshao, line 59: Ususally we discourage using `console.x()` to print error messages to user, especially raw messages, unless there is an absoulte need or we are sure no sensitive info will be leaking through the print

src/Discovery/DiscoveryActionBar.tsx
```diff
@@ -27,8 +29,97 @@ interface Props {
...
+  },
+  setDownloadStatusMessage: (Object) => void;
 }
 
+const handleDownloadZipClick = async (
+  selectedResources: any[],
+  setDownloadInProgress: (boolean) => void,
+  setDownloadStatusMessage: (object) => void,
+) => {
+  const studyIDs = selectedResources.map((study) => study.project_number);
```
mfshao, line 48: I noticed that there is no more checks about manifest field exist for `selectedResources`, is BE job handling it?

src/Discovery/DiscoveryActionBar.tsx
```diff
@@ -27,8 +29,97 @@ interface Props {
...
+      const pollForJobStatusUpdate = () => {
+        fetchWithCreds({ path: `${jobAPIPath}status?UID=${uid}` }).then(
+          (statusResponse) => {
+            const { status } = statusResponse.data;
+            if (status === 'Failed') {
+              fetchWithCreds({ path: `${jobAPIPath}output?UID=${uid}` }).then(
+                (outputResponse) => {
+                  setDownloadStatusMessage({
+                    title: 'Download failed',
+                    message: outputResponse.data.output,
```
mfshao, line 91: is `data` always available in `outputResponse`? if not, this line might causing a crash

src/Discovery/DiscoveryActionBar.tsx
```diff
@@ -27,8 +29,97 @@ interface Props {
...
+                    message: 'Your download has been prepared. If your download doesn\'t start automatically, please follow this direct link:',
+                    active: true,
+                    url: output,
+                  });
+                  setDownloadInProgress(false);
+                  setTimeout(() => window.open(output), 2000);
+                },
+              ).catch(handleJobError);
+            } else {
+              setTimeout(pollForJobStatusUpdate, 5000);
```
mfshao, line 113: This will cause the component stuck in an infinite loop if `uid` is not returned from https://github.com/uc-cdis/data-portal/pull/906/files#diff-1ac68dcc0067af6b35b142c058129318cc50ee20bfa9195afdb3d99f150d9d15R81

You can reproduce it now in qa-heal, since the mock auth is turned on, you will be login by default as a test account, which doesn't have `sower` access. The job dispatch request will return as 403 error, and no `uid` is returned, but this section of code will repeatly checking with `uid=undefined`

Better to handle fetched response with HTTP status code, rather than plain text

mfshao (commented):
since you have added more config field into portal config, can you also update https://github.com/uc-cdis/data-portal/blob/master/docs/portal_config.md

src/Discovery/DiscoveryActionBar.tsx
mcannalte: Yes, I made some changes to return a more user-friendly error if the job fails, and display that: https://github.com/uc-cdis/sower-jobs/blob/ba70c21736323882735c2663639bbdf19964b3f6/export_job/export_zip_from_manifest_job.py#L67

src/Discovery/DiscoveryActionBar.tsx
mcannalte: yes, this should be guaranteed by the `fetchWithCreds` function:  https://github.com/uc-cdis/data-portal/blob/275cff44c31565020ab699a819d916342a34bab5/src/actions.js#L80

src/Discovery/DiscoveryActionBar.tsx
mcannalte: True. I will add some checks for HTTP status code for cases like this. For the `/status` calls, however, the status code from Sower is 200 regardless of job state, so I will have to keep some plain text checks in there

mfshao: make sense! probably only the first call to `/job/dispatch` needs a status code check

src/Discovery/DiscoveryActionBar.tsx
```diff
@@ -27,8 +29,96 @@ interface Props {
...
+    active: true,
+    url: '',
+  });
+
+  const handleJobError = () => {
+    setDownloadInProgress(false);
+    setDownloadStatusMessage({
+      title: 'Download failed',
+      message: 'There was a problem preparing your download.'
+                + 'Please consider using the gen3 client to download these files via a manifest.',
```
mfshao, line 63: We probably should change this to `Gen3 SDK for Python (w/ CLI)` since `Gen3 Client` doesn't support DRS URIs now and no plans to add it

ZakirG: done

ZakirG (approved):
Works well, good job ~

mfshao (changes requested):
Are we going to merge into master with this issue? https://github.com/uc-cdis/data-portal/pull/906#discussion_r688639950

mcannalte:
> Are we going to merge into master with this issue? [#906 (comment)](https://github.com/uc-cdis/data-portal/pull/906#discussion_r688639950)

I have a fix I'm working on now, which will also check for sower status codes and raise errors accordingly. I'd prefer if we can get that change in, and also test on live data in preprod before merging. It sounds like we will have adequate time to do so

mfshao (approved):
code LGTM! great work 👍

PR #907, opened 2021-08-12
Feat/heal beta user fixes

mpingram:
Discovery page fixes for HEAL beta users

- Allow sorting table by access; sort table by access by default 
- Replace n/a icon with new dash icon (looks like `---`)
- Rename Data Access column -> Data Availability
- Fix visual bug when Adv Search Filter is disabled


nmetokishlubsky (approved):

mfshao (approved):
code-wise lgtm, but it would be great if you can put those bullet points into appropriate sections in PR comment

PR #893, opened 2021-07-20
PXP-8364 fix: strict csp

mfshao:
Jira Ticket: [PXP-8364](https://ctds-planx.atlassian.net/browse/PXP-8364)

Remove CSP `connect-src: https://*.s3.amazonaws.com https://*.mapbox.com` to tighten the scope

### Improvements
- Misc: change CSP for mapbox according to [this guide](https://docs.mapbox.com/mapbox-gl-js/api/#csp-directives)
- Misc: change CSP `connect-src: https://*.s3.amazonaws.com` to `connect-src: https://<DATA_UPLOAD_BUCKET_NAME>.s3.amazonaws.com`

### Deployment changes
- Requires https://github.com/uc-cdis/cloud-automation/pull/1687, otherwsie `Indexing` page may not work properly

PR #908, opened 2021-08-12
(COV-926) Feat/total vaccinated map layer

ocshawn:
Jira Ticket: [COV-926](https://occ-data.atlassian.net/browse/COV-926)

### New Features

- add vaccine map layer 

### Improvements

- refactor map data handling logic
- update colors of Mobility Data layers


src/index.ejs
```diff
@@ -10,7 +10,7 @@
...
+    <meta http-equiv="Content-Security-Policy" content="default-src 'self' https://login.bionimbus.org https://wayf.incommonfederation.org; child-src blob:; img-src 'self' https://opendata.datacommons.io https://static.planx-pla.net data: https://*.s3.amazonaws.com https://www.google-analytics.com blob:; script-src 'self' 'unsafe-eval' https://www.google-analytics.com <%= htmlWebpackPlugin.options.script_src %>; worker-src 'self' blob:; style-src 'self' 'unsafe-inline' https://fonts.googleapis.com; object-src 'none'; font-src 'self' data: https://fonts.googleapis.com https://fonts.gstatic.com; connect-src 'self' https://*.s3.amazonaws.com https://login.bionimbus.org https://wayf.incommonfederation.org https://opendata.datacommons.io https://static.planx-pla.net https://www.google-analytics.com <%= htmlWebpackPlugin.options.connect_src %>; frame-src <%= htmlWebpackPlugin.options.connect_src %> 'self' https://auspice.planx-pla.net https://auspice.pandemicresponsecommons.org;;">
```
mfshao, line 13: This was changed by https://github.com/uc-cdis/data-portal/pull/893 as requested by the security team, since putting `https://*.s3.amazonaws.com` in CSP was identified as a vulnerability by Veracode scan

You will have either to dynamically generate a more strict entry for this CSP, or work with the security team to figure out an alternative solution

src/Covid19Dashboard/IllinoisMapChart/index.jsx
```diff
@@ -161,14 +205,13 @@ class IllinoisMapChart extends React.Component {
...
-      workingId = 'C';
-    }
     function msToDays(ms) {
       return Math.floor(ms / (1000 * 60 * 60 * 24));
     }
-    const curentDateRange = this.state.dataDateRange[workingId];
+    if (!this.state.dataDateRange[id]) {
+      return;
+    }
+    const curentDateRange = this.state.dataDateRange[id];
```
mfshao, line 214:
```suggestion
    const currentDateRange = this.state.dataDateRange[id];
```

src/Covid19Dashboard/IllinoisMapChart/index.jsx
```diff
@@ -148,6 +169,29 @@ class IllinoisMapChart extends React.Component {
 
       this.setState({ mapColors: this.mapData.colors }); // eslint-disable-line react/no-did-update-set-state, max-len
     }
+
+    if (!this.state.vaccine_data.fetchStatus && Object.entries(this.props.jsonVaccinated.il_county_list).length > 0) {
+      const vaccinegeoJson = this.addDataToGeoJsonBase(
```
mfshao, line 174:
```suggestion
      const vaccineGeoJson = this.addDataToGeoJsonBase(
```

mfshao (changes requested):

src/index.ejs
ocshawn: removed this and changed webpack.config.js to accept an array in process.env.DATA_UPLOAD_BUCKET , let me know if this solution would not work for you or if you would like any changes made

webpack.config.js
```diff
@@ -17,7 +17,13 @@ if (DAPTrackingURL) {
   connectSrcURLs.push(DAPTrackingURL);
 }
 if (process.env.DATA_UPLOAD_BUCKET) {
-  connectSrcURLs.push(`https://${process.env.DATA_UPLOAD_BUCKET}.s3.amazonaws.com`);
+  if (Array.isArray(process.env.DATA_UPLOAD_BUCKET)) {
+    // if DATA_UPLOAD_BUCKET is an array
+    connectSrcURLs.push(...process.env.DATA_UPLOAD_BUCKET.map(dataBucket => (`https://${dataBucket}.s3.amazonaws.com`)));
```
mfshao, line 22: is this going to work? isn't `process.env.XXX` will always be strings?

how are you going to mount this env vars into portal pod? is there an accompanying `cloud-automation` PR?

ocshawn: from what i see DATA_UPLOAD_BUCKET is configured in fence-config-public.yaml i was just going to set it to be an array like 
`DATA_UPLOAD_BUCKET: ['qa-covid19-databucket-gen3', covd-map-occ-prc-qa]`

the new code will then add these new values as well as not breaking the old functionality. What do you think?

mfshao: First, according to https://github.com/uc-cdis/fence/blob/e31d29d5533683d91996f909f4b364e0cf798168/fence/config-default.yaml#L591-L594 `DATA_UPLOAD_BUCKET` is just the name of one bucket, I think Fence might not like it if you set it as an array

Second, like I said `process.env.XXX` will always be strings. So even if you change `DATA_UPLOAD_BUCKET` to be an array in fence config, when passed into the Portal pod it will becomes `"['qa-covid19-databucket-gen3', 'covd-map-occ-prc-qa']"`. And you have to do something manually with `. replace()` and `.split()`

You need to test this in k8s to see if your portal pods is mounting something as you have expected, and is parsing it correctly

mfshao: And fundamentally why would you put `covd-map-occ-prc-qa` as one of Fence's `DATA_UPLOAD_BUCKET`? It doesn't sound right since we are not going to upload anything to there right? Just pulling things down?

IMO an more appropriate would be to define another optional env var like `CONNECTED_S3_BUCKET`, put it some where in the manifest (like in the `portal` section), and mount it into portal pods as env var

Example https://github.com/uc-cdis/cloud-automation/blob/master/kube/services/portal/portal-deploy.yaml#L98-L103

ocshawn: thanks for the suggestion will implement it that way, branching this off as its own separate jira COV-996

and will remove the above code from this PR

mfshao (approved):
👍

PR #911, opened 2021-08-17
(PXP-8275): Feat/hpwebform

ocshawn:
Jira Ticket: [PXP-8275](https://ctds-planx.atlassian.net/browse/PXP-8275)
### New Features
- NCT: Add a web form component to allow signing up for email lists at govdelivery.com


src/Index/page.jsx
```diff
@@ -98,6 +98,11 @@ class IndexPageComponent extends React.Component {
                 {homepageCharts}
               </Carousel>
             </MediaQuery>
+            {homepageEmailForm
```
ocshawn, line 101:
```suggestion
            test
            {homepageEmailForm
```

docs/portal_config.md
```diff
@@ -91,6 +91,14 @@ Below is an example, with inline comments describing what each JSON block config
         }
       ]
     },
+    "emailSignUpForm": { // optional; used to show a govdelivery.com Email SignUp Form on the homepage
```
mfshao, line 94: is this feature also applicable to other envs?

if this is an NCT specific thing, we should make these hardcoded, instead of keep enlarging the portal config file

src/Index/page.jsx
mfshao, line 101: you probably want to do this for the `nctIndex` page instead

NCT is on a special portal bundle

src/Index/reduxer.jsx
```diff
@@ -28,6 +29,20 @@ export const ReduxIndexBarChart = (() => {
...
+export const ReduxEmailSignUpForm = (() => {
+  const mapStateToProps = () => {
+    if (components.index.emailSignUpForm) {
+      return { ...components.index.emailSignUpForm };
+    }
+    return {};
+  };
+
+  // ReduxEmailSignUpForm does not dispatch anything
+  const mapDispatchToProps = function mapDispatch() { return {}; };
```
mfshao, line 41: if nothing to dispatch, you don't have to have this

`connect()` can also work with only 1 arg

src/Index/page.jsx
ocshawn, line 101:
```suggestion
            test
            {homepageEmailForm
```

ocshawn: thanks for commenting i was trying to get this to show and wondering why nothing was working

docs/portal_config.md
ocshawn: currently NCT specific, currently the action and subscribeToTopicId are the only required values everything else has defaults set. Do you think this would be better entirely hard coded on the nctPage? or do you just mean to have an option to disable enable it here?

mfshao: I would just hardcode the entire thing, or at most have one option to turn it on/off just in case the sponsors change their minds 🤷‍♂️ 

But if you want to have an option then I suggest to name it as `nctSignupForm` so others know it is for NCT only

ocshawn: Sorry id not understand you at first. It could be used for any govdelivery email signup, dont know if more services could use it in the future that is why i made it extremely flexible. Next time im asked to do something like this ill do it the fast way as i have already spent too much time on this

mfshao: No other env needs the govdelivery signup now and if they will need something to subscribe, it is not necessarily from govdelivery

That is why I suggest to make it less customizable, and place it only in `nctPage` for now

mfshao (approved):
💯

PR #912, opened 2021-08-18
PXP-8448 - Register users for download

rnerella92:
Description of changes
When the users are not logged in and try to download data they will be redirected to the login page. This change will come into effect if the 'registerUsersForDownload' variable is set to 'true' in gitops.json. 

### New Features
Redirects users to the login page if they try to download the data without logging in. 

### Deployment changes
Requires registerUsersForDownload flag set to true in gitops


src/GuppyDataExplorer/ExplorerButtonGroup/index.jsx
```diff
@@ -796,10 +799,12 @@ Currently, in order to export a File PFB, \`enableLimitedFilePFBExport\` must be
       btnTooltipText = 'Currently you cannot export files with different Data Types. Please choose a single Data Type from the Data Type filter on the left.';
     }
 
+
+
     return (
       <Button
         key={buttonConfig.type}
-        onClick={clickFunc}
+        onClick={() => {if (!this.isRegisterUsersForDownloadEnabled()) {clickFunc()}}}
```
mfshao, line 807: would be easier to read and understand if refactor into using ternary operators, e.g.: https://github.com/uc-cdis/data-portal/blob/master/src/Discovery/DiscoveryActionBar.tsx#L303-L306

src/GuppyDataExplorer/configTypeDef.js
```diff
@@ -46,6 +46,7 @@ export const ButtonConfigType = PropTypes.shape({
   terraExportURL: PropTypes.string,
   terraTemplate: PropTypes.arrayOf(PropTypes.string),
   sevenBridgesExportURL: PropTypes.string,
+  registerUsersForDownload: PropTypes.bool,
```
mfshao, line 49: I would rename this to something like `loginForButtonFeature` which reflects the meaning of this option more correct

Because you are only redirecting user to login page by this option, and it is depending on Fence to decide wether to show the registration form or not

src/GuppyDataExplorer/ExplorerButtonGroup/index.jsx
mfshao: Also, the implementation is applying redirect to login for all button types, not download button only

Which means all export buttons will be affected as well, you should confirm with PM this is aligned with the requirement (MIDRC don't have any export buttons for now, but it may in the future)

src/GuppyDataExplorer/configTypeDef.js
mfshao: also you should update the portal config markdown doc with this new options

mfshao (commented):
Besides each individual comments, there are some eslint errors exist

You should run `npm run eslint` and commit in the changes

src/GuppyDataExplorer/ExplorerButtonGroup/index.jsx
```diff
@@ -799,8 +800,8 @@ Currently, in order to export a File PFB, \`enableLimitedFilePFBExport\` must be
     return (
       <Button
         key={buttonConfig.type}
-        onClick={clickFunc}
-        label={buttonTitle}
+        onClick={() => { if (!this.isLoginForButtonFeatureEnabled(buttonConfig)) { clickFunc(); } }}
+        label={(this.props.user.username) ? buttonTitle : `Login to ${buttonTitle}`}
```
mfshao, line 804: this will change all button text to `Login to ...`, not only for download buttons

and for export buttons, it will not redirect, which will be very confusing

rnerella92: Updated this logic to redirect only for download buttons

src/GuppyDataExplorer/configTypeDef.js
rnerella92: Updated config variable name and added the description of the config variable to portal_config.md file

src/GuppyDataExplorer/ExplorerButtonGroup/index.jsx
rnerella92: Updated the logic to redirect only for download buttons.

src/GuppyDataExplorer/ExplorerButtonGroup/index.jsx
```diff
@@ -799,8 +800,8 @@ Currently, in order to export a File PFB, \`enableLimitedFilePFBExport\` must be
     return (
       <Button
         key={buttonConfig.type}
-        onClick={clickFunc}
-        label={buttonTitle}
+        onClick={() => { if (!this.isLoginForButtonFeatureEnabled() && this.isDownloadButton(buttonConfig)) { clickFunc(); } }}
+        label={(this.props.user.username && this.isDownloadButton(buttonConfig)) ? buttonTitle : `Login to ${buttonTitle}`}
```
mfshao, line 804: `buttonTitle` usually have first char as caps, so `Login to ${buttonTitle}` will looks like `Login to Download ...`
```suggestion
        label={(this.props.user.username && this.isDownloadButton(buttonConfig)) ? buttonTitle : `Login to ${buttonTitle.toLowerCase()}`}
```

src/GuppyDataExplorer/ExplorerButtonGroup/index.jsx
```diff
@@ -813,6 +814,30 @@ Currently, in order to export a File PFB, \`enableLimitedFilePFBExport\` must be
...
+    if ((buttonConfig.type.startsWith('data')
+    || buttonConfig.type === 'manifest'
+    || buttonConfig.type === 'file-manifest')) {
+      return true;
+    }
+    return false;
+  }
+
   render() {
     const dropdownConfigs = calculateDropdownButtonConfigs(this.props.buttonConfig);
```
mfshao, line 842: we should apply the same logic to dropdown buttons as well right?

src/GuppyDataExplorer/ExplorerButtonGroup/index.jsx
```diff
@@ -813,6 +814,30 @@ Currently, in order to export a File PFB, \`enableLimitedFilePFBExport\` must be
...
+  isLoginForButtonFeatureEnabled = () => {
+    if (this.props.buttonConfig.loginForButtonFeature) {
+      return this.goToLogin();
+    }
+    return false;
+  }
+
+  goToLogin = () => {
+    if (!this.props.user || !this.props.user.username) {
+      this.props.history.push('/login');
```
mfshao, line 826:
```suggestion
      this.props.history.push('/login', { from: `${this.props.location.pathname}` });
```
So it can get back to this page after login

But you will need to add `location` as props of this component

mfshao (commented):
Much better, but I found 1 problem when testing on qa-midrc

When I get redirected to the Login page, after login I was presented with the Fence registration page. But if I click `Maybe Later` instead of filling out the form, portal still redirects me back. And I can begin to download even if I didn't actually finishes the registration

Maybe this is a bug from your Fence branch? 🤔

src/GuppyDataExplorer/ExplorerButtonGroup/index.jsx
rnerella92: Updated this.

src/GuppyDataExplorer/ExplorerButtonGroup/index.jsx
rnerella92: I am not sure about this.

mfshao: Which part is confusing you?

`ExplorerButtonGroup` can have two types of buttons, regular button and dropdown buttons. What you are changing now will not affect dropdown buttons, because they are not generated by calling `renderButton()`

So `loginForButtonFeature` will not work on dropdown buttons. MIDRC currently don't have any of them bit other envs do. So it would be ideal if all download buttons work in the same way

Or at least make a new ticket if you think this requirement is out of scope of your ticket

src/GuppyDataExplorer/ExplorerButtonGroup/index.jsx
```diff
@@ -813,6 +816,29 @@ Currently, in order to export a File PFB, \`enableLimitedFilePFBExport\` must be
...
+    if (this.props.buttonConfig.loginForButtonFeature) {
+      return true;
+    }
+    return false;
+  }
+
+  goToLogin = () => {
+    if (!this.props.user || !this.props.user.username) {
+      this.props.history.push('/login', { from: `${this.props.location.pathname}` });
+      // this.props.history.push('/login');
```
mfshao, line 829: remove this comment

src/GuppyDataExplorer/ExplorerButtonGroup/index.jsx
```diff
@@ -864,7 +890,10 @@ Currently, in order to export a File PFB, \`enableLimitedFilePFBExport\` must be
             .map((dropdownId) => {
               const entry = dropdownConfigs[dropdownId];
               const btnConfigs = entry.buttonConfigs;
-              const dropdownTitle = entry.dropdownConfig.title;
+              // const dropdownTitle = entry.dropdownConfig.title;
```
mfshao, line 893: don't change dropdown title, you should change each individual button's text instead (if the button is a download button)

src/GuppyDataExplorer/ExplorerButtonGroup/index.jsx
```diff
@@ -881,7 +910,8 @@ Currently, in order to export a File PFB, \`enableLimitedFilePFBExport\` must be
                             key={btnCfg.type}
                             leftIcon='datafile'
                             rightIcon='download'
-                            onClick={onClick}
+                            onClick={() => ((!this.props.user.username && this.isLoginForButtonFeatureEnabled()
+                              && this.isDownloadButton(btnCfg)) ? this.goToLogin() : onClick())}
                           >
                             {btnCfg.title}
```
mfshao, line 916: change here

docs/portal_config.md
```diff
@@ -310,6 +310,7 @@ Below is an example, with inline comments describing what each JSON block config
         "rightIcon": "download"
       }
     ],
+    "loginForButtonFeature": true, //optional; redirects user to login page if they tries to download data without logging in.
```
mfshao, line 313: minor: I know we went back and forth on this, but since we are doing redirect for download buttons only, you should rename this to `loginForDownload` to be more clearly reflect what this is doing

mfshao: you also need to change here

src/GuppyDataExplorer/ExplorerButtonGroup/index.jsx
```diff
@@ -813,6 +816,29 @@ Currently, in order to export a File PFB, \`enableLimitedFilePFBExport\` must be
     );
   };
 
+  isLoginForDownloadEnabled = () => {
+    if (this.props.buttonConfig.loginForDownload) {
+      return true;
+    }
+    return false;
```
mfshao, line 823: minor: effectively, you can turn this 5 line code block into an one-liner

this is the same as `isLoginForDownloadEnabled = () => !!this.props.buttonConfig.loginForDownload`

and you can even conbine this function with the `isDownloadButton` function down below and give it a more meaningful name. Since both of them are just returning a boolean value

src/GuppyDataExplorer/ExplorerButtonGroup/index.jsx
```diff
@@ -813,6 +816,29 @@ Currently, in order to export a File PFB, \`enableLimitedFilePFBExport\` must be
...
+  }
+
+  isDownloadButton = (buttonConfig) => {
+    if ((buttonConfig.type.startsWith('data')
+    || buttonConfig.type === 'manifest'
+    || buttonConfig.type === 'file-manifest')) {
+      return true;
+    }
+    return false;
+  }
```
mfshao, line 840: This is effectively the same as 

```
  isDownloadButton = (buttonConfig) => (
    buttonConfig.type.startsWith('data')
    || buttonConfig.type === 'manifest'
    || buttonConfig.type === 'file-manifest')
```

src/GuppyDataExplorer/ExplorerButtonGroup/index.jsx
mfshao: This feedback has not been resolved yet

PR #918, opened 2021-08-30
Chore/discovery for more bundles

mfshao:
### Improvements
- Misc: add the optional `Discovery` page to `nct` and `covid19` bundles (note the `workspace` bundle still don't have it)

### Deployment changes
- Misc: add option for bundle name: `ecosystem`, which is the same as `heal` bundle under the hood


haraprasadj (approved):

src/covid19Index.jsx
```diff
@@ -381,6 +382,37 @@ async function init() {
                       />
                     )
                     : null}
+                  {isEnabled('discovery')
```
williamhaley, line 385: Does there also need to be an update to the tests? Maybe not, but just wanted to call this out in case https://github.com/uc-cdis/gen3-qa/blob/master/utils/apiUtil.js#L476

mfshao: not for now since they are not changing the homepage of these bundles

actually we are adding them because our QA team will be able to run discovery page test on other bundles, and if any of these envs needs to enable the discovery page in the future, they are ready to go

williamhaley (approved):

PR #928, opened 2021-09-30
Add tag column name configuration option

mcannalte:
Jira Ticket: https://ctds-planx.atlassian.net/browse/HP-580

### Improvements
- Add config for setting name of tags column on discovery page

### DIFF
--- a/docs/portal_config.md
+++ b/docs/portal_config.md
@@ -578,7 +578,7 @@ Below is an example, with inline comments describing what each JSON block config
         "display": true
       }
     ],
-    "tagColumnName": "Tags" // optional, overrides the name of the mandatory tags column
+    "tagsColumnName": "Tags" // optional, overrides the name of the mandatory tags column
   },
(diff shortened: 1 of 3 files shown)


mcannalte:
Note: we now have 3 places to configure the "study characteristics" label in gitops.json (the dropdown button name, the study details page, and the discovery page which this pr adds).
This isn't great, but I couldn't think of a way around it, and it seemed possible (though unlikely) that some project might want to name these different things.

xritter1 (approved):

mfshao (commented):
> This isn't great, but I couldn't think of a way around it, and it seemed possible (though unlikely) that some project might want to name these different things.

I agree with this comment. I think ideally we should change this `tagsColumnName` name to something more general like `tagsDisplayName`. And then for all the UI components that are rendering tags, we renaming its title/heading/whatever with this value if is defined. Otherwise, just display `Tags` as default value

But I'm fine with if you think this can get deferred

PR #929, opened 2021-09-30
(PXP-8570) Export PFB to arbitrary URL

mpingram:
Jira Ticket: [PXP-8570](https://ctds-planx.atlassian.net/browse/PXP-8570)

A new button in the Explorer which allows us to send a PFB of data exported from the explorer to an arbitrary URL.
Used to transfer data to external analysis platforms, such as Seven Bridges, Terra, and Cavatica.

### New Features
- Add `"export-pfb-to-url"` button to Explorer, which allows exporting PFBs to arbitrary URLs.

### DIFF
--- a/src/GuppyDataExplorer/ExplorerButtonGroup/index.jsx
+++ b/src/GuppyDataExplorer/ExplorerButtonGroup/index.jsx
@@ -813,12 +813,6 @@ Currently, in order to export a File PFB, \`enableLimitedFilePFBExport\` must be
       return this.isPFBRunning()
         && this.state.exportingToSevenBridges;
     }
-    if (buttonConfig.type === 'export-pfb-to-url') {
-      // export to seven bridges button is pending if a pfb export job is running
-      // and it's an export to seven bridges job.
-      return this.isPFBRunning()
-        && this.state.exportingPFBToURL;
-    }
     return false;
   };
 
--- a/src/GuppyDataExplorer/ExplorerButtonGroup/index.jsx
+++ b/src/GuppyDataExplorer/ExplorerButtonGroup/index.jsx
@@ -436,7 +436,7 @@ class ExplorerButtonGroup extends React.Component {
   // in the component state, create the pfb with exportToPFB(), and send the
   // pfb to the target url with `sendPFBToURL()`.
   // `sendPFBToURL()` is called from componentWillUpdate.
-  exportPFBToURL = (pfbExportTargetURL) => {
+  exportPFBToURL = (pfbExportTargetURL) => () => {
     this.setState({
       exportingPFBToURL: true,
       pfbExportTargetURL,
--- a/src/GuppyDataExplorer/ExplorerButtonGroup/index.jsx
+++ b/src/GuppyDataExplorer/ExplorerButtonGroup/index.jsx
@@ -26,6 +26,8 @@ class ExplorerButtonGroup extends React.Component {
       toasterErrorText: 'There was an error exporting your cohort.',
       exportingToTerra: false,
       exportingToSevenBridges: false,
+      exportingPFBToURL: false,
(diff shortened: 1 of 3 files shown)


src/GuppyDataExplorer/ExplorerButtonGroup/index.jsx
```diff
@@ -764,6 +815,13 @@ Currently, in order to export a File PFB, \`enableLimitedFilePFBExport\` must be
       return this.isPFBRunning()
         && this.state.exportingToSevenBridges;
     }
+    if (buttonConfig.type === 'export-pfb-to-url') {
+      return this.isPFBRunning()
+        && this.state.exportingPFBToURL
+        // because we can have multiple `export-pfb-to-url` buttons, only make the
+        // one the user clicked have the pending state.
+        && this.state.targetURLTemplate === buttonConfig.targetURLTemplate;
```
mfshao, line 823: should we allow user to click on another `export-pfb-to-url` button while there is another one running? 😕

src/GuppyDataExplorer/ExplorerButtonGroup/index.jsx
```diff
@@ -413,6 +429,37 @@ class ExplorerButtonGroup extends React.Component {
...
+  }
+
+  sendPFBToURL = (targetURLTemplate, presignedURL) => {
+    const signedURL = encodeURIComponent(presignedURL);
+    // the PFB export target URL is a template URL that should have a {{PRESIGNED_URL}} template
+    // variable in it.
+    const PRESIGNED_URL_TEMPLATE_VARIABLE = '{{PRESIGNED_URL}}';
+    const targetURLIsBad = this.state.targetURLTemplate.indexOf(PRESIGNED_URL_TEMPLATE_VARIABLE) === -1;
+    if (targetURLIsBad) {
+      throw new Error(`Misconfiguration error! An \`export-pfb-to-url\` button has a bad \`targetURLTemplate\` property. The string \`{{PRESIGNED_URL}}\` must appear in the \`targetURLTemplate\` property. Bad \`targetURLTemplate\`: ${this.state.targetURLTemplate}`);
```
mfshao, line 457: maybe better to check and throw this error in the same time when loading the button (in the same place that check if `targetURLTemplate` exists)?

src/GuppyDataExplorer/ExplorerButtonGroup/index.jsx
```diff
@@ -413,6 +429,37 @@ class ExplorerButtonGroup extends React.Component {
...
+  // This is a generic way to export a PFB to a third party, deprecating
+  // exportToTerra and exportToSevenBridges.
+  // See docs/export-pfb-to-url.md
+  // Code flow (it's confusing):
+  // 1. User clicks a 'export-pfb-to-url' button
+  // 2. Store target URL as state.targetURLTemplate
+  // 2. Kick off a PFB export job with this.exportToPFB()
+  // 3. componentDidUpdate polls the exportToPFB job status, when complete it will call
+  //    sendPFBToURL(this.state.targetURLTemplate)
+  exportPFBToURL = (targetURLTemplate) => () => {
```
mfshao, line 441: I hope since this is general enough, that we can remove the support and related codes/configs for `exportingToTerra` and `exportingToSevenBridges` and use this general solution instead

It will make our code and config a lot cleaner. And I think BDCat is the only project that is deploying those 2 buttons into prod, so it shouldn't break a lot of stuff 😅

mpingram: Yes, we can replace the exportToTerra and exportToSevenBridges code/configs with this new feature! BDC and Anvil both use the export to Terra / SBG feature in prod.

src/GuppyDataExplorer/ExplorerButtonGroup/index.jsx
mpingram: Good idea, I'll make the change.

PR #942, opened 2021-11-02
Discovery Page as a Fully Controlled Component

mcannalte:
### New Features
- persist discovery page state when navigating tabs
- reload prior discovery page state and download/export action after logging in

### Bug Fixes
- keep tags filters consistent between discovery table and dropdown tag viewer
- correctly handle job status response when status code is 200 but user isn't logged in

### Improvements
- don't require clicking "OK" to filter by data accessibility
- redirect back to `dev.html` when logging out in dev mode
- remove lots of redundant filtering from whenever discovery component renders


mcannalte:
Lots of these changes were made with upcoming features in mind. These upcoming features should require less work now:
- Migrating to server-side search, pagination 
- "saving" searches (which will be easy now that discovery page state can be serialized to a url)

src/Discovery/Discovery.css
```diff
@@ -204,6 +204,10 @@
   color: var(--g3-color__titanium);
 }
 
+#discovery-table-of-records {
```
mcannalte, line 207: Refactoring pagination out of the table results in a vertical scroll bar in the table. This is the only way I've found to remove that.

src/Discovery/DiscoveryActionBar.tsx
```diff
@@ -387,14 +391,15 @@ const DiscoveryActionBar = (props: Props) => {
       fetchWithCreds({ path: `${jobAPIPath}list` }).then(
         (jobsListResponse) => {
           const { status } = jobsListResponse;
-          if (status === 200) {
+          // jobsListResponse will be boilerplate HTML when not logged in
+          if (status === 200 && typeof jobsListResponse.data === 'object') {
```
mcannalte, line 395: Fixes "uncaught typeError in promise... o.default is undefined" 
see https://healdata.org while not logged in for example of incorrectly parsing a 200 response

mcannalte:
deployed in qa-heal.planx-pla.net

src/Discovery/reduxer.js
```diff
@@ -2,8 +2,10 @@ import { connect } from 'react-redux';
 import DiscoveryActionBar from './DiscoveryActionBar';
 
 const ReduxDiscoveryActionBar = (() => {
-  const mapStateToProps = (state) => ({
+  const mapStateToProps = (state, dispatch) => ({
     user: state.user,
+    discovery: state.discovery,
+    dispatch,
```
mfshao, line 8: this could work but a more standard way is to use the `mapDispatchToProps` for `connect` and pass in those dispatchable actions as props, see https://react-redux.js.org/using-react-redux/connect-mapdispatch

mfshao (commented):
super cool feature! I do have some comments

src/Discovery/reducers.js
```diff
@@ -0,0 +1,83 @@
...
+    };
+  case 'ACCESS_SORT_DIRECTION_SET':
+    return {
+      ...state,
+      accessSortDirection: action.accessSortDirection,
+    };
+  case 'REDIRECTED_FOR_ACTION':
+    return {
+      ...state,
+      ...action.redirectState,
```
mfshao, line 71: just want to confirm that I don't see a line of code that is dispatching an action to reset this state value, which means if a user logs in, select some studies, logs out and relog in with another username, the state of table will still be presisted and that is an intended behavior?

mfshao (approved):
apart from that single comment, this PR LGTM

awesome impmentation 💯

src/Discovery/reducers.js
mcannalte: When I tried this out, it looks like logging out will reset the redux store completely and the filters / selections are gone. Same thing when I log back in as a different user. 
Do we persist the redux store differently in production?

mfshao: nope, I think you are right

thanks for checking 👍

PR #944, opened 2021-11-03
updating base docker image for lets encrypt certs

MichaelLukowski:
### New Features
adding some ssl libraries to support new lets encrypt certificates


m0nhawk (approved):
We need to make sure in the future that those packages and ca-certificates are installed.

PR #959, opened 2022-01-11
(COV-1060): Remove commented code for Global/World tabs

nss10:
Jira Ticket: [COV-1060](https://occ-data.atlassian.net/browse/COV-1060)

### Improvements
- Removed commented code related to Global/World tabs


ocshawn (approved):
looks good

paulineribeyre (approved):
looks good, i think we should also remove config such as [this](https://github.com/uc-cdis/cdis-manifest/blob/9983fa9/chicagoland.pandemicresponsecommons.org/portal/gitops.json#L1154) and we can disable the part of the `covid19-notebook-etl` job that generates `top10.txt`

PR #964, opened 2022-01-24
chore(deps): upgrade dependencies

Avantol13:
### Dependency updates

- Upgrade tar to version 6.1.9, 5.0.10, 4.4.18 or higher.
- Upgrade prismjs to version 1.21.0 or higher.
- Upgrade trim-newlines to version 3.0.1, 4.0.1 or higher.
- Upgrade immer to version 9.0.6 or higher.
- Upgrade object-path to version 0.11.8 or higher.
- Upgrade zrender to version 5.2.1 or higher.
- Upgrade ansi-regex to version 6.0.1, 5.0.1 or higher.
- Upgrade css-what to version 5.0.1 or higher.
- Upgrade nth-check to version 2.0.1 or higher.
- Upgrade path-parse to version 1.0.7 or higher.
- Upgrade tmpl to version 1.0.5 or higher.
- Upgrade xss to version 1.0.10 or higher.


package.json
```diff
@@ -43,6 +45,7 @@
     "history": "^4.7.2",
     "html-react-parser": "^0.13.0",
     "ignore-loader": "^0.1.2",
+    "immer": "^9.0.6",
```
mfshao, line 48: I think something like this won't get rid of the old package if they are a dependency imported indirectly by another package right?

We probably need to look at where are these old packages get imported and fix from there

Avantol13: yeah I'm not as familiar with NPM. If you specify a requirement in pip/poetry of a sub-dependency, it will enforce what you specify across the board. But I think you're right... b/c I see multiple instances of inmer in the lock file now and some are still out of date :facepalm:

m0nhawk: NPM is completely different in the dependency management, I haven't tried that, but I think [`overrides`](https://docs.npmjs.com/cli/v8/configuring-npm/package-json#overrides) section will fix it:

```
  "overrides": {
    "foo": "^9.0.6"
  }
```

mfshao: yeah either override those or I can later try if I can upgrade them from the dependency that imports those old packages

is the list in the PR description comperhensive for what we want to update to?

mfshao:
closing this, has been addressed by https://github.com/uc-cdis/data-portal/pull/968

PR #969, opened 2022-01-31
(COV-1095): Feat/prc homepage chicago test

ocshawn:
Jira Ticket: [COV-1095](https://occ-data.atlassian.net/browse/COV-1095)

### Improvements

- The new default view is "Chicagoland", which is zipcode-based. The code can still display county-based data. Chicago neighborhoods boundaries are displayed by default. The "Illinois" view has been removed.

- The zoom levels are limited (both min and max) to range from neighborhood to state level.

- All of the definitions for both a layer and its associated legend are in the layer description (rather than being split between the main ChicagoMapChart/index.jsx and the overlay layer index.jsx). This allows the setMapLegendColors() function to be fully generalized, for instance. The specific information displayed as hover text is still defined in ChicagoMapChart/index.jsx as it is tied to source data rather than specific layers and data files are fetched in ChicagoMapChart/index.jsx. This could be moved to a function in ReduxCovid19Dashboard.js if desired. (The JHU data was loaded that way; the mobility data was fetched by the IllinoisMapChart/index.jsx).

- There are a total of 8 layers derived from three data files intended to demonstrate a range of possibilities. All of these are zipcode-based:

   - Vaccinations (fetched at runtime)

   - Cases and Testing (fetched at runtime)

   - Demographics (loaded from a static local file)

- The file src/params.js is not part of the co-dev model but needs to be tweaked to make the carousel appear. Specifically the "illinois" section has to be renamed to "chicago". It may be that this file is derived dynamically from the gitops.json file in the manifest directory; I don't really know.


mfshao:
don't remove the lock! it is needed by `npm ci` so we don't have unperdicted depnedency discrepancy between dev and prod

ocshawn:
> don't remove the lock! it is needed by `npm ci` so we don't have unperdicted depnedency discrepancy between dev and prod

yep outside developer removed it, i placed it back

ocshawn:
currently being backlogged, due to amount of effort

PR #974, opened 2022-02-14
PRC homepage: date slider style improvements

urvi-occ:
~Code is added to add a scroll-down effect to map div when the PRC homepage is loaded. 
MutationObserver is added to the `covid19-dashboard_panel` and it will observe attributes and nodes added to it. Once  `map-slider` has been added as a child to this div, scroll method will be applied~

~Reference for MutationObserver : https://developer.mozilla.org/en-US/docs/Web/API/MutationObserver~

### Improvements
- PRC homepage: date slider style improvements

### DIFF
--- a/src/Covid19Dashboard/MapSlider/mapSlider.less
+++ b/src/Covid19Dashboard/MapSlider/mapSlider.less
@@ -14,7 +14,7 @@
     color: #000;
     background: rgba(255,255,255,0.85);
     box-shadow: 0 2px 4px rgba(0,0,0,0.3);
-    margin: 10px;
+    padding: 10px;
     visibility: visible;
 }
 .map-slider input {
--- a/src/Covid19Dashboard/MapSlider/mapSlider.less
+++ b/src/Covid19Dashboard/MapSlider/mapSlider.less
@@ -5,7 +5,7 @@
     width: 50%;
     min-width: 15rem;
     max-width: 20rem;
-    padding: 1rem;
+    margin: 10px;
     float: right;
     z-index: 3;
 }
@@ -14,7 +14,7 @@
     color: #000;
     background: rgba(255,255,255,0.85);
     box-shadow: 0 2px 4px rgba(0,0,0,0.3);
-    padding: 0.5rem;
+    margin: 10px;
     visibility: visible;
 }
 .map-slider input {
--- a/src/Covid19Dashboard/MapSlider/mapSlider.less
+++ b/src/Covid19Dashboard/MapSlider/mapSlider.less
@@ -14,7 +14,6 @@
     color: #000;
     background: rgba(255,255,255,0.85);
     box-shadow: 0 2px 4px rgba(0,0,0,0.3);
-    border-radius: 3px;
     padding: 0.5rem;
     visibility: visible;
 }
--- a/src/Covid19Dashboard/MapSlider/mapSlider.less
+++ b/src/Covid19Dashboard/MapSlider/mapSlider.less
@@ -3,20 +3,19 @@
     font: 12px/20px 'Helvetica Neue', Arial, Helvetica, sans-serif;
     position: relative;
     width: 50%;
-    min-width: 150px;
-    max-width: 500px;
-    bottom: 120px;
-    padding: 10px;
+    min-width: 15rem;
+    max-width: 20rem;
+    padding: 1rem;
     float: right;
     z-index: 3;
 }
 .map-slider .map-overlay-inner {
     /* background-color: rgba(77, 0, 102, 0.7); */
-    background-color: black;
-    box-shadow: 0 1px 2px rgba(0, 0, 0, 0.2);
+    color: #000;
+    background: rgba(255,255,255,0.85);
+    box-shadow: 0 2px 4px rgba(0,0,0,0.3);
     border-radius: 3px;
-    padding: 10px;
-    margin-bottom: 10px;
+    padding: 0.5rem;
     visibility: visible;
 }
 .map-slider input {
@@ -26,11 +25,4 @@
     position: relative;
     margin: 0;
     cursor: ew-resize;
-}
-.map-slider span {
-    font-size: 24px;
-    line-height: 24px;
-    display: block;
-    margin: 0 0 10px;
-    color: white;
 }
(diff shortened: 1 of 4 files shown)


src/Covid19Dashboard/index.jsx
```diff
@@ -43,6 +43,42 @@ class Covid19Dashboard extends React.Component {
...
     );
+
+    // scrolling effect added to show homepage starting from covid counts
+    const targetNode = document.getElementsByClassName('covid19-dashboard_panel')[0];
+
+    // defined a scroll flag to only scroll for one time
+    let scrollFlag = false;
+
+    // Options for the observer (which mutations to observe)
+    const config = { attributes: true, childList: true, subtree: true };
```
mfshao, line 54: does this observer needs to observe for attribute changes? looks like it is only looking for adding a node

mfshao (commented):
made a small comment, but I'll defer to @paulineribeyre or @ocshawn for deploying for testing and the final approval since this is a PRC specific feature

src/Covid19Dashboard/index.jsx
urvi-occ: So according to [example](https://developer.mozilla.org/en-US/docs/Web/API/MutationObserver#example) here, We can use MutationObserver to observe changes in element either it changes the attribute or add a child list or a subtree. I have added condition to scroll the PRC homepage to show the Case Count div on top when map-slider div (which corresponds to the date slider at home page in PRC) is added in the HTML Document.

mfshao: right, this is what I meant, if you don't care about attribute changes, why seeting `attributes: true` in the observer config?

urvi-occ: Oh, that makes sense. Let me see if changing that works or not. Thank you so much for the note.

urvi-occ: Oh, that makes sense. I have changed setting `attributes: true` to `attributes: false` and made the commit

paulineribeyre:
I am trying to deploy this to QA so we can see it in action, but the service doesn't start. When I try to run it locally, it waits for user input, I suspect that's what is failing when we deploy it:
```
INFO: Ready for webpack
npx webpack serve
The command moved into a separate package: @webpack-cli/serve
Would you like to install serve? (That will run npm install -D @webpack-cli/serve) (yes/NO) : 
```
this doesn't seem related to the PR, @mfshao any ideas?

mfshao:
> 

the production bundle is build with `npx webpack build`, I think it is taking too long to build this bundle and k8s just restarts the pod

as for your local dev, that pkg should be placed in the loack already, have you tried `npm i` first?

mfshao:
oh... the prod pod actually failed with 

>     Last State:     Terminated
>       Reason:       OOMKilled

ocshawn:
Why is this functionality needed? is there a jira you can link to? 
There has to be a more elegant  way to implement. Perhaps implementing in the IllinoisMapChart/index.jsx will eliminate the need for a observer
Side note: `.scrollIntoView();` can also work in this situation

src/Covid19Dashboard/index.jsx
```diff
@@ -43,6 +43,42 @@ class Covid19Dashboard extends React.Component {
...
+
+    /**
+     * An observer instance linked to the callback function
+     * reference for MutationObserver: https://developer.mozilla.org/en-US/docs/Web/API/MutationObserver
+     */
+
+    const observer = new MutationObserver(callback);
+
+    // Start observing the target node for configured mutations
+    observer.observe(targetNode, config);
```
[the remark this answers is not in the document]
urvi-occ: Let me add and test that

urvi-occ:
@paulineribeyre @mfshao @ocshawn I have implemented the solution from the suggestions above on my local with changes as following
1. Reduced the calls to MutationObserver by making scroll action earlier in the page load process than it was already. I.e. it used to make around 5 calls to observer and now it is reduced to 2.
2. Removed `window.scrollTo` method and replaced it with `scrollIntoView`

The reason why I put this code in `src/Covid19Dashboard/index.jsx` file is because I wanted this code to execute when  `componentDidMount` method is executed. I tried finding a place in `IllinoisMapChart/index.jsx` but the `src/Covid19Dashboard/index.jsx` file seemed appropriate according to how both files are loading content. I would appreciate some more insight on finding the right location.

The reason why I introduced observer is that I wanted to trigger this event after the page is loaded. I tried following other methods in react life cycle like putting the scroll method directly but it triggered even before the page was loaded. I tried setting the interval method but the interval time might vary according to page load so I ended up doing an observer. The idea was to only have a scrolling effect after the page load.

I haven't made a new commit with the above changes yet. Let me know how should we proceed with this.

urvi-occ:
@paulineribeyre My first approach was to have the date slider on the top right on the map. I tried it on my local but because the map slider's current design it was covering the top part of Illinois counties. One solution I can think of is, we can decrease the opacity of the black color to overcome this but still it can be discussed further.

paulineribeyre:
@urvi-occ right, as mentioned i think it could (and should) be smaller. Also closer to the edge of the map, like the legend panel on the left

Screen Shot 2022-02-23 at 1 41 46 PM

urvi-occ:
Thank you for the screenshot it helps. Let me work on it and get back to you

ocshawn:
that looks good to me

urvi-occ:
@ocshawn I made the new commit in this same PR.

src/Covid19Dashboard/MapSlider/mapSlider.less
```diff
@@ -3,20 +3,19 @@
...
     z-index: 3;
 }
 .map-slider .map-overlay-inner {
     /* background-color: rgba(77, 0, 102, 0.7); */
-    background-color: black;
-    box-shadow: 0 1px 2px rgba(0, 0, 0, 0.2);
+    color: #000;
+    background: rgba(255,255,255,0.85);
+    box-shadow: 0 2px 4px rgba(0,0,0,0.3);
     border-radius: 3px;
```
paulineribeyre, line 17: let's remove the border-radius to match the other components

paulineribeyre (commented):
thanks @urvi-occ!

paulineribeyre:
@urvi-occ thanks for making the changes, it is deployed in our QA and i am waiting for PMs to have a look and give the green light.

paulineribeyre (approved):

PR #975, opened 2022-02-16
feat(disco-bar): floating discovery action bar, pointer cursor

mcannalte:
Jira Ticket: https://ctds-planx.atlassian.net/browse/HP-696, https://ctds-planx.atlassian.net/browse/HP-695

### New Features
- Floating action bar on discovery page with download / export buttons

### Improvements
- cursor pointer on mouse over study row
-  tool tip for download button


mfshao (commented):
overall LGTM and very cool work 💯 , will see if I can get quay to cooprate

src/Discovery/DiscoveryActionBar.tsx
```diff
@@ -455,164 +479,214 @@ const DiscoveryActionBar = (props: Props) => {
...
+        </Space>
+      </div>
+      {
+        floating && props.config.features.exportToWorkspace?.enabled
+        && (
+          <div className='discovery-floating-action-bar'>
+            <Card style={{ borderRadius: '5px' }}>
+              <Row gutter={[8, 0]} align="middle" justify="space-around">
+                <Col>
+                  <SelectOutlined style={{fontSize: "125%"}}/>
```
mfshao, line 669: use single quotes for strings for consistency?

src/Discovery/Discovery.css
```diff
@@ -343,6 +344,12 @@
   border-color: var(--g3-color__base-blue);
 }
 
+.discovery-action-bar-button:hover {
+  color: var(--secondary-color_light);
+  border-color: var(--secondary-color_light);
```
mcannalte, line 349: This isn't ideal. I'd prefer #fff, but then the text becomes invisible on BRH because that's the primary color

mfshao: I don't think BRH has a primary color, the buttons in BRH has the style that defined within their component library

I can see the point, but since HEAL's color scheme is highly customized, I would suggest to put this entrie `.discovery-action-bar-button:hover` section out into `gitops.css` rather than put them in here are default styles

mfshao: or we can set `#fff` as default color for `color` and `boarder-color`, and `--g3-color__base-blue` for `background-color`

then overrider the `background-color` in HEAL's gitops.css

PR #980, opened 2022-03-17
PXP-9656 Link from Explorer to DICOM viewer

paulineribeyre:
Jira Ticket: [PXP-9656](https://ctds-planx.atlassian.net/browse/PXP-9656)

### New Features
- Add an `explorerConfig.table.dicomViewerId` setting to link from the Explorer to the DICOM viewer


src/GuppyDataExplorer/ExplorerTable/index.jsx
```diff
@@ -7,6 +7,7 @@ import { Switch } from 'antd';
 import 'react-table/react-table.css';
 import IconicLink from '../../components/buttons/IconicLink';
 import { GuppyConfigType, TableConfigType } from '../configTypeDef';
+import { dicomViewerPath } from '../../localconf';
```
mfshao, line 10: if this will only be used by data explorer, we can put this value into explorer config as well rather than importing from localconf to simplifies code

src/GuppyDataExplorer/ExplorerTable/index.jsx
```diff
@@ -126,6 +127,17 @@ class ExplorerTable extends React.Component {
             return rowComp;
           }
         }
+
+        // if this field is the `dicomViewerId`, convert the value to a link to the DICOM viewer
+        if (this.props.tableConfig.dicomViewerId && this.props.tableConfig.dicomViewerId === field && valueStr) {
```
mfshao, line 132: make sense to put this section into the `// handling some special field types` section down below?

paulineribeyre: i tried it and it works too but doesn't look right IMO. I chose to separate it so the code would be less confusing and easier to maintain. if we move this to the `switch` statement:
- the order of the terms in the "if" cannot change (because of how "switch" compares the value of `field`, for example `this.props.tableConfig.dicomViewerId &&` can't be first)
- the order of the switch cases cannot change (because this new logic relies on the "linkFields" case to run afterwards to transform the value into a link)

paulineribeyre: i can change it if you think it's cleaner

mfshao: oh good point, don't worry about this then

src/GuppyDataExplorer/ExplorerTable/index.jsx
paulineribeyre: oh ok, sure i can do that

src/localconf.js
```diff
@@ -61,6 +62,7 @@ function buildConfig(opts) {
     workspaceURL,
     manifestServiceURL,
     requestorURL,
+    dicomViewerURL,
```
mfshao, line 65: remove these if they are not being used?

paulineribeyre: 🤦‍♀️

paulineribeyre:
Only 1 unrelated test failed (study viewer) and Jenkins keeps `aborting build due to out of date git hash` so i am merging

PR #982, opened 2022-03-30
Fix/dep

mfshao:
### Bug Fixes
- Add missing `rel='noreferrer'` for `target='_blank'` link

### Dependency updates
- Update gen3-ui-component to 0.11.1 (for stylelint 14)
- Update guppy to 0.15.1 (for stylelint 14)
- Upgrade stylelint to version 14.0.0 or higher.


MaribelleHGomez (dismissed):
Should we change version in package in package.json?

mfshao:
> Should we change version?

yes, the ui-component PR will go in first and I'll update those imports in a sequence

thanh-nguyen-dang (approved):

package.json
```diff
@@ -14,8 +14,8 @@
     "@fortawesome/free-brands-svg-icons": "^5.14.0",
     "@fortawesome/free-solid-svg-icons": "^5.2.0",
     "@fortawesome/react-fontawesome": "^0.1.0",
-    "@gen3/guppy": "^0.15.0",
-    "@gen3/ui-component": "^0.11.0",
+    "@gen3/guppy": "uc-cdis/guppy#fix/dep",
+    "@gen3/ui-component": "uc-cdis/gen3-ui-component#fix/dep",
```
thanh-nguyen-dang, line 18: I think we should release the new version instead of using the one in feature branch

thanh-nguyen-dang (changes requested):

package.json
```diff
@@ -39,7 +39,7 @@
     "file-saver": "^1.3.3",
     "focus-trap": "^6.5.0",
     "fuse.js": "^3.3.0",
-    "graphiql": "~1.5.0",
+    "graphiql": "1.5.18",
```
mfshao, line 42: have to pin to this version for now since the next version `1.5.19` of `graphiql` bumps up one of its dependency to `graphql-language-service: 5.0.0+`, which breaks how the `graphql-language-service` get loaded in from webpack

need some time to investigate how to properly upgrade them, in a later PR

Avantol13 (approved):

PR #986, opened 2022-04-19
Feat/wts login gwas

kmhernan:
Add in WTS login flow to `useEffect` that runs only on first mount of component (alongside the getting sources). In the future we could set a state for `connectedWts` and poll the connected endpoint to automatically login again, but for now this seems fine.


alilahrime (approved):

kmhernan:
@alilahrime cherry picked and integrated into branch already

PR #990, opened 2022-04-29
Fix/basepath

mfshao:
Jira Ticket: [HP-741](https://ctds-planx.atlassian.net/browse/HP-741)

Make portal works with customized `BASENAME`, in both production mode and dev mode

### New Features
- Make portal works with customized `BASENAME`, in both production mode and dev mode

### Improvements
- webpackdev-server works now with proper websocket connections, and fixed page hot-reload


src/UserAgreement/ReduxCertPopup.js
```diff
@@ -23,9 +23,10 @@ export const submitForm = (data, questionList, quiz) => (dispatch) => fetchWithC
 })
   .then(
     ({ status }) => {
+      const cleanBasename = basename.replace(/^\/+/g, '').replace(/(dev.html$)/, '');
```
mcannalte, line 26: Should `cleanBasename` ever change? 
Sometimes I see 

`const cleanBasename = basename.replace(/(dev.html$)/, '');`

And sometimes I see 

`const cleanBasename = basename.replace(/^\/+/g, '').replace(/(dev.html$)/, '');`

I think it would be cleaner to always use the second one and export it from `localconf`

mfshao: Yeah I agree this is confusing 😕 

So the differences is `basename.replace(/^\/+/g, '').replace(/(dev.html$)/, '')` actually takes out the leading `/` in the basename, if any. This is being used when the base name will be appending to somewhere that will have a trailing `/` later, for example, the `hostname` in portal. While the first one keeps the leading `/` in there.

we maybe should do an overhaul over all those names and paths in portal so they are more consistent, but I think we can create anotehr ticket for that task

mfshao: Or maybe we can always export the clean base name without the leading `/` from localconf, and add a leading `/` to it when needed?

mcannalte: That makes sense. Exporting the clean base name without a `/` sounds slightly cleaner, but it may not be worth it so I think this is fine

mcannalte:
I'm having some trouble running these changes locally, could you add an example usage to the README?
Not quite sure if:
- `BASENAME=portal` or `BASENAME=/portal`
- Should open `{HOSTNAME}/dev.html/portal` or `{HOSTNAME}/portal/dev.html`
 
It works okay when pointing to the QA instance:
`HOSTNAME=qa-heal.planx-pla.net/portal NODE_ENV=auto bash ./runWebpack.sh `

However this does not work:
`BASENAME=/portal HOSTNAME=qa-heal.planx-pla.net NODE_ENV=auto bash ./runWebpack.sh`

I get a blank page at `healdata.org/dev.html/portal/discovery` and 404 responses for `bundle.js` and `themeoverride.css`

mfshao:
> 

Good catch! I have update the document, hopefully that makes things more clear. The currect URL to access the dev server with a customized basename should be `{HOSTNAME}/portal/dev.html`

And also I think you have located an authentical bug in there when the access to `/dev.html` got a redirection after login. I have pushed a fix for that

mfshao:
> Thanks for the documentation. I tried pulling the latest code and running like this: `HOSTNAME=preprod.healdata.org NODE_ENV=auto BASENAME=/portal bash ./runWebpack.sh`
> 
> but then navigating to `https://preprod.healdata.org/portal/dev.html` in my browser I get a 404 page. Could I be missing something?

I think that is expected, when you hit `{HOSTNAME}/dev.html` it is actually pulling down the `/dev.html` and scripts from remote
Since the remote portal container in preprod is not running this branch with updated `dev.html` file, it won't work 😞 
But if you test it against qa-heal it should work

mcannalte:
> > Thanks for the documentation. I tried pulling the latest code and running like this: `HOSTNAME=preprod.healdata.org NODE_ENV=auto BASENAME=/portal bash ./runWebpack.sh`
> > but then navigating to `https://preprod.healdata.org/portal/dev.html` in my browser I get a 404 page. Could I be missing something?
> 
> I think that is expected, when you hit `{HOSTNAME}/dev.html` it is actually pulling down the `/dev.html` and scripts from remote Since the remote portal container in preprod is not running this branch with updated `dev.html` file, it won't work disappointed But if you test it against qa-heal it should work

qa-heal does work. It's a bit confusing because it doesn't work with `BASENAME` set to nothing (or something other than `/portal` so I'm guessing whenever we want to change this value, we have to first set the `BASENAME` of the deployment, then match that change locally. I think this is okay though since changing the `BASENAME` should be very rare

mcannalte (dismissed):

mfshao:
> 

I think this is a nuance worth mentioning in the doc, can you take a look again?

mcannalte:
Thanks for clarifying in the doc, LGTM!

mcannalte (approved):

PR #996, opened 2022-05-25
(PXP-9838): feat/studyViewerExternalDownload: add flag to specify field for external file download

ocshawn:
Jira Ticket: [PXP-9838](https://ctds-planx.atlassian.net/browse/PXP-9838)

### New Features
- add flag to specify field for external file download

### DIFF
--- a/src/StudyViewer/StudyDetails.jsx
+++ b/src/StudyViewer/StudyDetails.jsx
@@ -138,17 +138,35 @@ class StudyDetails extends React.Component {
       && this.isDataAccessible(this.props.data.accessibleValidationValue)
       && this.props.fileData.length > 0;
 
-      button = displayDownloadButton ? (
-        <Button
-          key={key}
-          label={'Download'}
-          buttonType='primary'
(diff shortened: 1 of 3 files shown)


src/StudyViewer/StudyDetails.jsx
```diff
@@ -138,17 +138,35 @@ class StudyDetails extends React.Component {
...
-        />
-      ) : null;
+      if (this.props.data.overrideDownloadUrlField && displayDownloadButton) {
+        button = (
+          <Button
+            key={key}
+            label={`Download ${this.props.data.requiredIdpField
+              ? `on ${this.props.data.requiredIdpField.toUpperCase()}`
+              : 'Externally'
+            }`}
```
paulineribeyre, line 148: Even if this is how the NIAID use case works, i don't think we can assume that the download URL is always going to the "required IDP" website, or that it's always external to Gen3. Maybe the button text can still be "Download"? but the `rightIcon` is a nice touch

paulineribeyre (approved):

PR #998, opened 2022-05-26
Fix/home link with ff

haraprasadj:
[HP-768](https://ctds-planx.atlassian.net/browse/HP-768)

### Improvements
- Make `navTitle` supports customized `homepageHref`


mfshao (approved):
thanks 👍

PR #1005, opened 2022-06-09
Fix/case control hare

kmhernan:
* Updated hare concept ID
* Fixed boolean logic that always would return error message in hare selection dropdown area
* passed the `useQuery` options to the 2 `useQueries` to turn off unwanted event responses (e.g., refresh on focus, etc.)


pieterlukasse:
Thanks @kmhernan !

PR #1024, opened 2022-06-22
fix(hare): pass concept value name

kmhernan:
The short names (e.g., `ASN`) were still being passed to argo wrapper which causes a problem since the CSV will write out the long name. Added even more state vars for the other string. We really should have only 1 state for that whole hare object tbh.


pieterlukasse:
thanks @kmhernan , I agree w/ your comment on having the object as the state instead of the individual values it holds.

PR #1029, opened 2022-07-05
Include dichotomous variables in requests

pieterlukasse:
...in requests to cohort-middleware and in final submit call to argo-wrapper

Jira Ticket: [VADC-2](https://ctds-planx.atlassian.net/browse/VADC-2), related to changes in https://github.com/uc-cdis/cohort-middleware/pull/44

### Breaking Changes

- include dichotomous variables in requests to cohort-middleware and in final submit call to argo-wrapper, using the new "variables" format

### DIFF
--- a/src/Analysis/GWASUIApp/CaseControlGWAS.jsx
+++ b/src/Analysis/GWASUIApp/CaseControlGWAS.jsx
@@ -66,6 +66,8 @@ const CaseControlGWAS = (props) => {
     const [form] = Form.useForm();
 
     const [selectedCovariates, setSelectedCovariates] = useState([]);
+    // TODO - remove dummy value - assigning a static value here until the UI component is ready to populate selectedDichotomousCovariates:
+    const [selectedDichotomousCovariates, setSelectedDichotomousCovariates] = useState([{'name': 'dummy dichotomous variable', 'cohort1': 400, 'cohort2': 401}]);
     const [selectedCovariateIds, setSelectedCovariateIds] = useState([]);
     const [selectedCovariateVars, setSelectedCovariateVars] = useState([]);
 
@@ -395,14 +397,12 @@ const CaseControlGWAS = (props) => {
     };
 
     async function fetchConceptStatsByHare(cohortDefinitionId) {
-        var conceptIds = [...selectedCovariates].map((val) => val.concept_id);
-        const conceptIdsPayload = { ConceptIds: conceptIds };
         const conceptStatsEndPoint = `${cohortMiddlewarePath}concept-stats/by-source-id/${sourceId}/by-cohort-definition-id/${cohortDefinitionId}/breakdown-by-concept-id/${hareConceptId}`;
         const reqBody = {
             method: 'POST',
             credentials: 'include',
             headers,
-            body: JSON.stringify(conceptIdsPayload),
+            body: JSON.stringify(getVariablesList(selectedDichotomousCovariates, selectedCovariates)),
         };
         const getConceptStats = await fetch(conceptStatsEndPoint, reqBody);
         return getConceptStats.json();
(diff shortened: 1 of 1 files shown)


src/Analysis/GWASUIApp/CaseControlGWAS.jsx
```diff
@@ -511,15 +511,28 @@ const CaseControlGWAS = (props) => {
         }
     };
 
+    // This function takes in list of dichotomous covariates (cohort-pair based) and other regular covariates (concept-based)
+    // and produces the generic "variables" list that can be used as input argument in different cohort-middleware endpoints.
+    function getVariablesList(dichotomousCovariates, otherCovariates) {
+        const variables = [];
+        for (let dichotomousCovariate of dichotomousCovariates) {
+            const variableItem = {'variable_type': 'custom_dichotomous', 'cohort_ids': [dichotomousCovariate.cohort1, dichotomousCovariate.cohort2] };
```
kmhernan, line 519: Will this maintain the order as selected by the user? Will we want to also have the `provided_name` or something here?

src/Analysis/GWASUIApp/CaseControlGWAS.jsx
```diff
@@ -511,15 +511,28 @@ const CaseControlGWAS = (props) => {
...
     async function fetchOverlapInfo() {
-        var conceptIds = [...selectedCovariates].map((val) => val.concept_id);
-        const conceptIdsPayload = { ConceptIds: conceptIds };
         const statsEndPoint = `${cohortMiddlewarePath}cohort-stats/check-overlap/by-source-id/${sourceId}/by-case-control-cohort-definition-ids/${caseCohortDefinitionId}/${controlCohortDefinitionId}/filter-by-concept-id-and-value/${hareConceptId}/${selectedHareValueAsConceptId}`;
         const reqBody = {
             method: 'POST',
             credentials: 'include',
             headers,
-            body: JSON.stringify(conceptIdsPayload),
+            body: JSON.stringify(getVariablesList(selectedDichotomousCovariates, selectedCovariates)),
```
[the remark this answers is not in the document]
kmhernan: ah this is case-control situation ignore

src/Analysis/GWASUIApp/CaseControlGWAS.jsx
```diff
@@ -567,7 +580,9 @@ const CaseControlGWAS = (props) => {
         const submitEndpoint = `${gwasWorkflowPath}submit`;
         const requestBody = {
             n_pcs: numOfPC,
+            // TODO - deprecated - remove this "covariates" item as selectedCovariates are already part of "variables" below:
             covariates: selectedCovariates.map((val) => val.prefixed_concept_id),
+            variables: getVariablesList(selectedDichotomousCovariates, selectedCovariates),
```
kmhernan, line 585: yeah I do like this approach I think... we could drop the `covariates` as we mentioned today

src/Analysis/GWASUIApp/CaseControlGWAS.jsx
alilahrime: having a `provided name` here as part of `variableItem` makes sense. if we need to preserve order then I'd suggest using javascript `map()` and then spread operator (ex. `variables = [...cdVariables, ...otherVariables]`) to put the lists together.

src/Analysis/GWASUIApp/CaseControlGWAS.jsx
alilahrime: yeah agreed. keeping state variables separate for these 2 and adding `getVariablesList()` to request is better since we modify (add/delete) those types of covs differently

src/Analysis/GWASUIApp/CaseControlGWAS.jsx
pieterlukasse: @kmhernan it is not preserving the order. I judged the order not to be important at this moment. But let me know if you think otherwise!

src/Analysis/GWASUIApp/CaseControlGWAS.jsx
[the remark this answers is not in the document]
pieterlukasse: Even for the quantitative scenario, it would probably be useful to have the outcome phenotype as part of the variables list (especially in the argo-wrapper `requestBody` below). Then you can just take this list as is and use it in the CSV and attrition table endpoints. The downside is that you need to remember to skip it whenever you need only the covariates in your own code.

src/Analysis/GWASUIApp/CaseControlGWAS.jsx
kmhernan: I guess it depends on how it is ultimately used @pieterlukasse we do care about order I think in attrition table, more important when the attrition UI component is there. Less important for the csv that is included in the tar archive. @alilahrime it would be more complicated than that in the more complicated variable UI where we could add a continuous followed by a dichotomous followed by a continuous etc. So you would have to have a single list. I'm also thinking about that new UI when there is the list of variables you selected we probably want to care about order. Maybe that is not important at all in this version though and I'm just thinking too far ahead.

src/Analysis/GWASUIApp/CaseControlGWAS.jsx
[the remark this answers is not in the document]
kmhernan: @pieterlukasse i am now working under the assumption the outcome phenotype is included in the variable list for continuous and the `outcome` field will be the concept ID of the outcome continuous phenotype. What i need us to standardize is the name of the property to hold the provided name of the custom dichotomous variable so I can include it in the metadata yaml.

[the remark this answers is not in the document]
pieterlukasse: we'll use `provided_name`

src/Analysis/GWASUIApp/CaseControlGWAS.jsx
pieterlukasse: @kmhernan thanks for the explanation. 

Here are my thoughts on this matter: as I see it now, the attrition table in the frontend will not use the CSV attrition table endpoint, for performance reasons. The way I'm thinking about the implementation right now is to have each new row call the _breakdown_ endpoint at the time it is added to the list/table in the UI. This is essentially what the CSV attrition table endpoint does when generating the final attrition table CSV in the backend... however, we don't want to generate all the rows again whenever just one extra row should be added in the frontend.

Long story short: I do not have a scenario yet in which we need to inform cohort-middleware about a specific order of the variables.

PR #1030, opened 2022-07-05
Add new fields in Discovery page manifest

nss10:
Added new fields in DiscoveryPortal manifest config.

### DIFF
--- a/src/Discovery/DiscoveryActionBar.tsx
+++ b/src/Discovery/DiscoveryActionBar.tsx
@@ -310,6 +310,9 @@ const handleDownloadManifestClick = (config: DiscoveryConfig, selectedResources:
           ...x,
           commons_url: ('commons_url' in x)
             ? x.commons_url : study.commons_url,
+            project_number: study.project_number,
+            study_name: study.study_name_title.trim().toLowerCase().replaceAll(' ', '_'),
+            repository_name: study.commons,
         })));
       } else {
         manifest.push(...study[manifestFieldName]);
@@ -363,6 +366,9 @@ const handleExportToWorkspaceClick = async (
           ...x,
           commons_url: ('commons_url' in x)
             ? x.commons_url : study.commons_url,
+            project_number: study.project_number,
+            study_name: study.study_name_title.trim().toLowerCase().replaceAll(' ', '_'),
+            repository_name: study.commons,
         })));
       } else {
         manifest.push(...study[manifestFieldName]);


src/Discovery/DiscoveryActionBar.tsx
```diff
@@ -310,6 +310,9 @@ const handleDownloadManifestClick = (config: DiscoveryConfig, selectedResources:
           ...x,
           commons_url: ('commons_url' in x)
             ? x.commons_url : study.commons_url,
+            project_number: study.project_number,
+            study_name: study.study_name_title.trim().toLowerCase().replaceAll(' ', '_'),
```
mfshao, line 314: not all metadata is expected to have a field called `study_name_title` to be used as `study_name`

this field should be configurable from the config file

mfshao (commented):
please fill out the PR template with proper section headings so changes can be included in the release notes

src/Discovery/DiscoveryActionBar.tsx
nss10: Thanks for this. Let me make it configurable. BTW, is there a common value we can fallback to? Or do we make this a mandatory field in the config?

aartivnkt: hey @mfshao @jawadqur any quick thoughts for how to configure that field from the config file? That could help Sai. I'll also start looking

nss10: I figured where / how to make that configurable, I just don't know what do we do about the `standard title`.

aartivnkt: I see in this config file there are different ways to define `full_name` https://github.com/uc-cdis/cdis-manifest/blob/master/brh.data-commons.org/metadata/aggregate_config.json. We could make a logic that includes all the different ways `full_name` is defined for each data_commons. Is that what @mfshao suggested, or something different?

aartivnkt: @nss10 The issue is that the metadata is not standardized for the different commons for us to have a `study_name` for each commons. I checked out metadata info for the individual commons by going to the following URLs. From what I see, we have potential matches for `study_name`. But MIDRC does not have a suitable match
```
https://gen3.biodatacatalyst.nhlbi.nih.gov/mds/metadata?data=True (study_name = study_subject_id)
https://ibdgc.datacommons.io/mds/metadata?data=True (study_name = dataset_title)
https://accessclinicaldata.niaid.nih.gov/mds/metadata?data=True (study_name = title)
https://data.midrc.org/mds/metadata?data=True (nothing for study_name)
https://jcoin.datacommons.io/mds/metadata?data=True (study_name = study_name_title)
https://gen3.datacommons.io/mds/metadata?data=True (study_name = study_title)
https://internalstaging.theanvil.io/mds/metadata?data=True (study_name = name)
```
Whereever possible we could use these to map but in many cases that `study_title` field is very long. We ideally need a short name for the study.

Perhaps for now, we can just name the study using the `commons_url` which is a little more descriptive than `exported-manifest-2022-06-27T19-42-38.355056` ? Eg: MIDRC one would be `exported-midrc-study` and inside that we can have sorted by date. Similar solution for all other commons. It's not ideal but it's a temporary solution for now.

Ultimately, we want to standardize fields for all commons, which requires standardization of metadata across all commons. That is a bigger issue to tackle for the future releases.

aartivnkt: Welcome any thoughts from others @jawadqur @mfshao that have looked at this more

ocshawn:
cleaning up old prs please reopen if this still needs to be addressed

PR #1037, opened 2022-07-27
feat read argo template version from config

UchicagoZchen138:
Jira Ticket: NA

### New Features

- read gwas template version from config


kmhernan (approved):
I think this works ;-)

PR #1064, opened 2022-08-22
Test/aggmds counts

craigrbarnes:
Jira Ticket: [BRH-169](https://ctds-planx.atlassian.net/browse/BRH-169)

### New Features
* Changes how the Discovery Page retrieves metadata from the aggregateMDS. Instead of pulling the ```__manifest``` field, it adds the count option to the aggregate MDS which returns the length of the array. If a study is selected the array contents of the ```__manifest``` field is pulled from the aggreateMDS.

### Bug Fixes
* if ```_subject_count``` is not a number, change the value to "N/A".


src/Discovery/Discovery.tsx
```diff
@@ -89,7 +89,7 @@ export const renderFieldContent = (content: any, contentType: 'string'|'paragrap
     if (Array.isArray(content)) {
       return content.join(', ');
     }
-    return content.toLocaleString();
+    return content ? content.toLocaleString() : 'N/A';
```
aartivnkt, line 92: nice I like this as this won't crash undef values

mfshao, line 92: Why not use valueIfNotAvailable to handle missing values? Just like what we were doing in existing codes

aartivnkt: Two questions: does a missing `content` value need to be type matched? eg: if an int is missing the portal code should be looking for 0 as opposed to `N/A`? also how about `__manifest`? If there is none, that should be [] as opposed to `N/A`?

ocshawn:
@craigrbarnes is this still needed?

craigrbarnes:
> @craigrbarnes is this still needed?

Yes. I'm looking at it now. It is needed for BRH

src/Discovery/Discovery.tsx
craigrbarnes: Good catch. I have updated the expression

craigrbarnes: That works, however, it requires the configuration to ensure all fields are using `valueIfNotAvailable`
But that is cleaner solution

ocshawn (approved):
looks good to me

PR #1067, opened 2022-08-25
Add Workspace Registration to Data-Portal

nss10:
Jira Ticket: [BRH-223](https://ctds-planx.atlassian.net/browse/BRH-223)

### New Features
* Add Workspace Registration feature where users can have an option to request for workspace access directly from dataportal

### Breaking Changes
* Replaced loading  the data in  `iframes` in Workspace page with Redirect component. 
* Deprecating the use of `default-workspaces` i. e an individual commons will no longer be able to deploy standalone jupyterhub images, they may use [hatchery service](https://github.com/uc-cdis/hatchery) instead. 
* `studyRegistrationConfig` in `gitops.json` must now be placed within the `regsitrationConfigs.features` block [[Refer this PR](https://github.com/uc-cdis/cdis-manifest/pull/5095)]
### Improvements
* Created a new `/403Error` page to redirect all users for a non-workspace related Unauthorized requests


src/Workspace/ErrorWorkspacePlaceholder.jsx
```diff
@@ -1,21 +1,33 @@
...
-        <h1>Error opening workspace...</h1>
-        <p>
-        Workspace access requires authorization. Please contact <a href={`mailto:${supportEmail}`}>{supportEmail}</a> for more information.
-        </p>
-        <NotFoundSVG />
-      </div>
-    );
+const ErrorWorkspacePlaceholder = () => {
+  const history = useHistory();
+  if (isEnabled('workspaceRegistration')) {
```
mfshao, line 11: I think you should wrap these stuff in a useEffect hook so it will be called only once when the component mounts. Otherwise it will be called each time this component renders, which is unnecessary

src/Workspace/ErrorWorkspacePlaceholder.jsx
```diff
@@ -1,21 +1,33 @@
...
+    if (workspaceRegistrationConfig) {
+      // eslint-disable-next-line react/prop-types
```
mfshao, line 13: Why disable this?

src/Workspace/ErrorWorkspacePlaceholder.jsx
```diff
@@ -1,21 +1,33 @@
...
+      history.push('/workspace/register');
+    } else {
+      // eslint-disable-next-line no-console
+      console.log('Unable to display registration form. Missing workspaceRegistrationConfig');
```
mfshao, line 17: I think this information doesn't needs a print out to the client side

src/Workspace/ErrorWorkspacePlaceholder.jsx
```diff
@@ -1,21 +1,33 @@
...
+    }
```
mfshao, line 18: I remember in the design doc, there is a page with button and some explanatory text to tell user why they cannot open a workspace and they need to request access for it. But you are re-directing user to the registration form directly in here. Has the design changed? 
Personally I think the current flow is a bit confusing to the user who got redirected directly to a form upon clicking workspace button

src/Workspace/index.jsx
```diff
@@ -736,17 +737,7 @@ class Workspace extends React.Component {
     } if (this.state.defaultWorkspace && this.state.connectedStatus) {
       // If this commons does not use Hatchery to spawn workspaces, then this
       // default workspace is shown.
```
mfshao, line 739: These comments are no longer valid

src/WorkspaceRegistration/WorkspaceRegistration.tsx
```diff
@@ -0,0 +1,299 @@
...
+      <div className='workspace-reg-container'>
+        <div className='workspace-reg-form-container'>
+          {(formSubmissionStatus.status === 'success') ? (
+            <Result
+              status={formSubmissionStatus.status}
+              title='Your access request has been submitted!'
+              subTitle='Thank you for your submission. Requests take up to 1 business day to complete. You will be notified of the status.'
+              extra={[
+                <Link key='discovery' to={'/discovery'}>
+                  <Button>Go To Discovery Page</Button>
```
mfshao, line 138: Does it needs a go to discovery page button? Would a go to homepage button be more meaningful?

src/WorkspaceRegistration/WorkspaceRegistration.tsx
```diff
@@ -0,0 +1,299 @@
...
+              ]}
+            />
+          ) : (
+            <Result
+              status={formSubmissionStatus.status}
+              title='A problem has occurred during submitting the request!'
+              subTitle={formSubmissionStatus.text}
+              extra={[
+                <Button type='primary' key='close' onClick={() => { setFormSubmissionStatus(null); setReqAccessRequestPending(false); }}>
+                  Close
```
mfshao, line 149: Does it needs a close button?

src/WorkspaceRegistration/WorkspaceRegistration.tsx
```diff
@@ -0,0 +1,299 @@
...
+  }
+
+  if (kayakoConfig) {
+    return (
+      <div className='workspace-reg-container'>
+        <div className='workspace-reg-form-container'>
+          <Form className='workspace-reg-form' {...layout} form={form} name='workspace-reg-request-form' onFinish={onFinish} validateMessages={validateMessages}>
+            <Divider plain>Workspace Registration Access Request</Divider>
+            <Typography style={{ textAlign: 'center' }}>
+              Please fill out this form to request and be approved for access to workspace with the BRH Platform.
```
mfshao, line 166: Don't hard code BRH in here. All hard coded texts should be generic. If you have to mention a specific env name, make it configurable

src/WorkspaceRegistration/WorkspaceRegistration.tsx
```diff
@@ -0,0 +1,299 @@
...
+      method: 'POST',
+      body: JSON.stringify(body),
+    }).then(
+      ({ data, status }) => {
+        if (status === 201) {
+          if (data && data.request_id) {
+            // request created, now create a kayako ticket
+            const fullName = `${formValues['First Name']} ${formValues['Last Name']}`;
+            const email = formValues['E-mail Address'];
+            const subject = 'Registration Access Request for Workspace';
```
mfshao, line 87: All request will have the same subject in Kayako?

src/components/ErrorPage403.jsx
```diff
@@ -0,0 +1,21 @@
+import React from 'react';
+import './NotFound.less';
+import NotFoundSVG from '../img/not-found.svg';
+import { components } from '../params';
+
+class ErrorPage403 extends React.Component {
+  render() {
+    const supportEmail = components.login?.email || 'support@datacommons.io';
+    return (
+      <div className='error-placeholder__error-msg_no_header_footer '>
```
mfshao, line 10: This class name doesn't exists in the stylesheets that this component has imported

src/localconf.js
```diff
@@ -298,6 +300,9 @@ function buildConfig(opts) {
   const { workspacePageTitle } = config;
   const { workspacePageDescription } = config;
 
+  const workspaceRegistrationConfig = resourceRegistrationConfig ? resourceRegistrationConfig.workspaceRegistrationConfig : null;
+  const kayakoConfig = resourceRegistrationConfig ? resourceRegistrationConfig.kayakoConfig : null;
```
mfshao, line 304: Probably need to think about how to better organize fields in config files so this config can be shared by both ws reg and study reg

src/Workspace/ErrorWorkspacePlaceholder.jsx
nss10: Oh! Is it? Thanks for letting me know. I can take it off, but I was wondering what would be the best way to let the developer know if this is what they're missing if a situation like this arises?

src/Workspace/ErrorWorkspacePlaceholder.jsx
nss10: Oh! That's a good catch. I think I understand what you mean, when you say it could be confusing for a new user when they suddenly look at this form. I agree with you. The idea in the document was either to welcome them with a paragraph and put a button, or do it along with a form to fill. So, the current design is still in line with the original design choice, but adding more context is what was missing.

src/WorkspaceRegistration/WorkspaceRegistration.tsx
nss10: I didn't see any harm in keeping the button there. It displays the error message and takes back the user to the form again, which seemed intuitive thing to do. Do you have any better suggestions?

src/WorkspaceRegistration/WorkspaceRegistration.tsx
nss10: That is a very good point. I always am an advocate against hardcoding, thank you for catching this.

src/components/ErrorPage403.jsx
nss10: Thanks. I missed this, I added a new class to make sure the css hack gets applied, but later when I changed the iframe to redirect, I eliminated the css class, but forgot to eliminate this implementation.

src/localconf.js
nss10: I would love to discuss about that with you. I was trying to bring a common config which forms the base for everything that is related to resource registration in general. Within which I added workspaceRegistration and KayakoConfig, Tried to add studyRegistrationConfig as well, but decided that would be like making too many changes in a single PR.

src/Workspace/ErrorWorkspacePlaceholder.jsx
mfshao: We usually just put these kind of print out in adhocly when doing development, and remove them before shipping to production

src/Workspace/ErrorWorkspacePlaceholder.jsx
mfshao: What I'm saying is I think implementation is not following details from its design doc. So update either of them or both to keep them in sync

src/WorkspaceRegistration/WorkspaceRegistration.tsx
mfshao: Apologies. I misread. Probably shouldn't review codes when 🍼

src/localconf.js
mfshao: I was thinking something like: 

{
 registrationConfigs: {
    features: [
       studyRegistration: {
         // study reg configs...
       },
       workspaceRegistration: {
         // workspace reg configs...
       },
       ... // other reg configs in the future
    ],
    kayakoConfig: {
      // Kayako configs...
    }
}

Of course this needs polish. But if we can do this, then we can put more breaking changes into one release, and bumping up the major version only once

src/Workspace/ErrorWorkspacePlaceholder.jsx
nss10: Hey Mingfei, after you mentioned this, and the using of useEffect() , I felt like we can use of Redirect here again instead of history.push, because, when I tried to do useEffect, I see the entire component first and then being redirected to the form. I will make a new push and let you know

src/localconf.js
nss10: Absolutely. Let me go ahead and test this config and make a new commit.

nss10: I did change the config based on your suggestion and ran some tests in qa-brh and qa-heal. The new changes are working really well. When I publish changes to gitops-qa and cdis-manifest with the new changes, I'll make sure I'll add heal's changes as well.

src/WorkspaceRegistration/WorkspaceRegistration.tsx
nss10: Thanks for pointing thing out. Ideally for BRH, we still need the redirect to be to the `Discovery Page`. But after you pointed that out, I've gone ahead and added this as a [configurable field](https://github.com/uc-cdis/gitops-qa/blob/6cb408012f1faf8e2d30ea6b305362d21359cda0/qa-brh.planx-pla.net/portal/gitops.json#L460-L462) in workspaceConfig

src/WorkspaceRegistration/WorkspaceRegistration.tsx
nss10: Good point, I reached out to user-services and they suggested that they could use the name of the commons in the title. I've added a new field called [commonsName](https://github.com/uc-cdis/data-portal/pull/1067/commits/eae35aece16e2ae74e6f3744d94f9f5440135612) in localconf.js

src/WorkspaceRegistration/WorkspaceRegistration.tsx
mfshao: This might be a bit over complicated, but for BRH and HEAL, their homepage is indeed the discovery page already, because they are using the ecosystem bundle

src/WorkspaceRegistration/WorkspaceRegistration.tsx
mfshao: I honestly think the `hostname` value is representitive enough. 🤷‍♂️  You just need to put it in the subject as well as in the body.

Besides, your logic for generating `commonsName` is not accurate. For example, https://gen3.biodatacatalyst.nhlbi.nih.gov/ will be GEN3, instead of BDCAT

So just use the plain `hostname` instead

src/localconf.js
mfshao: Please update the design doc as well as the description of this PR to reflect what changes has been made for portal config. For PR description, they should be either in `Beaking changes` or `Deployment changes`

Also update this file https://github.com/uc-cdis/data-portal/blob/master/docs/portal_config.md#the-portal-config-file
If you feel you need to write sore explanatory texts, you can also create a dedicated markdown file in here https://github.com/uc-cdis/data-portal/tree/master/docs

src/Workspace/index.jsx
```diff
@@ -734,19 +735,7 @@ class Workspace extends React.Component {
         </div>
       );
     } if (this.state.defaultWorkspace && this.state.connectedStatus) {
```
mfshao, line 737: there is no more `defaultWorkspace`, so you should change the name of this state variable to better represent the current logic, otherwise it is super confusing to ppl who reads this code

src/Workspace/index.jsx
```diff
@@ -734,19 +735,7 @@ class Workspace extends React.Component {
...
-          <iframe
-            title='Workspace'
-            frameBorder='0'
-            className='workspace__iframe'
-            src={workspaceUrl}
-            onLoad={this.oniframeLoad}
-          />
-        </div>
-      );
+      return <Redirect to={workspaceUrl} />;
```
mfshao, line 738: why redirect use to `/lw-workspace` in this case? Shouldn't just redirect user to the workspace error page?

src/Workspace/ErrorWorkspacePlaceholder.jsx
nss10: Since we have agreed to change the implementation to use `<Redirect>` instead of `useHistory()`

src/WorkspaceRegistration/WorkspaceRegistration.tsx
nss10: What do you suggest would be the best way to go? The advantage of making it configurable is, other commons which don't rely on `Discovery Page` can also directly include this feature.

src/WorkspaceRegistration/WorkspaceRegistration.tsx
nss10: Thank you. I was only thinking of BRH and QA-BRH when I was writing that logic. This example definitely helps. Will change it back to `hostname`.

src/Workspace/index.jsx
```diff
@@ -733,20 +736,12 @@ class Workspace extends React.Component {
...
-            onLoad={this.oniframeLoad}
-          />
-        </div>
-      );
+    } if (this.state.connectedStatus && !this.state.hasWorkspaceAccess) {
+      if (isEnabled('workspaceRegistration')) {
+        console.log('This is navigated to workspace register page');
+        return <Redirect to='/workspace/register' />;
+      }
+      return <Redirect to={workspaceErrorUrl} />;
```
mfshao, line 744: why not redirect to `Error403Url` in this case?

mfshao: Actually, if we can tune the `hasWorkspaceAccess` value a bit finer, maybe we can do something like this:

- If https://github.com/uc-cdis/data-portal/pull/1067/files#diff-f907a00fd9e7338bc39b77dc677d401f5dad62c284b04a4f1d94345854432f24R130 gives a 403 error AND workspace registration is enabled: redirect to ws reg page
- If https://github.com/uc-cdis/data-portal/pull/1067/files#diff-f907a00fd9e7338bc39b77dc677d401f5dad62c284b04a4f1d94345854432f24R130 gives a 403 error AND workspace registration is NOT enabled: redirect to error 403 page
- If https://github.com/uc-cdis/data-portal/pull/1067/files#diff-f907a00fd9e7338bc39b77dc677d401f5dad62c284b04a4f1d94345854432f24R130 gives a NON-403 error: redirect to the generic workspace error page

WDYT about this approach? 🤔

mfshao (commented):
generally lgtm

src/Workspace/index.jsx
nss10: That seems like a good way to go! Just to make sure I understand it right, 
we redirect the users to- 
1.  `/no-workspace-access` page for any `non-403` error,
2. `/workspace/register` page for a `403 error` with `workspaceRegistration` flag enabled, 
3. but if the response's status is `403` we then just update the `this.setState({ hasWorkspaceAccess: false });` and redirect users to the `/403Error` page [here](https://github.com/uc-cdis/data-portal/blob/e543a1ec4db83b3f1348a3a490e92b959c375c01/src/Workspace/index.jsx#L747-L747)

mfshao: yeah, although we should rename the path the generic workspace error page to something else, because in that case, it is not because of no workspace access

maybe the 403 error page should take the `/no-workspace-access` path and the generic workspace error page should be `/workspace-error`

mfshao: Update: nginx redirects the traffic to workspace error page now, so we won't see 403 from portal side

We can let nginx stop from doing redirection, but it is out of the scope of this ticket

nss10: Well, I still don't think the 403 page should be `/no-workspace-access`, for instance, I guess, thor has nothing to do with workspaces but I still see it being navigated to `/no-workspace-access` [here](https://github.com/uc-cdis/cloud-automation/blob/master/kube/services/revproxy/gen3.nginx.conf/thor-service.conf#L18-L18)

mfshao: yeah that is correct, we don't have a generic 403 error page at that time and everything was being redirected to ws error page, not good

PR #1054, opened 2022-08-04
GWAS UI v2

alilahrime:
### New Features
latest gwas ui post-refactor 

### Dependency updates
sync w/ master on removing babel-eslint

PR #1079, opened 2022-09-06
Fix content security policy typo

paulineribeyre:
Fix typo introduced in #1054 which is causing issues loading the content security policy and is preventing us from accessing clinicaltrials.gov:
> "The Content-Security-Policy directive name 'https://clinicaltrials.gov' contains one or more invalid characters. Only ASCII alphanumeric characters or dashes '-' are allowed in directive names."


ocshawn (approved):
looks good

PR #1086, opened 2022-09-12
Discovery page: fix "undefined tags" error

paulineribeyre:
### Bug Fixes
- Discovery page: fix "undefined tags" error when an entry does not have tags


george42-ctds (approved):
Nice!

paulineribeyre:
skipping `DrsAPI get drs presigned-url no auth header` which is broken at the moment https://cdis.slack.com/archives/C0294A7KVE0/p1662999554721989

PR #1095, opened 2022-09-19
Centralize WTS refresh token initialization

pieterlukasse:
Jira Ticket: [VADC-154](https://ctds-planx.atlassian.net/browse/VADC-154)

### Improvements
- WTS is a useful service for other backend components that need to access gen3 services using a token bound to a specific user account.

### Dependency updates
- WTS becomes a dependency for data-portal in general, instead of just for the /workspace subsection of the portal.
  - The change IS backwards compatible, since this is controlled by a new optional "feature flag" called `workspaceTokenServiceRefreshTokenAtLogin`. If this is set to `true` it will refresh the WTS token directly at portal login (recommended mode). If not set, this refresh happens only when the user enters the workspace section of the portal (default/old/previous mode).

### Deployment changes

- this change assumes WTS is already a core/central service. The dependency is hereby moved to the homepage (instead of the ./workspace page)

### DIFF
--- a/docs/portal_config.md
+++ b/docs/portal_config.md
@@ -182,6 +182,7 @@ Below is an example, with inline comments describing what each JSON block config
     "explorerFilterValuesToHide": ["array of strings"], // optional, Values set in array will be hidden in guppy filters. Intended use is to hide missing data category from filters, for this it should be set to the same as `missing_data_alias` in Guppy server config
     "studyRegistration": true, // optional, whether to enable the study registration feature
     "workspaceRegistration": true, // optional, whether to enable the workspace registration feature
+    "workspaceTokenServiceRefreshTokenAtLogin": true, // optional, whether to refresh the WTS token directly at portal login (recommended mode). If not set, this refresh happens only when the user enters the workspace section of the portal (default/old/previous mode).
   },
   "dataExplorerConfig": { // required only if featureFlags.explorer is true; configuration for the Data Explorer (/explorer); can be replaced by explorerConfig, see Multi Tab Explorer doc
     "charts": { // optional; indicates which charts to display in the Data Explorer
--- a/src/Login/ProtectedContent.jsx
+++ b/src/Login/ProtectedContent.jsx
@@ -29,8 +29,8 @@ let lastAuthMs = 0;
  * @param filter {() => Promise} optional filter to apply before rendering the child component
  */
 class ProtectedContent extends React.Component {
-  constructor(props, context) {
-    super(props, context);
+  constructor(props) {
+    super(props);
     this.state = {
       authenticated: false,
       dataLoaded: false,
@@ -58,8 +58,8 @@ class ProtectedContent extends React.Component {
           )
           .then(
             () => this.checkLoginStatus(store, this.state)
-              .then((newState) => this.props.public || this.checkQuizStatus(newState))
-              .then((newState) => this.props.public || this.checkApiToken(store, newState)),
+              .then((newState) => ((this.props.public) ? { ...newState, redirectTo: null } : this.checkQuizStatus(newState))) // don't redirect for public pages
+              .then((newState) => ((this.props.public) ? { ...newState, redirectTo: null } : this.checkApiToken(store, newState))),
           )
           .then(
             (newState) => {
--- a/src/Login/ProtectedContent.jsx
+++ b/src/Login/ProtectedContent.jsx
@@ -73,8 +73,9 @@ class ProtectedContent extends React.Component {
                 latestState.dataLoaded = true;
                 this.setState(latestState);
                 if (newState.authenticated && isEnabled('workspaceTokenServiceRefreshTokenAtLogin')) {
-                    // initialize WTS:
-                    initWorkspaceRefreshToken();
+                  // initialize WTS:
+                  const { location } = this.props; // this is the react-router "location"
+                  initWorkspaceRefreshToken(location);
                 }
               };
               return filterPromise.then(
--- a/src/Login/ProtectedContent.jsx
+++ b/src/Login/ProtectedContent.jsx
@@ -13,6 +13,8 @@ import ReduxAuthTimeoutPopup from '../Popup/ReduxAuthTimeoutPopup';
 import ReduxSystemUseWarningPopup from '../Popup/SystemUseWarningPopup';
 import { intersection, isPageFullScreen } from '../utils';
 import './ProtectedContent.css';
+import isEnabled from '../helpers/featureFlags';
+import { initWorkspaceRefreshToken } from '../Workspace/WorkspaceRefreshToken';
 
 let lastAuthMs = 0;
 
@@ -70,6 +72,10 @@ class ProtectedContent extends React.Component {
                 const latestState = { ...newState };
                 latestState.dataLoaded = true;
                 this.setState(latestState);
+                if (newState.authenticated && isEnabled('workspaceTokenServiceRefreshTokenAtLogin')) {
+                    // initialize WTS:
+                    initWorkspaceRefreshToken();
+                }
               };
               return filterPromise.then(
                 finish, finish,
--- a/src/Workspace/WorkspaceRefreshToken.js
+++ b/src/Workspace/WorkspaceRefreshToken.js
@@ -1,20 +1,27 @@
 import { fetchWithCreds } from '../actions';
 import { wtsPath } from '../localconf';
 
+let lastRefreshMs = 0;
+const debounceMs = 60000;
+
 // start workspace session for WTS, call optional connectedCallBack if initialized/connected
 export const initWorkspaceRefreshToken = (connectedCallBack) => {
-    console.log('init/renew WTS refresh token...');
-    fetchWithCreds({
-      path: `${wtsPath}connected`,
-      method: 'GET',
-    })
-      .then(
-        ({ status }) => {
-          if (status !== 200) {
-            window.location.href = `${wtsPath}/authorization_url?redirect=${window.location.pathname}`;
-          } else if (connectedCallBack) {
-              connectedCallBack();
-          }
-        },
-      );
-  }
+    const nowMs = Date.now();
+    if (nowMs - lastRefreshMs > debounceMs) {
+      console.log('init/renew WTS refresh token...');
+      fetchWithCreds({
+        path: `${wtsPath}connected`,
+        method: 'GET',
+      })
+        .then(
+          ({ status }) => {
+            if (status !== 200) {
+              window.location.href = `${wtsPath}/authorization_url?redirect=${window.location.pathname}`;
+            } else if (connectedCallBack) {
+                connectedCallBack();
+            }
+          },
+        );
+        lastRefreshMs = Date.now();
+    }
+}
--- a/src/Workspace/WorkspaceRefreshToken.js
+++ b/src/Workspace/WorkspaceRefreshToken.js
@@ -1,0 +1,20 @@
+import { fetchWithCreds } from '../actions';
+import { wtsPath } from '../localconf';
+
+// start workspace session for WTS, call optional connectedCallBack if initialized/connected
+export const initWorkspaceRefreshToken = (connectedCallBack) => {
+    console.log('init/renew WTS refresh token...');
+    fetchWithCreds({
+      path: `${wtsPath}connected`,
+      method: 'GET',
+    })
+      .then(
+        ({ status }) => {
+          if (status !== 200) {
+            window.location.href = `${wtsPath}/authorization_url?redirect=${window.location.pathname}`;
+          } else if (connectedCallBack) {
+              connectedCallBack();
+          }
+        },
+      );
+  }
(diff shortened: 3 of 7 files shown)


pieterlukasse:
@m0nhawk @kmhernan who else needs to be added as reviewer here?

Avantol13:
I'm not sure I agree with the concept here. WTS should not be a hard dependency of portal. We have data commons that need portal but don't need WTS

Avantol13:
The assumption that WTS is configured and used everywhere is not a safe assumption

pieterlukasse:
@Avantol13 thanks for the review! You bring up a good point, and I thought about this as well when making this PR. One reason I added it here is because I see other workspace related items in `src/index.jsx` already. Do these then also not belong there, or is there a part of "workspace" that functions and is useful without WTS?

Avantol13:
> @Avantol13 thanks for the review! You bring up a good point, and I thought about this as well when making this PR. One reason I added it here is because I see other workspace related items in `src/index.jsx` already. Do these then also not belong there, or is there a part of "workspace" that functions and is useful without WTS?

Honestly I don't think workspace things should necessarily be there either, but adding more coupling (for example to WTS) is just making things worse in my opinion. I don't have a ton of portal experience so ultimately I need to defer to @mfshao 

imo portal should support commons that don't have workspaces at all, commons that only have workspaces for themselves (e.g. no need to WTS), and it should support ecosystems where we want workspaces and WTS. 

We have running, production examples of all of those things, and it seems like this PR presumes that we want everything to be an ecosystem with workspaces and WTS, and that just isn't the case.

pieterlukasse:
thanks @Avantol13, that makes sense. I agree that in this case probably other references to workspaces should also be removed. @mfshao what do you think?

pieterlukasse:
@mfshao thanks for your comment. I added this commit https://github.com/uc-cdis/data-portal/pull/1095/commits/14562674d3fb5d43d75d662d97cd73ab9abce9d6 which will trigger the initialization/refresh based on a very specific feature flag (I called it `workspaceTokenServiceRefreshTokenAtLogin`). This is still based on @Avantol13 's comment that some installations can have workspace without needing WTS. Please let me know what you think.

docs/portal_config.md
```diff
@@ -182,6 +182,7 @@ Below is an example, with inline comments describing what each JSON block config
     "explorerFilterValuesToHide": ["array of strings"], // optional, Values set in array will be hidden in guppy filters. Intended use is to hide missing data category from filters, for this it should be set to the same as `missing_data_alias` in Guppy server config
     "studyRegistration": true, // optional, whether to enable the study registration feature
     "workspaceRegistration": true, // optional, whether to enable the workspace registration feature
+    "workspaceTokenServiceRefreshTokenAtLogin": true, // optional, whether to refresh the WTS token directly at portal login (recommended mode). If not set, this refresh happens only when the user enters the workspace section of the portal (default/old/previous mode).
```
Avantol13, line 185: I feel like the default here should be `false` if we're trying to be backwards compatible w/o needing to update config. That way, people that want to take advantage of this can turn it on in there config but there's no change to those who want the old previous mode

pieterlukasse: Yes, the default is "not set", so `isEnabled('workspaceTokenServiceRefreshTokenAtLogin')` should return `false`.

src/Workspace/index.jsx
```diff
@@ -73,19 +74,7 @@ class Workspace extends React.Component {
...
-      .then(
-        ({ status }) => {
-          if (status !== 200) {
-            window.location.href = `${wtsPath}/authorization_url?redirect=${window.location.pathname}`;
-          } else {
-            this.connected();
-          }
-        },
-      );
+    initWorkspaceRefreshToken(this.connected);
```
mfshao, line 77: so, if this new feature is enabled, when user navigates from a page to workspace, another wts call will be dispatched? seems like a bit redundant

src/index.jsx
```diff
@@ -67,6 +67,12 @@ import ErrorWorkspacePlaceholder from './Workspace/ErrorWorkspacePlaceholder';
 import { ReduxStudyViewer, ReduxSingleStudyViewer } from './StudyViewer/reduxer';
 import NotFound from './components/NotFound';
 import ErrorPage403 from './components/ErrorPage403';
+import { initWorkspaceRefreshToken } from './Workspace/WorkspaceRefreshToken';
+
+if (isEnabled('workspaceTokenServiceRefreshTokenAtLogin')) {
```
mfshao, line 72: other index pages (which are being used by other portal bundles) also needs this change?

mfshao (commented):
I don't think we support "workspace without WTS" a long while ago... 

But this PR has some linting issues that needs to be fixed

src/index.jsx
pieterlukasse: not sure. Is this one not used by all?

src/Workspace/index.jsx
pieterlukasse: Good point... But I think that if we want to do this with a feature flag that is `false` by default, then we will need to keep some of this redundancy for backwards compatibility purposes. Also, an extra refresh is not so bad...it is better than _no_ refresh (which happens if users use the portal but don't go into the workspace for a long period - a scenario we have in our project).

pieterlukasse: Please note that this diff is technically not really a code/behaviour change: I basically moved the code on the left into a generic function that can be used in different places.

pieterlukasse: Another option is to make a breaking change and make a feature flag for workspace as a whole, and make this flag `true` by default. This would simplify the code, but would require some special attention from teams that have a deployment that does not use workspace/ WTS.

paulineribeyre: why not check if `workspaceTokenServiceRefreshTokenAtLogin` is enabled, and only make the call if it's not?

src/index.jsx
paulineribeyre: no, some Data Commons have their own landing page, so they have their own index page (`covid19Index.jsx`, `nctIndex.jsx`, etc)

src/Workspace/index.jsx
pieterlukasse: That could work, but IMO it is an unnecessary optimisation. But if you prefer this, I can easily add it. Let me know.

src/index.jsx
pieterlukasse: so do we need to add this change there as well?

src/Login/ProtectedContent.jsx
```diff
@@ -70,6 +72,10 @@ class ProtectedContent extends React.Component {
                 const latestState = { ...newState };
                 latestState.dataLoaded = true;
                 this.setState(latestState);
+                if (newState.authenticated && isEnabled('workspaceTokenServiceRefreshTokenAtLogin')) {
+                    // initialize WTS:
+                    initWorkspaceRefreshToken();
```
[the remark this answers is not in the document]
pieterlukasse: @paulineribeyre would that be a problem? I think it is a really light call.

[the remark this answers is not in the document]
paulineribeyre: Yes i think that's a problem, from a quick search i see 26 `ProtectedContents` in `src/index.jsx`, we should not make so many API calls every time someone loads the page. Can we make a single call to WTS when we portal is loaded?

src/index.jsx
paulineribeyre: The logic needs to happen for every Commons, not just the ones that use the default landing page

src/Workspace/index.jsx
paulineribeyre: it makes sense to me... depending on the flag, the call is made either globally, or when the workspace page loads, and not both. But we can defer to @mfshao who brought this up in the first place

src/Login/ProtectedContent.jsx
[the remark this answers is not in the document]
pieterlukasse: Interesting, since the same could be said regarding the call to `checkLoginStatus` just a few lines up:

```js
.then(
            () => this.checkLoginStatus(store, this.state)
```
Maybe that is why `checkLoginStatus` has [this code ](https://github.com/uc-cdis/data-portal/pull/1095/files#diff-c8ea73aaef1661450270bc0f167f41abd12c3d3a78f9894a48460f3727aeb0baR123-R126) to avoid repeated calls? Should we do something similar here? Or should we try to find another location for our init code?
```js
  checkLoginStatus = (store, initialState) => {
    const newState = { ...initialState };
    const nowMs = Date.now();
    newState.authenticated = true;
    newState.redirectTo = null;
    newState.user = store.getState().user;
    if (nowMs - lastAuthMs < 60000) {
      // assume we're still logged in after 1 minute ...
      return Promise.resolve(newState);
    }
  ...
```
@mfshao what would you suggest for this?

src/index.jsx
pieterlukasse: Ok, I'll add that

src/Workspace/index.jsx
mfshao: That is the point I bring this up: we should be able to add a check so this didn't get called repeatly

src/Login/ProtectedContent.jsx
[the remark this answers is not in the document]
mfshao: you can do something similar like this

or change in your wts init code so that if a user is not logged in, don't force them to do an OIDC dance with WTS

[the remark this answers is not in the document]
paulineribeyre: i am curious why you chose to add it to `ProtectedContent` instead of `index.jsx` like in your initial code?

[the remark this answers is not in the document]
pieterlukasse: @paulineribeyre I was asked by @mfshao to do this...I now realise this was in slack and is not here in the PR comments. 

@mfshao do you mind replying to this here?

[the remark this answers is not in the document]
pieterlukasse: @paulineribeyre @mfshao I added a proposal here in this last commit: https://github.com/uc-cdis/data-portal/pull/1095/commits/82b7b20dd878e31ad3137ba41619ed758e315f65. Please let me know what you think. It is based on the current solution for `checkLoginStatus` in `ProtectedContent.jsx`

src/Workspace/index.jsx
pieterlukasse: Fixed here: https://github.com/uc-cdis/data-portal/pull/1095/commits/882043f31b3f99be0e18cc94337e64b8ba8ae442

src/index.jsx
pieterlukasse: Or not... This change has been reverted in favor of the one in `src/Login/ProtectedContent.jsx` 😅

paulineribeyre (commented):
lgtm, deferring to @mfshao

src/Workspace/WorkspaceRefreshToken.js
```diff
@@ -0,0 +1,27 @@
...
+    if (nowMs - lastRefreshMs > debounceMs) {
+      console.log('init/renew WTS refresh token...');
+      fetchWithCreds({
+        path: `${wtsPath}connected`,
+        method: 'GET',
+      })
+        .then(
+          ({ status }) => {
+            if (status !== 200) {
+              window.location.href = `${wtsPath}/authorization_url?redirect=${window.location.pathname}`;
```
mfshao, line 19: Notice that the redirect here only contains `window.location.pathname`, so if the page that the user is at contains some query params in its URL, they will be lost

This was not a problem before because WTS login redirect only happens in certian pages that don't need query params. But if you are moving this redirect so it can happen virtually from any portal pages, you need to accomodate accordingly, because this will have a product-wise impact

Also there are some location and history based redirect/state restore happens in other pages of portal. We should make sure this change doens't breaks them

mfshao (commented):
I think we should test this features more. At least we should configure the portal to have some public/protected pages and try to login with user who already has WTS connected and who doesn't have WTS connected. To see if there will be any unexpected behaviors

mfshao:
Also the tests need to get fixed before it can be merged

src/Workspace/WorkspaceRefreshToken.js
pieterlukasse: hi @mfshao, as discussed, I've update the redirect code to use the same logic as is used in the `src/Login/Login.jsx` page. See latest commit added to this PR: https://github.com/uc-cdis/data-portal/pull/1095/commits/8950eec53745de249fe0445cc6dffdce212fbfdf

Please let me know if this is the change you were looking for.

src/Workspace/index.jsx
```diff
@@ -73,19 +73,13 @@ class Workspace extends React.Component {
...
-          if (status !== 200) {
-            window.location.href = `${wtsPath}/authorization_url?redirect=${window.location.pathname}`;
-          } else {
-            this.connected();
-          }
-        },
-      );
+    // Check if workspaceTokenServiceRefreshTokenAtLogin is NOT set.
+    // Because if is already enabled, then an extra refresh is not
+    // really needed, since it has already happened at login:
```
mfshao, line 78:
```suggestion
    // really needed, since it has already happened at login, so just call the callback:
```

mfshao (dismissed):
looks cool

pieterlukasse (commented):
tested locally:
- when `workspaceTokenServiceRefreshTokenAtLogin` is `false`: WTS refresh occurs only at `/workspace`
- when `workspaceTokenServiceRefreshTokenAtLogin` is `true`: WTS refresh occurs only at login / browser refresh event

 ✅

mfshao (approved):

PR #1106, opened 2022-09-26
return local arborist if no externally connected commons

aartivnkt:
Jira Ticket:
https://ctds-planx.atlassian.net/browse/BRH-259

If no externally connected commons, return results from local arborist in portal
Make sure to follow the DEV guidelines (https://gen3.org/resources/developer/dev-introduction/) before asking for review.

- Describe what this pull request does.
- Add short descriptive bullet points for each section if relevant. Keep in mind that they will be parsed automatically to generate official release notes.
- Test manually.
- Maintain or increase the test coverage (if relevant).
- Update the documentation, or justify if not needed.

-->

### Bug Fixes
- return local arborist if no externally connected commons

### DIFF
--- a/src/actions.js
+++ b/src/actions.js
@@ -532,11 +532,18 @@ export const fetchUserAuthMapping = async (dispatch) => {
   let authMapping;
   let aggregateAuthMappings = {};
   if (authzMappingURL === wtsAggregateAuthzPath) {
-    authMapping = fetchedAuthMapping[hostnameWithSubdomain];
-    aggregateAuthMappings = fetchedAuthMapping;
-    if (Object.keys(aggregateAuthMappings).length === 0) {
-      authMapping = fetchedAuthMapping;
-      aggregateAuthMappings[hostnameWithSubdomain] = fetchedAuthMapping;
+    // If WTS enabled and no externally connected commons and
+    // fetchedAuthMapping is empty through WTS, get local arborist mapping
+    if (Object.keys(fetchedAuthMapping).length === 0){
+	const localArboristMapping = await fetch(authzMappingPath)
+        authMapping = localArboristMapping[hostnameWithSubdomain];
+        aggregateAuthMappings = localArboristMapping;
+    }
+    // WTS enabled and there are externally connected commons
+    // fetchedAuthMapping is not empty
+    else{
+        authMapping = fetchedAuthMapping[hostnameWithSubdomain];
+        aggregateAuthMappings = fetchedAuthMapping;
     }
   } else {
     authMapping = fetchedAuthMapping;
--- a/src/actions.js
+++ b/src/actions.js
@@ -501,51 +501,40 @@ export const fetchUserAccess = async (dispatch) => {
   });
 };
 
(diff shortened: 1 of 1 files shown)


src/actions.js
```diff
@@ -534,6 +534,10 @@ export const fetchUserAuthMapping = async (dispatch) => {
   if (authzMappingURL === wtsAggregateAuthzPath) {
     authMapping = fetchedAuthMapping[hostnameWithSubdomain];
     aggregateAuthMappings = fetchedAuthMapping;
+    if (Object.keys(aggregateAuthMappings).length === 0) {
+      authMapping = fetchedAuthMapping;
+      aggregateAuthMappings[hostnameWithSubdomain] = fetchedAuthMapping;
```
mfshao, line 539: can you explain why this works? we are already doing `aggregateAuthMappings = fetchedAuthMapping;` in L536 so if L537 is true, doesn't that means `fetchedAuthMapping` is also empty?

mfshao: I think the fall back should happen in an much eariler place, like https://github.com/uc-cdis/data-portal/pull/1106/files#diff-98405f0e77bce0c636c493239fdc977dd78f320f5313d5378b701c3effef541cR518

Logic should be:
1. if `discoveryUseAggWTS` is on, try to fetch agg WTS path first. If that request fails with a non-200, then fetch local arborist
2. if `discoveryUseAggWTS` is off, then just fetch local arborist

aartivnkt: Yeah, you're right. All we're trying to do here is if agg WTS is on, and that `aggregateAuthMappings` is empty because there are no externally connected commons, return local arborist

aartivnkt: I'll see if I can move this to earlier

aartivnkt: Sure, Sai, go for it

nss10: > if discoveryUseAggWTS is on, try to fetch agg WTS path first. If that request fails with a non-200, then fetch local arborist

One slight modification to this comment is that -- The wts/aggregate wouldn't respond with a non-200 response, but simply returns an empty object in this case. 

I've made necessary changes and pushed a fresh [commit](https://github.com/uc-cdis/data-portal/pull/1106/commits/e2f6543c04550599ed29f599485b26854ea67da2) to this branch, please take a look. 

`Note: This is agreed to be a band-aid solution until an efficient solution is introduced as a part of this ->` [JIRA ticket](https://ctds-planx.atlassian.net/browse/BRH-260)

mfshao: What about when WTS is unstable and returns 502? 🤔

nss10: Ah! that is a good point. Co-incidentally, I think [this part](https://github.com/uc-cdis/data-portal/blob/e2f6543c04550599ed29f599485b26854ea67da2/src/actions.js#L511-L514) of the logic covers that case as well.

Since we get a `{}` response in either of the cases, we'd go ahead and pull data from local arborist

mfshao (approved):
awesome 🎉

PR #1125, opened 2022-10-19
Feat/combine gwas workflow progress bar

jarvisraymond-uchicago:
Jira Ticket: [VADC-216](https://ctds-planx.atlassian.net/browse/VADC-216)

### New Features
This adds a progress bar to show the user's progress through the steps of GWAS Workflow V2. 
* Per the design, the steps text changes from edit to select for the first three steps depending on their progress
* The color of the step should change if it is the step the user is currently on
* A "New to GWAS? Get started here" button is included, but the tutorial logic will be done in task VADC-250
* A Storybook for the component is included and mocks up each step's UI
* A JEST test is included to test that an active class is added based on the _current_ prop's value and to test that the text changes based on the _current_ prop's value

### DESIGN
image

### STORYBOOK
output

### IMPLEMENTATION
output


src/Analysis/GWASV2/Shared/ProgressBar/ProgressBar.css
```diff
@@ -0,0 +1,109 @@
...
+
+.progress-bar .ant-steps-item-active:before {
+  content: "";
+  position: absolute;
+  left: 0;
+  bottom: 0;
+  width: 0;
+  border-bottom: solid 3px #ef8523;
+  animation: border_anim 0.3s linear forwards;
+}
```
alilahrime, line 41: nice ui polish

src/Analysis/GWASV2/GWASContainer.jsx
```diff
@@ -1,100 +1,87 @@
...
     <React.Fragment>
+      <ProgressBar current={current} />
       {/* Inline style block needed so centering rule doesn't impact other workflows */}
-      <style>{`.analysis-app__actions > div:nth-child(1) {margin: 0 auto; }`}</style>
-      <div className="GWASV2">
-        <Space direction={"vertical"} style={{ width: "100%" }}>
-          <div className="steps-content">
+      <style>
+        {'.analysis-app__actions > div:nth-child(1) { width: 100%; }'}
+      </style>
```
alilahrime, line 47: can you include this styling without it cascading to other steps

alilahrime (approved):

src/Analysis/GWASV2/GWASContainer.jsx
jarvisraymond-uchicago: Per our discussion, let's hold off on this. It results in the progress bar not taking up the full width. My guess is because the code changing the margins have been removed, it won't impact the subsequent steps

image

src/Analysis/GWASV2/Shared/ProgressBar/ProgressBar.jsx
```diff
@@ -0,0 +1,29 @@
...
+        {gwasV2Steps.map((item, index) => (
+          <Step
+            key={item.title}
+            icon={<React.Fragment>{index + 1}</React.Fragment>}
+            title={`${current <= index ? item.title : item.secondaryTitle}`}
+          />
+        ))}
+      </Steps>
+    </div>
+    <Button>New to GWAS? Get started here</Button>
```
pieterlukasse, line 21: this button will render on top of step4 if the screen is narrow enough
Screenshot 2022-10-20 at 17 23 57

src/Analysis/GWASV2/Shared/ProgressBar/ProgressBar.stories.jsx
```diff
@@ -0,0 +1,29 @@
...
+
+export const ThirdStepActive = Template.bind({});
+ThirdStepActive.args = {
+  current: 2,
+};
+
+export const FourthStepActive = Template.bind({});
+FourthStepActive.args = {
+  current: 3,
+};
```
pieterlukasse, line 29: fantastic! Thanks 👍

src/Analysis/GWASV2/Shared/ProgressBar/ProgressBar.test.jsx
```diff
@@ -0,0 +1,135 @@
...
+  it('should render second step', () => {
+    testElementText(wrapper, 2, '2Edit Outcome Phenotype');
+  });
+  it('should render third step', () => {
+    testElementText(wrapper, 3, '3Edit Covariate Phenotype');
+  });
+  it('should render fourth step', () => {
+    testElementText(wrapper, 4, '4Configure GWAS');
+  });
+});
```
pieterlukasse, line 135: thanks for adding tests 👍 

I would like to propose a few simplifications to enhance the maintainability without loosing much:
- `should render ... step with active class when current is ...` -> implement a loop of 4?
- "Test Dynamic Text for Step" -> this could also be a single test with a loop instead of 4 tests. Also do not test against the whole text but just check if the step starts with "Edit". In this way, the test does not break if we decide to make small changes to the title.

src/Analysis/GWASV2/Shared/constants.js
```diff
@@ -0,0 +1,19 @@
...
+  },
+  {
+    title: 'Select Covariate Phenotype',
+    secondaryTitle: 'Edit Covariate Phenotype',
+  },
+  {
+    title: 'Configure GWAS',
+    secondaryTitle: 'Configure GWAS',
+  },
+];
```
pieterlukasse, line 19: regarding my previous comments in the unit tests file...maybe use these variables instead of repeating the titles in the tests?

pieterlukasse (commented):
@jarvisraymond-uchicago the new progress bar looks great and I like the subtle animation of the lines under the step names 👍 

Regarding the behavior with next and previous buttons: I found that in step1 it is possible to click "next" even when no study population is selected. Instead, the navigation should only be enabled after a selection is made. This conditional enabling of the "next" button will be the case in most of the steps.

src/Analysis/GWASV2/Shared/ProgressBar/ProgressBar.jsx
jarvisraymond-uchicago: Good catch! This should be fixed now.

src/Analysis/GWASV2/Shared/constants.js
jarvisraymond-uchicago: Good idea! I imported the constants file and used those variables.

src/Analysis/GWASV2/Shared/ProgressBar/ProgressBar.test.jsx
jarvisraymond-uchicago: Interesting idea! I did this and got it to work. 

A concern of mine is that this approach can lead to more difficult to interpret tests. For example: if something breaks it might be easier for a developer to quickly interpret explicitly stated test code compared with the same logic implemented with loops.

jarvisraymond-uchicago:
> @jarvisraymond-uchicago the new progress bar looks great and I like the subtle animation of the lines under the step names 👍
> 
> Regarding the behavior with next and previous buttons: I found that in step1 it is possible to click "next" even when no study population is selected. Instead, the navigation should only be enabled after a selection is made. This conditional enabling of the "next" button will be the case in most of the steps.

Good point! I added back in the logic for ```nextButtonEnabled``` so the next button should only be enabled after a user selects a selected study population cohort in view 1.

src/Analysis/GWASV2/Shared/ProgressBar/ProgressBar.test.jsx
pieterlukasse: I agree that test readability is also super important. I think the tests are still readable.

src/Analysis/GWASV2/Shared/ProgressBar/ProgressBar.jsx
pieterlukasse: nice!

pieterlukasse (approved):

PR #1128, opened 2022-10-20
feat(eslintOnlyOnCurrentBranch): Added new command eslint-new

jarvisraymond-uchicago:
Jira Ticket: [VADC-267](https://ctds-planx.atlassian.net/browse/VADC-267)

### New Features
This adds a new command to the scripts section of data-portal/package.json
```eslint-new```

A developer should be able to run this using: 
```npm run eslint-new```

This runs eslint, but limits it to files that have been changed in the developer's current branch and have been added (i.e. with ```git add -A```). This should prevent the issue of unrelated files showing up in FE developer's branch after running eslint. FE developer's should probably not use ```npm run eslint``` anymore after this is merged (unless they want to check the entire repository). 

This line of code is based on: https://stackoverflow.com/questions/54511168/enable-eslint-only-for-edited-files

The codebase's README.md has been updated with a Linting section as well to document this for other developers.

### CODE REVIEW

**Comment by kmhernan** [comment]
@jarvisraymond-uchicago I am wondering if we need to document this anywhere? We may not have any existing documentation, but if we do we may want to at least mention what this is?

**Comment by alilahrime** [comment]
> 

` "lint": "npm run lint:eslint"` is a command that comes up often in multiple scripts for diff envs in`cdis-manifest`, otherwise not sure where its documented explicitly readme style

**Comment by kmhernan** [comment]
> @jarvisraymond-uchicago I am wondering if we need to document this anywhere? We may not have any existing documentation, but if we do we may want to at least mention what this is?

@jarvisraymond-uchicago I don't see lint mentioned anywhere in current docs so ignore my comment

**Comment by jarvisraymond-uchicago** [comment]
> > @jarvisraymond-uchicago I am wondering if we need to document this anywhere? We may not have any existing documentation, but if we do we may want to at least mention what this is?
> 
> @jarvisraymond-uchicago I don't see lint mentioned anywhere in current docs so ignore my comment

Per Pieter's suggestion, I've updated the codebase's README.md to include a section on Linting commands.

File: package.json [hunk 1]
```diff
@@ -190,6 +190,7 @@
     "schema": "node ./data/getSchema",
     "params": "node ./data/getTexts > src/params.js",
     "eslint": "./node_modules/.bin/eslint --ext js --ext jsx --ext ts --ext tsx --ignore-pattern '**/__generated__/**' --ignore-pattern '**/*.min.js' --ignore-pattern '**/setupJest.js' --ignore-pattern '**/gqlHelper.js' --ignore-pattern '**/formHooks.jsx' --ignore-pattern '**/*.stories.jsx' --fix src data",
+    "eslint-new": "./node_modules/.bin/eslint --ext js --ext jsx --ext ts --ext tsx --ignore-pattern '**/__generated__/**' --ignore-pattern '**/*.min.js' --ignore-pattern '**/setupJest.js' --ignore-pattern '**/gqlHelper.js' --ignore-pattern '**/formHooks.jsx' --ignore-pattern '**/*.stories.jsx' --fix $(git diff --name-only HEAD | xargs)",
```
alilahrime, line 193: I dont think this command is behaving as expected. I tested locally and this was results after running following commands 
`npm run eslint-new`
Screen Shot 2022-10-24 at 10 35 38 AM
`npm run eslint`
Screen Shot 2022-10-24 at 10 37 09 AM

alilahrime: the new command for this pr picks up oddities in files we're not supposed to change and it didnt lint the random spacings I added like the original [second] command does

pieterlukasse, line 193: This works, although I would probably expect something like `git diff --name-only master` instead.

README.md
```diff
@@ -302,3 +324,11 @@ class Component extends React.Component {
   }
 }
 ```
+
+### Linting
+
+We use ESLint and Stylelint to lint and automatically format code.
+
+- `npm run eslint` Will run ESLint on the entire code base and automatically try to fix all JS and JS like files.
+- `npm run eslint-new` Will run ESLint only on newly added files in the current git branch and automatically try to fix the JS and JS like files.
```
pieterlukasse, line 333: instead of "only on newly added files in the current git branch" this should state "only on uncommitted files". OR you could change HEAD to `master` in the command.

jarvisraymond-uchicago: I've updated HEAD to master

package.json
jarvisraymond-uchicago: I've updated HEAD to master

pieterlukasse (approved):
Looks good to me 👍 

Did a simple test as well and it worked!

PR #1129, opened 2022-10-27
Feat/combine gwas workflow attrition table styling and setup

jarvisraymond-uchicago:
Jira Ticket: [vadc-243](https://ctds-planx.atlassian.net/browse/vadc-243)

### New Features
This updates adds the attrition tables to the GWAS V2 workflow and provides styling for it. 
* Additional logic to update the table's functions, props, parameters etc. will be done in ticket VADC-277. 
* A Storybook is included, the Storybook will be updated and JEST tests will be written as part of VADC-277
* Currently whether a Euler diagram icon or chart icon renders for covariates is chosen randomly. Logic will be written as part of VADC-277 to determine the chart Icon based on the covariate type

### Bug Fixes
* This removes the duplicate key warnings that were present in the initial workflows; the antD List component has been replaced with a JS map using the covariateSubsets array (line 82 in AttritionTable.jsx). 
* All tabular information is now rendered using table elements, making the markup more semantic, accessible, etc. 
* The command npm run eslint-new has an ignore pattern added to it, so it ignores css files. 

### Design 
image

#### Storybook
output
#### Application
output


alilahrime (approved):

pieterlukasse (approved):
overall looks good to me.

A few comments, questions, maybe for 277:
- should we have chart icon left aligned?
- should numbers be right aligned?
- in case of error, show something else instead of 0?
- one or two attrition tables:
  - this should be decided dynamically based on what is selected in Step 2

jarvisraymond-uchicago:
> overall looks good to me.
> 
> A few comments, questions, maybe for 277:
> 
>     * should we have chart icon left aligned?
> 
>     * should numbers be right aligned?
> 
>     * in case of error, show something else instead of 0?
> 
>     * one or two attrition tables:
>       
>       * this should be decided dynamically based on what is selected in Step 2

Sounds good, I think they'll need to be multiple clean up styling ticket where any discrepancies between the implementation and design can be remedied. Issues with the logic will need to be remedied in ticket 277 or subsequent tickets to that one.

PR #1133, opened 2022-11-03
test

mfshao:
attempt fix for synk issue


src/Login/Login.jsx
```diff
@@ -23,6 +23,12 @@ export const getUrlForRedirectLocation = (location) => {
   if (queryParams.next) {
     next = basename === '/' ? queryParams.next : basename + queryParams.next;
   }
+  const regexp = /^\/\w+|\/\?/gi;
```
pieterlukasse, line 26: maybe use 
```js
const regexp = /^\/\w+|\/$/gi;
```

PR #1138, opened 2022-11-16
Chore/add defaults for Euler diagram wrapper

pieterlukasse:
Jira Ticket: NA

### Improvements

- code made more uniform by adding defaults for euler diagram wrapper and using "null" instead of "undefined"


jarvisraymond-uchicago:
Looks good to me!

PR #1140, opened 2022-11-16
(HP-900): handle null value

george42-ctds:
Jira Ticket: [HP-900](https://ctds-planx.atlassian.net/browse/HP-900)

### Bug Fixes

* Handle null values in discovery page

### DIFF
--- a/src/Discovery/Discovery.tsx
+++ b/src/Discovery/Discovery.tsx
@@ -157,7 +157,7 @@ export const renderFieldContent = (content: any, contentType: 'string'|'paragrap
 };
 
 const highlightSearchTerm = (value: string, searchTerm: string, highlighClassName = 'matched'): {highlighted: React.ReactNode, matchIndex: number} => {
-  const matchIndex = value.toLowerCase().indexOf(searchTerm.toLowerCase());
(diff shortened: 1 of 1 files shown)


src/Discovery/Discovery.tsx
```diff
@@ -157,7 +157,7 @@ export const renderFieldContent = (content: any, contentType: 'string'|'paragrap
 };
 
 const highlightSearchTerm = (value: string, searchTerm: string, highlighClassName = 'matched'): {highlighted: React.ReactNode, matchIndex: number} => {
-  const matchIndex = value.toLowerCase().indexOf(searchTerm.toLowerCase());
+  const matchIndex = value ? value.toLowerCase().indexOf(searchTerm.toLowerCase()) : -1;
```
george42-ctds, line 160: Something in my data or config was triggering an error here. I haven't been able to reproduce the error after loading more data and doing another aggMDS.

mfshao: I think this will happen if a searchbale field doesn't have any value in metadata, which makes sense to have this fix

mfshao (approved):

PR #1142, opened 2022-11-22
Feat/gwas unified flow add attrition table

jarvisraymond-uchicago:
Jira Ticket: [VADC-325](https://ctds-planx.atlassian.net/browse/VADC-325)

### New Features
* This adds the attrition table back to the combined workflow; it is now configured to use the state created by the reducer. 

### Screenshots
output


src/Analysis/GWASV2/Shared/AttritionTableWrapper/AttritionTable/AttritionTableRow.jsx
```diff
@@ -96,13 +89,20 @@ const AttritionTableRow = ({
 
   const determineChartIcon = () => {
     if (rowType === 'Covariate' || rowType === 'Outcome') {
-      if (rowObject.variable_type === 'concept') {
+      if (
+        rowObject.variable_type === 'concept'
+        || 'concept_type' in rowObject
```
jarvisraymond-uchicago, line 94: The rowObject was missing the ```variable_type``` key and was causing the application to crash. I updated this to also check for the ```concept_type```to determine if a bar chart should be shown here as a work around.

src/Analysis/GWASV2/Shared/AttritionTableWrapper/AttritionTable/AttritionTableRow.jsx
```diff
@@ -111,7 +112,7 @@ const AttritionTableRow = ({
     if (rowType === 'Outcome') {
       return outcome.variable_type === 'concept'
         ? outcome.concept_name
-        : 'this is for case control, TODO';
+        : 'this is for case control, TODO!';
```
jarvisraymond-uchicago, line 115: Not sure what needs to be done here: Can we just output outcome.concept_name if there is not a outcome.variable_type? This returns "this is for case control, TODO!" when the outcome object is missing the variable_type key.

src/Analysis/GWASV2/Shared/AttritionTableWrapper/AttritionTable/AttritionTable.jsx
```diff
@@ -42,17 +37,21 @@ const AttritionTable = ({
...
         ],
-        provided_name: 'Autogenerated variable for filtering out Case Population',
+        provided_name:
+          'Autogenerated variable for filtering out Case Population',
       };
-    } return {};
+    }
+    return {};
   };
+  const objectHasData = (obj) => Object.keys(obj).length > 0;
```
pieterlukasse: do we need this back?

pieterlukasse (commented):
tested and it works 👍 

Code also looks good to me!

pieterlukasse (approved):

PR #1157, opened 2022-12-08
Feat/gwas unified flow table scroll

jarvisraymond-uchicago:
Jira Ticket: [VADC-240](https://ctds-planx.atlassian.net/browse/VADC-240)

### New Features
This adds vertical scroll bar functionality to the attrition table when the table size exceeds 300px, or about 5 covariate rows. 

### Bug Fixes
This updates AttritionTable.stories.jsx so it works again, with an changeable number of covariate rows for the first table 

### Screenshots
output


pieterlukasse (approved):
tested locally in the app and it works well!

PR #1159, opened 2022-12-12
adding covid19 community analysis tab

urvi-occ:
### Improvements

Added code to add new tab for COVID-19 Community Analysis tab in which we have three dashboards created in AWS Quicksight embed into Chicagoland Pandemic Response Commons


src/index.ejs
```diff
@@ -10,29 +11,32 @@
...
+    <meta http-equiv="Content-Security-Policy" content="default-src 'self' https://login.bionimbus.org https://wayf.incommonfederation.org; child-src blob:; img-src 'self' <%= htmlWebpackPlugin.options.imgSrc %> https://opendata.datacommons.io https://static.planx-pla.net data: https://*.s3.amazonaws.com blob:; script-src 'self' 'unsafe-eval' <%= htmlWebpackPlugin.options.scriptSrc %>; worker-src 'self' blob:; style-src 'self' 'unsafe-inline' https://fonts.googleapis.com; object-src 'none'; font-src 'self' data: https://fonts.googleapis.com https://fonts.gstatic.com; connect-src 'self' https://login.bionimbus.org https://wayf.incommonfederation.org https://opendata.datacommons.io https://static.planx-pla.net <%= htmlWebpackPlugin.options.connectSrc %>; frame-src <%= htmlWebpackPlugin.options.connectSrc %> 'self' https://auspice.planx-pla.net https://auspice.pandemicresponsecommons.org https://*.quicksight.aws.amazon.com;">
```
paulineribeyre, line 14: could you add a comment in `src/index.ejs` and `dev.html` to explain why we need this one, similar to [this](https://github.com/uc-cdis/data-portal/blob/99df0afb3ad99e4e8b2ca9de8888f842ecd2a7c1/dev.html#L10)?

src/Covid19Dashboard/Covid19Dashboard.less
```diff
@@ -152,3 +152,47 @@
...
+}
+
+.gen3-dashboard_col {
+  -ms-flex-preferred-size: 0;
+  flex-basis: 0;
+  -ms-flex-positive: 1;
+  flex-grow: 1;
+  float: left;
+  width: 50%;
+  // padding: 0.5vh;
```
paulineribeyre, line 185: let's remove the commented out lines in this file

src/Covid19Dashboard/Covid19Dashboard.less
```diff
@@ -152,3 +152,47 @@
...
+  display: -ms-flexbox;
+  display: flex;
+  -ms-flex-wrap: wrap;
+  flex-wrap: wrap;
+  margin-right: -15px;
+  margin-left: -15px;
+  // padding: 0.5vh;
+}
+
+.gen3-dashboard_col {
```
paulineribeyre, line 178: i'm not sure what is causing this, but when i resize the window:

at first the charts scale when i make the window wider:
Screen Shot 2022-12-16 at 1 19 58 PM

but after a certain width, the chart on the left gets very small:
Screen Shot 2022-12-16 at 1 20 10 PM

src/Covid19Dashboard/index.jsx
```diff
@@ -272,6 +273,33 @@ class Covid19Dashboard extends React.Component {
...
+            <TabPanel className='gen3-dashboard_panel'>
+              <div className='gen3-dashboard_quicksight'>
+                {/* this component doesn't need the mapboxAPIToken but it's a way to make
+                sure this is the COVID19 Commons and the iframe contents will load */}
+                { mapboxAPIToken
+                  && (
+                    <div className="gen3-dashboard_container">
+                      <div className="gen3-dashboard_row">
+                        <div className="gen3-dashboard_col">
+                          <iframe className="gen3-dashboard_frame" src="https://us-east-1.quicksight.aws.amazon.com/sn/embed/share/accounts/043745044068/dashboards/6652754a-1e60-4aaf-b160-1346cbc5fc66?directory_alias=gen3-dashboards">
```
paulineribeyre, line 285: the code lgtm but i have questions about the data we're displaying:
1. how often is it updated?
2. how is it updated? is there a cronjob? is the code public?
3. is the code pushing the data versioned? if something breaks, can we revert to an older version?

I'm asking because it looks like we are adding a new way of pushing data to the PRC. There are already multiple ways of pushing data, and as far as i know they are not documented(?)

paulineribeyre:
@urvi-occ is this work still in progress?

urvi-occ:
> @urvi-occ is this work still in progress?

no. It is done on our end.

src/index.ejs
urvi-occ: The reason for adding it into the two places is because `dev.html` is used for local development for portal and I thought it was standard practice to add it into it as well

src/Covid19Dashboard/index.jsx
urvi-occ: This URLs will be updated rarely and only if we are changing the whole dashboard, otherwise, we will update the dashboards from our end and this url will get the new changes. This URLS are dashboards from AWS Quicksight and we had a project with AWS Team in which our goal was to get amazon Quicksight dashboards into PRC and this PR is for the same.

src/Covid19Dashboard/Covid19Dashboard.less
urvi-occ: This is a really good find. I fixed it now.

urvi-occ:
> @urvi-occ is this work still in progress?

@paulineribeyre I am sorry about these changes and comments coming in this late. I have made changes as asked and updated the comments as well. :)

src/index.ejs
paulineribeyre: yes we need to add it, what i'm saying is that it would be nice to update the comment to explain what we use it for

src/Covid19Dashboard/index.jsx
paulineribeyre: Is the data pulled from the https://chicagoland.pandemicresponsecommons.org `cityofchicago` project using the Gen3 API?

urvi-occ: yes. I am pulling the data whenever the dashboard data needs to be updated. I am doing calculation on that data and I am uploading it into the env where we have Quicksight architecture setup. You can find the demo for it in here https://cdis.slack.com/files/U030AK74946/F04F0V2CPUN/ddi_reserach_tool_demo.mp4

paulineribeyre: I see, so you manually run the scripts and update the data. Is there a long term plan to automatize this?

urvi-occ:
@paulineribeyre do we have anything outstanding in this PR from our end?

src/Covid19Dashboard/index.jsx
urvi-occ: Yes, it is. But currently, the plan is to get those dashboards up and make it available to users

src/index.ejs
urvi-occ: Make sense. Thank you :)

paulineribeyre:
Tests passed in #1235

PR #1171, opened 2022-12-15
Feat/add size info to hare dropdown

pieterlukasse:
Jira Ticket: NA

### Bug Fixes

- show final population size in modal popup

Screenshot 2022-12-15 at 19 47 16

### DIFF
--- a/src/Analysis/GWASV2/Components/SelectHare/SelectHareDropDown.jsx
+++ b/src/Analysis/GWASV2/Components/SelectHare/SelectHareDropDown.jsx
@@ -1,4 +1,4 @@
-import React, { useState, useEffect } from 'react';
+import React from 'react';
 import PropTypes from 'prop-types';
 import { useQueries } from 'react-query';
 // import Dropdown from '@gen3/ui-component/dist/components/Dropdown';
@@ -15,10 +15,9 @@ const SelectHareDropDown = ({
   selectedCohort,
   covariates,
   outcome,
-  handleHareChange,
+  dispatch,
 }) => {
   const { source } = useSourceContext();
-  const [hares, setHares] = useState([]);
   const hareResults = useQueries(
     outcome.variable_type === 'concept'
       ? [
@@ -94,13 +93,18 @@ const SelectHareDropDown = ({
       ],
   );
 
-  const { status: statusX, data: dataX } = hareResults.at(
-    hareResults.length === 1 ? 0 : -1,
-  );
-
-  useEffect(() => {
-    setHares(dataX?.concept_breakdown);
-  }, [dataX]);
+  const getFinalPopulationSize = (selectedHareItem) => {
+    const hareBreakdownPart1 = hareResults[0].data.concept_breakdown.find(
+      (hare) => selectedHareItem.value === hare.concept_value,
+    ).persons_in_cohort_with_value;
+    if (outcome.variable_type === 'concept') {
+      return hareBreakdownPart1;
+    }
+    const hareBreakdownPart2 = hareResults[1].data.concept_breakdown.find(
+      (hare) => selectedHareItem.value === hare.concept_value,
+    ).persons_in_cohort_with_value;
+    return hareBreakdownPart1 + hareBreakdownPart2;
+  };
 
   const onChange = (h) => {
     const selectedHareOutput = {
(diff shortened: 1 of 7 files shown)


src/Analysis/GWASV2/Components/SelectHare/SelectHareDropDown.jsx
```diff
@@ -94,72 +93,59 @@ const SelectHareDropDown = ({
...
-
-  useEffect(() => {
-    setHares(dataX?.concept_breakdown);
-  }, [dataX]);
+  const getFinalPopulationSize = (selectedHareItem) => {
+    const hareBreakdownPart1 = hareResults[0].data.concept_breakdown.find((hare) => selectedHareItem.value === hare.concept_value).persons_in_cohort_with_value;
+    if (outcome.variable_type === 'concept') {
+      return hareBreakdownPart1;
+    } else {
+      const hareBreakdownPart2 = hareResults[1].data.concept_breakdown.find((hare) => selectedHareItem.value === hare.concept_value).persons_in_cohort_with_value;
```
jarvisraymond-uchicago, line 101: 1.) Are there a more descriptive variable names that could be used here other than hareBreakdownPart1 and hareBreakdownPart1 ? Before we were discussing 
```
Control Size        123
Case Size           456
Total Size          579
```
Is hareBreakdownPart1 the Control Size and part 2 the Case size? If so can the variable names be updated to reflect that information? 

Also, could a line break be added to break up the rather long lines (lines 97 & 101)? I have to scroll to read the function.

src/Analysis/GWASV2/Components/SelectHare/SelectHareDropDown.jsx
```diff
@@ -94,72 +93,59 @@ const SelectHareDropDown = ({
...
       type: ACTIONS.UPDATE_SELECTED_HARE,
       payload: selectedHareOutput,
     });
+    dispatch({
+      type: ACTIONS.UPDATE_FINAL_POPULATION_SIZE,
+      payload: getFinalPopulationSize(h),
+    });
   };
 
-  //   const getHareDescriptionBreakdown = (singleHare, allHares) => {
```
jarvisraymond-uchicago, line 117: Thank you for cleaning this!

src/Analysis/GWASV2/Components/SelectHare/SelectHareDropDown.jsx
```diff
@@ -94,72 +93,59 @@ const SelectHareDropDown = ({
...
-  useEffect(() => {
-    setHares(dataX?.concept_breakdown);
-  }, [dataX]);
+  const getFinalPopulationSize = (selectedHareItem) => {
+    const hareBreakdownPart1 = hareResults[0].data.concept_breakdown.find((hare) => selectedHareItem.value === hare.concept_value).persons_in_cohort_with_value;
+    if (outcome.variable_type === 'concept') {
+      return hareBreakdownPart1;
+    } else {
+      const hareBreakdownPart2 = hareResults[1].data.concept_breakdown.find((hare) => selectedHareItem.value === hare.concept_value).persons_in_cohort_with_value;
+      return hareBreakdownPart1 + hareBreakdownPart2;
```
jarvisraymond-uchicago, line 102: **QUESTIONI** I pulled this local and it looks like this is working good (correctly sums control and case when there is dichotomous), otherwise it outputs the lowest value when it is non-dichotomous outcome when done without a covariate phenotype.

However, I'm not sure if it includes the attrition from the covariate  phenotype in the second table, see screenshots below for non-hispanic white. Can you confirm this is expected behavior?

image

Final Size is 99 (I thought it would be 57?)
image

jarvisraymond-uchicago (commented):
Speedy work! I left some questions.

src/Analysis/GWASV2/Components/SelectHare/SelectHareDropDown.jsx
pieterlukasse: "other" is also a HARE group in some of our dummy data. It is not different from any of the other groups.

src/Analysis/GWASV2/Components/SelectHare/SelectHareDropDown.jsx
pieterlukasse: Yes, I think we should have this in a separate PR (i.e. the breaking down into Case and Control). This PR is just to get the total size, so right now in this PR it is not important to try to figure out which one is Case, which one is Control

pieterlukasse: Long lines are broken up now.

jarvisraymond-uchicago (approved):
Looks great!

PR #1172, opened 2022-12-15
Feat/gwas unified flow modal back button

jarvisraymond-uchicago:
Jira Ticket: [VADC-368](https://ctds-planx.atlassian.net/browse/VADC-368)
### New Features
This updates the modal back button in the GWAS++ app to send the user back to the start of step 4 (currentStep3) when dismissed using the cancel button. This helps avoid a situtation where the end user is unable to trigger the modal again without clicking the previous button and then the next button again

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -36,9 +36,9 @@ module.exports = {
     requireConfigFile: false,
     babelOptions: {
       plugins: [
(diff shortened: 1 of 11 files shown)


pieterlukasse (approved):
works really well 👍

PR #1174, opened 2022-12-15
Hiding DICOM Viewer button for the imaging studies that doesn't have one

m0nhawk:
Jira Ticket: N/A

### New Features
* Hiding DICOM Viewer button for the imaging studies that doesn't have one


src/GuppyDataExplorer/ExplorerTable/index.jsx
```diff
@@ -248,10 +252,46 @@ class ExplorerTable extends React.Component {
...
+    const haveField = this.props.rawData.filter((x) => Object.keys(x).includes(this.props.tableConfig.dicomViewerId));
+    if (haveField.length === this.props.rawData.length) {
+      this.setState({ loading: true });
+      // eslint-disable-next-line array-callback-return
+      Promise.all(this.props.rawData.map((x) => {
+        const dicomServerLink = `${hostname}dicom-server/dicom-web/studies/${x[this.props.tableConfig.dicomViewerId]}/series`;
+        return fetch(dicomServerLink, {
+          method: 'GET',
+        })
+          .then((resp) => ({ ...x, has_dicom_images: resp.headers.get('content-length') !== '22' }),
```
mfshao, line 269: what if the request fails here?

src/GuppyDataExplorer/ExplorerTable/index.jsx
```diff
@@ -1,6 +1,6 @@
-import React from 'react';
+import React, { useEffect, ReactNode } from 'react';
```
mfshao, line 1: why importing `ReactNode` here?

m0nhawk: I think here we can remove both, I don't see `useEffect` anywhere either.

src/GuppyDataExplorer/ExplorerVisualization/index.jsx
```diff
@@ -8,7 +8,7 @@ import { components } from '../../params';
 import { guppyUrl, tierAccessLevel, tierAccessLimit } from '../../localconf';
 import DataSummaryCardGroup from '../../components/cards/DataSummaryCardGroup';
 import ExplorerHeatMap from '../ExplorerHeatMap';
-import ExplorerTable from '../ExplorerTable';
+import ExplorerTable, {DicomHasImageLinkObserver} from '../ExplorerTable';
```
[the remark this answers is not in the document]
m0nhawk: I think this is some really old code.

src/GuppyDataExplorer/ExplorerTable/index.jsx
m0nhawk: The idea is to check if it has any response, so if it fails it means that the image is not available, and then we won't show the button.

mfshao: I guess my question is why it is checking for content length? does that endpoint returns 200 even if the DICOM viewer ID doesn;t exists?

m0nhawk: Yes, this endpoint will return 200 if there is no series available.

To be more specific, it would be an empty list:
```
[]
```

Which has `Content-Length: 22`.

mfshao: thanks for the explanation

mfshao (approved):

mfshao:
BTW there seems to be some linting issues

PR #1177, opened 2022-12-16
feat: get case control population sizes in modal

pieterlukasse:
Jira Ticket: [VADC-371](https://ctds-planx.atlassian.net/browse/VADC-371)

### New Features
- show Case/Control populations sizes as well

Screenshot:
Screenshot 2022-12-16 at 12 00 49


src/Analysis/GWASV2/Components/SelectHare/SelectHareDropDown.jsx
```diff
@@ -93,17 +93,26 @@ const SelectHareDropDown = ({
...
     ).persons_in_cohort_with_value;
-    return hareBreakdownPart1 + hareBreakdownPart2;
+    return [
+      { population: 'Control', size: finalControlCohortSize },
+      { population: 'Case', size: finalCaseCohortSize },
+      { population: 'Total', size: finalControlCohortSize + finalCaseCohortSize },
+    ];
   };
 
   const onChange = (h) => {
```
jarvisraymond-uchicago, line 118: **Question** On line 121 we're hard coding the concept_id passed to the dispatch with action UPDATE_SELECTED_HARE. Is this correct?

src/Analysis/GWASV2/Steps/ConfigureGWAS/ConfigureGWAS.jsx
```diff
@@ -255,10 +255,12 @@ const ConfigureGWAS = ({
               <div>Outcome Phenotype</div>
               <div>{outcome?.concept_name ?? outcome?.provided_name}</div>
             </div>
-            <div style={flexRow}>
-              <div>Final Population Size</div>
-              <div>{finalPopulationSize}</div>
-            </div>
+            {finalPopulationSizes.map((item, key) => (
+              <div key={key} style={flexRow}>
```
jarvisraymond-uchicago, line 259: **Nitpick**: Using an iterator as a react key I think is discouraged (https://dev.to/devcer/why-you-should-never-use-array-index-as-key-prop-in-your-react-applications-3m9o). If we certain the items are unique, I believe you could use an item as the key.  

ESLINT isn't complaining though so its probably OK

jarvisraymond-uchicago (changes requested):
Looks good, I tested locally and it works!

* I had a question about UPDATE_SELECTED_HARE hard coded value
* Had a nitpick about iterator as key that can probably be ignored since the linter doesn't seem to care. 
* Ran ESLINT-new to autoformat files, this included changes from Master. Then I updated SPRINT25 to be up-to-date with Master, but these updates are showing as changed files in this PR now; I don't think that's right; not sure why that is occurring.

src/Analysis/GWASV2/Components/SelectHare/SelectHareDropDown.jsx
pieterlukasse: good catch. Fixed.

pieterlukasse: good catch.

jarvisraymond-uchicago (approved):
Really good!

src/Analysis/GWASV2/Steps/ConfigureGWAS/ConfigureGWAS.jsx
pieterlukasse: we tried to fix and failed. The error complained `[object Object]` was not unique 🤔

PR #1204, opened 2023-01-18
Feat/messages list

pieterlukasse:
Jira Tickets: [VADC-241](https://ctds-planx.atlassian.net/browse/VADC-241),  [VADC-372](https://ctds-planx.atlassian.net/browse/VADC-372)

### New Features
- New messages framework for displaying success and error messages
- New validation + error messages for specific error scenarios

Example:
Screenshot 2023-01-18 at 20 58 05

### Improvements
- Improved variable name in AttritionTable.jsx 

[VADC-241]: https://ctds-planx.atlassian.net/browse/VADC-241?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[VADC-372]: https://ctds-planx.atlassian.net/browse/VADC-372?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Analysis/GWASV2/Components/Covariates/CustomDichotomousCovariates.jsx
+++ b/src/Analysis/GWASV2/Components/Covariates/CustomDichotomousCovariates.jsx
@@ -7,6 +7,7 @@ import '../../../GWASUIApp/GWASUIApp.css';
 
 const CustomDichotomousCovariates = ({
   dispatch,
+  handleSelect,
   handleClose,
   studyPopulationCohort,
   covariates,
@@ -26,7 +27,7 @@ const CustomDichotomousCovariates = ({
       ],
       provided_name: providedName,
     };
-    dispatch(dichotomous);
+    handleSelect(dichotomous);
     handleClose();
   };
 
@@ -92,6 +93,7 @@ const CustomDichotomousCovariates = ({
                 <div>Select your cohorts to assess overlap</div>
               ) : (
                 <CohortsOverlapDiagram
+                  dispatch={dispatch}
                   selectedStudyPopulationCohort={studyPopulationCohort}
                   selectedCaseCohort={firstPopulation}
                   selectedControlCohort={secondPopulation}
@@ -110,6 +112,7 @@ const CustomDichotomousCovariates = ({
 
 CustomDichotomousCovariates.propTypes = {
   dispatch: PropTypes.func.isRequired,
+  handleSelect: PropTypes.func.isRequired,
   studyPopulationCohort: PropTypes.object.isRequired,
   covariates: PropTypes.array,
   outcome: PropTypes.object,
--- a/src/Analysis/GWASV2/Components/Diagrams/CohortsOverlapDiagram/CohortsOverlapDiagram.jsx
+++ b/src/Analysis/GWASV2/Components/Diagrams/CohortsOverlapDiagram/CohortsOverlapDiagram.jsx
@@ -107,31 +107,31 @@ const CohortsOverlapDiagram = ({
   ]);
 
   const {
-    statusPopCase,
-    statusPopControl,
-    statusCaseControl,
-    statusPopCaseControl,
-    dataPopCase,
-    dataPopControl,
-    dataCaseControl,
-    dataPopCaseMinusPopCaseControl,
+    statusStudyPopulationAndCase,
+    statusStudyPopulationAndControl,
+    statusCaseAndControl,
+    statusStudyPopulationAndCaseAndControl,
+    dataStudyPopulationAndCase,
+    dataStudyPopulationAndControl,
+    dataCaseAndControl,
+    dataStudyPopulationAndCaseMinusStudyPopulationAndCaseAndControl,
   } = {
-    statusPopCase: results[0].status,
-    statusPopControl: results[1].status,
-    statusCaseControl: results[2].status,
-    statusPopCaseControl: results[3].status,
-    dataPopCase: results[0].data,
-    dataPopControl: results[1].data,
-    dataCaseControl: results[2].data,
-    dataPopCaseMinusPopCaseControl: results[3].data,
+    statusStudyPopulationAndCase: results[0].status,
+    statusStudyPopulationAndControl: results[1].status,
+    statusCaseAndControl: results[2].status,
+    statusStudyPopulationAndCaseAndControl: results[3].status,
+    dataStudyPopulationAndCase: results[0].data,
+    dataStudyPopulationAndControl: results[1].data,
+    dataCaseAndControl: results[2].data,
+    dataStudyPopulationAndCaseMinusStudyPopulationAndCaseAndControl: results[3].data,
   };
 
   useEffect(() => {
     // Validate and give error message if there is no overlap:
-    if (dataPopCase?.cohort_overlap
-        && dataPopControl?.cohort_overlap) {
-      if (dataPopCase.cohort_overlap.case_control_overlap === 0
-        || dataPopControl.cohort_overlap.case_control_overlap === 0) {
+    if (dataStudyPopulationAndCase?.cohort_overlap
+        && dataStudyPopulationAndControl?.cohort_overlap) {
+      if (dataStudyPopulationAndCase.cohort_overlap.case_control_overlap === 0
+        || dataStudyPopulationAndControl.cohort_overlap.case_control_overlap === 0) {
         dispatch({
           type: ACTIONS.ADD_MESSAGE,
           payload: OVERLAP_ERROR,
@@ -143,24 +143,24 @@ const CohortsOverlapDiagram = ({
         });
       }
     }
-  }, [dataPopCase, dataPopControl]);
+  }, [dataStudyPopulationAndCase, dataStudyPopulationAndControl]);
 
   if (
     [
-      statusPopCase,
-      statusPopControl,
-      statusCaseControl,
-      statusPopCaseControl,
+      statusStudyPopulationAndCase,
+      statusStudyPopulationAndControl,
+      statusCaseAndControl,
+      statusStudyPopulationAndCaseAndControl,
     ].some((status) => status === 'error')
   ) {
     return <React.Fragment>Error getting data for diagram</React.Fragment>;
   }
   if (
     [
-      statusPopCase,
-      statusPopControl,
-      statusCaseControl,
-      statusPopCaseControl,
+      statusStudyPopulationAndCase,
+      statusStudyPopulationAndControl,
+      statusCaseAndControl,
+      statusStudyPopulationAndCaseAndControl,
     ].some((status) => status === 'loading')
   ) {
     return <Spin />;
@@ -169,12 +169,12 @@ const CohortsOverlapDiagram = ({
     set1Size: selectedStudyPopulationCohort.size,
     set2Size: selectedCaseCohort.size,
     set3Size: selectedControlCohort.size,
-    set12Size: dataPopCase.cohort_overlap.case_control_overlap,
-    set13Size: dataPopControl.cohort_overlap.case_control_overlap,
-    set23Size: dataCaseControl.cohort_overlap.case_control_overlap,
+    set12Size: dataStudyPopulationAndCase.cohort_overlap.case_control_overlap,
+    set13Size: dataStudyPopulationAndControl.cohort_overlap.case_control_overlap,
+    set23Size: dataCaseAndControl.cohort_overlap.case_control_overlap,
     set123Size:
-      dataPopCase.cohort_overlap.case_control_overlap
-      - dataPopCaseMinusPopCaseControl.cohort_overlap.case_control_overlap,
+      dataStudyPopulationAndCase.cohort_overlap.case_control_overlap
+      - dataStudyPopulationAndCaseMinusStudyPopulationAndCaseAndControl.cohort_overlap.case_control_overlap,
     set1Label: selectedStudyPopulationCohort.cohort_name,
     set2Label: selectedCaseCohort.cohort_name,
     set3Label: selectedControlCohort.cohort_name,
--- a/src/Analysis/GWASV2/Components/Diagrams/CohortsOverlapDiagram/CohortsOverlapDiagram.jsx
+++ b/src/Analysis/GWASV2/Components/Diagrams/CohortsOverlapDiagram/CohortsOverlapDiagram.jsx
@@ -1,4 +1,4 @@
-import React from 'react';
+import React, { useEffect } from 'react';
 import PropTypes from 'prop-types';
 import { useQueries } from 'react-query';
 import { Spin } from 'antd';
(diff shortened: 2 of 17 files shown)


src/Analysis/GWASV2/Components/Covariates/CustomDichotomousCovariates.jsx
```diff
@@ -7,6 +7,7 @@ import '../../../GWASUIApp/GWASUIApp.css';
 
 const CustomDichotomousCovariates = ({
   dispatch,
+  handleSelect,
```
pieterlukasse, line 10: TODO - update storybook

src/Analysis/GWASV2/Components/Diagrams/CohortsOverlapDiagram/CohortsOverlapDiagram.jsx
```diff
@@ -9,8 +9,14 @@ import {
 } from '../../../Utils/cohortMiddlewareApi';
 import Simple3SetsEulerDiagram from './Simple3SetsEulerDiagram';
 import { useSourceContext } from '../../../Utils/Source';
+import ACTIONS from '../../../Utils/StateManagement/Actions';
 
+const OVERLAP_ERROR = {
```
jarvisraymond-uchicago, line 14: Idea: Do we want to put this and the other messages the application will use into a single message file the same way we do with ACTIONS? That might make it easier to maintain since all the errors would be in one place.

src/Analysis/GWASV2/Components/DismissibleMessage/DismissibleMessage.jsx
```diff
@@ -19,11 +24,9 @@ const DismissibleMessage = ({
             tabIndex='0'
             role='button'
             aria-label='Close Message'
-            onClick={() => {
-              setOpen(false);
-            }}
+            onClick={close}
             onKeyDown={(e) => {
-              if (isEnterOrSpace(e)) setOpen(false);
+              if (isEnterOrSpace(e)) close();
```
jarvisraymond-uchicago, line 29: If the user closes the message manually, do we also want to remove it from the messages list in the global state? Its not immediately clear to me if that is occurring in the close() function all of the time since there's a conditional there.

src/Analysis/GWASV2/Components/DismissibleMessage/DismissibleMessage.jsx
```diff
@@ -7,8 +7,13 @@ const DismissibleMessage = ({
   title = 'Placeholder Title',
   description = 'placeholder description',
   messageType = 'success',
+  dismissMessage = null,
 }) => {
   const [open, setOpen] = useState(true);
+  const close = () => {
+    if (dismissMessage) dismissMessage();
```
jarvisraymond-uchicago, line 14: Personal preference: I recommend always including brackets with if statements. Sometimes errors can arise when other developers add statements after the if statement without them being surrounded by brackets, leading to unexpected errors and execution. 

For example these two conditionals execute foobar() under different circumstances
``` 
if (dismissMessage) 
          dismissMessage();
          foobar();
```
          
          vs
          
```
 if (dismissMessage) { 
          dismissMessage();
          foobar();
       }   
```

jarvisraymond-uchicago, line 14: Since dismissMessage is required, do we need to check for it? I don't think you'd pass prop validation without it (line 52). Does checking for this imply there are going to be error messages that aren't added to the global state?

src/Analysis/GWASV2/GWASContainer.jsx
```diff
@@ -95,6 +96,15 @@ const GWASContainer = () => {
         selectedCohort={state.selectedStudyPopulationCohort}
         outcome={state.outcome}
       />
+      <DismissibleMessagesList
```
jarvisraymond-uchicago, line 99: A drawback of this approach is that the error messages will not render near the input that causes the error. But that might be a reasonable trade off

src/Analysis/GWASV2/Utils/StateManagement/reducer.js
```diff
@@ -45,6 +45,15 @@ const reducer = (state, action) => {
     return { ...state, selectedHare: action.payload };
   case ACTIONS.UPDATE_FINAL_POPULATION_SIZES:
     return { ...state, finalPopulationSizes: action.payload };
+  case ACTIONS.ADD_MESSAGE:
+    return { ...state, messages: [...state.messages, action.payload] };
+  case ACTIONS.DELETE_MESSAGE:
+    return {
+      ...state,
+      messages: state.messages.filter(
```
jarvisraymond-uchicago, line 53: I noticed this approach, like the card approach we have, leads to there being duplicate messages being shown (for example I got this to render pretty quick) but closing one of them closes them all. Do we want to add a conditional to the add message action so duplicate error messages don't display (i.e. only add the message if it is not in state.messages )

image

src/Analysis/GWASV2/Steps/SelectOutcome/SelectOutcome.jsx
```diff
@@ -39,12 +39,13 @@ const SelectOutcome = ({
...
         <div className='select-outcome-container'>
           <CustomDichotomousCovariates
+            dispatch={dispatch}
             studyPopulationCohort={studyPopulationCohort}
             outcome={outcome}
             handleClose={() => {
               setSelectionMode('');
             }}
-            dispatch={(chosenOutcome) => {
+            handleSelect={(chosenOutcome) => {
```
jarvisraymond-uchicago, line 48: Should the error message for Selecting Covariates be identical to the error message for Selecting an Outcome? We might want them to be different so it is more clear to the user what they've done to cause an error

Idea: the error message information could be passed as a prop here instead of it being hardcoded in CustomDichotomousCovariates. Error messages could all live in a single file that could be imported i.e. ErrorMessages.js

src/Analysis/GWASV2/Components/DismissibleMessage/DismissibleMessage.jsx
jarvisraymond-uchicago, line 14: Is this conditional necessary? I think the props validation requires the function should this should always be true (I think? I don't think prop validation would accept a falsey function)

jarvisraymond-uchicago (commented):
Very impressive! 

- Left comments and questions, happy to chat about them over a huddle if you want.

src/Analysis/GWASV2/Components/DismissibleMessage/DismissibleMessage.jsx
```diff
@@ -7,8 +7,13 @@ const DismissibleMessage = ({
   title = 'Placeholder Title',
   description = 'placeholder description',
```
jarvisraymond-uchicago, line 8: I'm wondering if we want the user to be able to always be able to dismiss the messages. We might want to add a prop to the component so the user cannot dismiss the error message manually. Do you think there will be error messages we do not want the user to be able to dismiss (i.e. some critical error?)

src/Analysis/GWASV2/Components/Covariates/CustomDichotomousCovariates.jsx
pieterlukasse: that is more like a note to myself: do it in this PR.

src/Analysis/GWASV2/Components/AttritionTableWrapper/AttritionTable/AttritionTable.jsx
```diff
@@ -12,7 +12,7 @@ const AttritionTable = ({
   const [covariatesProcessed, setCovariatesProcessed] = useState([]);
   // Creates an array of arrays such that given input arr [A,B,C]
   // it returns arr [[A], [A,B], [A,B,C]]
-  const getCovariateRow = (inputArr) => {
+  const getCovariateRows = (inputArr) => {
```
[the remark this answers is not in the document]
pieterlukasse: yeah, now that you mention it, maybe `getCovariatesProcessed` makes more sense given it is used to set `covariatesProcessed`... Or maybe a name that indicates its use...something like `getAttritionLists` 

Further down in the code, each list is mapped to a single `AttritionTableRow`. Anyway, this maybe can move to a separate ticket/discussion. I'll do that and revert this.

src/Analysis/GWASV2/Components/Diagrams/CohortsOverlapDiagram/CohortsOverlapDiagram.jsx
```diff
@@ -120,6 +126,25 @@ const CohortsOverlapDiagram = ({
     dataPopCaseMinusPopCaseControl: results[3].data,
   };
 
+  useEffect(() => {
+    // Validate and give error message if there is no overlap:
+    if (dataPopCase?.cohort_overlap
+        && dataPopControl?.cohort_overlap) {
+      if (dataPopCase.cohort_overlap.case_control_overlap === 0
```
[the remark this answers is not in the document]
pieterlukasse: good suggestion. I can rename.

src/Analysis/GWASV2/Components/DismissibleMessage/DismissibleMessage.jsx
pieterlukasse: I'm suspecting this was done by the linter...let me check

src/Analysis/GWASV2/Utils/StateManagement/reducer.js
pieterlukasse: > add a conditional to the add message action so duplicate error messages don't display

I think that is feasible 👍

src/Analysis/GWASV2/Steps/SelectOutcome/SelectOutcome.jsx
pieterlukasse: > If the user creates an error in SelectOutcome and then also in Select Covariates, then the user fixes the error in Select Covariates - i think the error for the SelectOutcome would be dismissed when it shouldn't be

Good point. I'll make the messages unique.

src/Analysis/GWASV2/Components/DismissibleMessage/DismissibleMessage.jsx
pieterlukasse: I thought about this scenario...We have only one such moment at the end of the flow, where we should not allow the user to trigger a workflow for a population of size == 0. I'll work on something for that...not sure yet what would be the best approach.

src/Analysis/GWASV2/Components/AttritionTableWrapper/AttritionTable/AttritionTable.jsx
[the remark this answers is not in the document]
pieterlukasse: https://ctds-planx.atlassian.net/browse/VADC-406

src/Analysis/GWASV2/Components/Diagrams/CohortsOverlapDiagram/CohortsOverlapDiagram.jsx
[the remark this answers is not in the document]
pieterlukasse: done. The names are longer now, but perhaps clearer.

src/Analysis/GWASV2/Components/Diagrams/CohortsOverlapDiagram/CohortsOverlapDiagram.jsx
pieterlukasse: done. Moved to the existing `constants.js`

src/Analysis/GWASV2/Components/DismissibleMessage/DismissibleMessage.jsx
pieterlukasse: if it is not removed from the global state, it will stay on the screen, since we have this in our app:
```js
      <DismissibleMessagesList
        messages={state.messages}
        dismissMessage={(chosenMessage) => {
          dispatch({
            type: ACTIONS.DELETE_MESSAGE,
            payload: chosenMessage,
          });
        }}
      />
```
i.e. the messages are always rendered based on `state.messages`

src/Analysis/GWASV2/Components/DismissibleMessage/DismissibleMessage.jsx
pieterlukasse: Done. Using the format you proposed.

pieterlukasse: It is not required. Default is `null`.

src/Analysis/GWASV2/Utils/StateManagement/reducer.js
pieterlukasse: done.

src/Analysis/GWASV2/Steps/SelectOutcome/SelectOutcome.jsx
pieterlukasse: As discussed, this will take some more work, given the components are reused in different steps, while they're not aware of which step they are in. I'll create a separate ticket for refining/improving the messages.

pieterlukasse: https://ctds-planx.atlassian.net/browse/VADC-413

src/Analysis/GWASV2/Components/DismissibleMessage/DismissibleMessage.jsx
pieterlukasse: Remaining population of size==0 will cause HARE dropdown to be empty, which will disable the "Next" button. I think that for now this is a good solution, but a better/clearer validation could be added in the future. This current one might also have some cases where it fails to disable the "Next" button, as I have not tested all scenarios:

Screenshot 2023-01-26 at 18 25 36

src/Analysis/GWASV2/Components/Covariates/CustomDichotomousCovariates.jsx
pieterlukasse: moving to a ticket

pieterlukasse: https://ctds-planx.atlassian.net/browse/VADC-414

jarvisraymond-uchicago (approved):
Looks good!

Per discussion, @pieterlukasse  to update error message for when only one of the selections does not overlap for Dichotomous selection.

PR #1228, opened 2023-02-07
Fix: make sure hare concept id is part of histogram and overlap queries

pieterlukasse:
Jira Ticket: [VADC-425](https://ctds-planx.atlassian.net/browse/VADC-425)

### Bug Fixes
-  fix to make sure hare concept id is part of histogram and overlap queries

### Improvements
- also cleaned up some dead code

[VADC-425]: https://ctds-planx.atlassian.net/browse/VADC-425?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Analysis/GWASV2/Utils/cohortMiddlewareApi.js
+++ b/src/Analysis/GWASV2/Utils/cohortMiddlewareApi.js
@@ -18,7 +18,8 @@ export const fetchSimpleOverlapInfo = async (
       {
         variable_type: 'concept',
         concept_id: hareConceptId,
-      }],
+      }].filter(Boolean), // filter out any undefined or null items (e.g. in some
+                          // scenarios "outcome" and "selectedCovariates" are still null and/or empty)
   };
   const statsEndPoint = `${cohortMiddlewarePath}cohort-stats/check-overlap/by-source-id/${sourceId}/by-cohort-definition-ids/${cohortAId}/${cohortBId}`;
   const reqBody = {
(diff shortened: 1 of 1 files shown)


src/Analysis/GWASV2/Utils/cohortMiddlewareApi.js
```diff
@@ -76,7 +13,12 @@ export const fetchSimpleOverlapInfo = async (
   outcome,
 ) => {
   const variablesPayload = {
-    variables: [...selectedCovariates, outcome],
+    variables: [...selectedCovariates, outcome,
```
pieterlukasse, line 16: TODO: this is a bit fragile code in case `outcome` is null. I believe it only happens in storybooks now, but more robust logic here would be nice.

pieterlukasse: Not introduced by this PR though...maybe can be addressed in a separate ticket.

jarvisraymond-uchicago (commented):
I believe the platform's guidelines state that PR's should be named using the type of pull request followed by a forward slash separator. It looks like this PR is using an underscore. I think the platform's documentation (https://gen3.org/resources/developer/dev-introduction/) states the branch name should be: 

```fix/align_breakdown_and_other_queries

instead of 

```fix_align_breakdown_and_other_queries

I don't know if this is important or if it matters. 

#### Screen grab from the documentation:
<img width="843" alt="image" src="https://user-images.githubusercontent.com/113449836/217558847-91cc3b65-d74b-4f74-9ba8-63f6281c6ca2.png">

src/Analysis/GWASV2/Utils/cohortMiddlewareApi.js
pieterlukasse: solved.

jarvisraymond-uchicago (approved):
Looks good, nice that we're able to remove that much unneeded code. I pulled the changes local and verified I could create both types of diagrams and submit a workflow.

PR #1231, opened 2023-02-09
Feat/cleanup old GWAS UI app code

pieterlukasse:
Jira Ticket: [VADC-399](https://ctds-planx.atlassian.net/browse/VADC-399)

### Improvements
- cleanup of OLD GWAS UI app code, keeping only V2 of the app and the Results viewer app

### Deployment changes
- this will require some changes in other deployment related config files. See:
   -  https://github.com/uc-cdis/gitops-qa/pull/2362
   - https://github.com/uc-cdis/cdis-manifest/pull/5713

[VADC-399]: https://ctds-planx.atlassian.net/browse/VADC-399?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


.github/linters/.eslintrc.json
```diff
@@ -225,21 +225,6 @@
                 "react/no-unused-prop-types": "off"
             }
         },
-        {
```
jarvisraymond-uchicago, line 228: #### Comment, no changes needed: 
This was probably a bad approach to begin with. Better would have been to ignore the line or the entire file rather than adding a new global rule for one file.

jarvisraymond-uchicago (approved):
Great work Pieter! I pulled the changes locally and confirmed only one GWAS app remains without causing any obvious errors. 

Before merging consider
* Do we want to rename the folder for the application from ```GWASV2``` to ```GWAS``` now that there's only one GWAS left?
* For tickets like this: https://ctds-planx.atlassian.net/browse/VADC-323 should the developer reference this PR to find all the removed storybooks?
* Are there follow up tickets to implement the needed ```changes in other deployment related config files```?

pieterlukasse:
> Do we want to rename the folder for the application from GWASV2 to GWAS now that there's only one GWAS left?

Good point, although it is probably best to think of a more unique name, maybe GWASApp. Also, with "GWASV2" string having 162 hits in 37 files, this might be a bit large of a change to include here on top of all the other changes. I would prefer to do this in a separate action. There are some more cleanup actions that need to take place, e.g. the `src/Analysis/GWASResults/GWASUIApp.css` is probably way too large (a copy of the legacy versions) and is also referenced in a number of places where it really should not be.

> Are there follow up tickets to implement the needed changes in other deployment related config files?

I've update the description with links to the other related PRs

PR #1234, opened 2023-02-10
fix: proper spinner for manifest count

m0nhawk:
Jira Ticket: N/A

### Bug Fixes
* fix for spinner for manifest button not showing until the count

### DIFF
--- a/src/GuppyDataExplorer/ExplorerButtonGroup/index.jsx
+++ b/src/GuppyDataExplorer/ExplorerButtonGroup/index.jsx
@@ -687,6 +687,9 @@ Currently, in order to export a File PFB, \`enableLimitedFilePFBExport\` must be
       } else {
         // otherwise, just query subject index for subject_id list,
         // and query file index for manifest info.
+        // Setting downloadingInProgress of manifest to true as a trick such that the button's isPending
+        // state is set to true while the manifest is being counted. This is because we want the button to show
+        // the spinner instead of the download icon while the counting is in progress. 
         this.setState({
           manifestEntryCount: 0, downloadingInProgress: { manifest: true }, toasterOpen: false,
         });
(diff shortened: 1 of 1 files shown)


src/GuppyDataExplorer/ExplorerButtonGroup/index.jsx
```diff
@@ -688,7 +688,7 @@ Currently, in order to export a File PFB, \`enableLimitedFilePFBExport\` must be
         // otherwise, just query subject index for subject_id list,
         // and query file index for manifest info.
         this.setState({
-          manifestEntryCount: 0,
+          manifestEntryCount: 0, downloadingInProgress: { manifest: true },
```
mfshao, line 691:
```suggestion
          manifestEntryCount: 0, downloadingInProgress: { manifest: true }, toasterOpen: false,
```
these fixes LGTM, I might just add above changes to ensure the toaster stay closed in these opeations, although I don;t think it will show up, so not a mandatory change

src/GuppyDataExplorer/ExplorerButtonGroup/index.jsx
```diff
@@ -688,7 +688,7 @@ Currently, in order to export a File PFB, \`enableLimitedFilePFBExport\` must be
         // otherwise, just query subject index for subject_id list,
         // and query file index for manifest info.
```
frankliuao, line 689:
```suggestion
        // and query file index for manifest info.
        // Setting downloadingInProgress of manifest to true as a trick such that the button's isPending
        // state is set to true while the manifest is being counted. This is because we want the button to show
        // the spinner instead of the download icon while the counting is in progress. 
```

frankliuao: Added some comments to make it clear for people who work on this in the future.

frankliuao (changes requested):
Suggested a few changes.

frankliuao (approved):
This is to add a trick to let the spinner run while the manifest count is progress. Needed for MIDRC urgently. Approved.

mfshao (approved):

mfshao:
CI is failing on ETL tests and I don't think this PR is the cause, merging

PR #1245, opened 2023-03-02
refactor/lintCleanup: fix eslint errors

ocshawn:
### Improvements
- fixing lint errors to try and get all tests to pass


ocshawn:
only change i made was in .github/linters/.eslintrc.json (made it match the main lint file which seems to have fixed the superlint errors) the rest was just fixing spacing and other lint stuff

mfshao:
Linting is ok, but I don't think this will really fix the superlinter acttion because they only run on files that has been changed for a commit
I think your last superlint GH action run passes because it only runs on the eslintrc config file...

jarvisraymond-uchicago:
@ocshawn : I can take at this, but could you please target merging into master? feat/vadc_sprint04 should probably limited to VADC related work, it is also already in PR and being reviewed. 

Also: could you include the lint errors for stylelint? There's a LESS file ```../../Covid19Dashboard/Covid19Dashboard.less``` that triggers the linter everytime I run the ```stylelint``` command.

src/actions.js
```diff
@@ -381,6 +381,7 @@ export const updateSystemUseNotice = (displayUseWarning) => (dispatch) => {
   });
 };
 
+// eslint-disable-next-line max-len
```
jarvisraymond-uchicago, line 384: Instead of adding a comment to disable the linter here, is there a reason for not adding line breaks to the function? Would something like this work? 

```
export const displaySystemUseNotice = (authenticated) => (dispatch, getState) =>
  dispatch(
    checkIfDisplaySystemUseNotice(
      authenticated,
      getState().popups.systemUseWarnPopup
    )
  );


```

ocshawn:
fixed issues mentioned above.

@mfshao i added the dev dependency @babel/plugin-proposal-class-properties because i saw main .eslintrc.js file uses it, is this needed? 

i also removed the changes i made to the superlint .eslintrc.js file and its appearing to pass now, any thoughts?

ocshawn:
will retry in a new pr

PR #1250, opened 2023-03-08
HP-1063 Fix/popup

mfshao:
Jira Ticket: [HP-1063](https://ctds-planx.atlassian.net/browse/HP-1063)

### Bug Fixes
- Fixed a popup height issue
- Fixed an issue that some urls may have double slash in them

[HP-1063]: https://ctds-planx.atlassian.net/browse/HP-1063?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/components/Popup.less
```diff
@@ -27,7 +27,7 @@
 
 .popup__box {
   width: 70%;
-  height: 100%;
```
mfshao, line 30: I think this should do the trick @craigrbarnes 🤔

src/components/Popup.less
```diff
@@ -27,7 +27,7 @@
...
+  max-height: 100%;
```
mfshao, line 30: I think this should do the trick @craigrbarnes 🤔

src/actions.js
```diff
@@ -381,6 +381,7 @@ export const updateSystemUseNotice = (displayUseWarning) => (dispatch) => {
   });
 };
 
+// eslint-disable-next-line max-len
```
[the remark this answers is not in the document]
mfshao: I think the `getState().popups.systemUseWarnPopup` is not being used, so should be ok to remove it to make this even shorter

jarvisraymond-uchicago (approved):
Looks good to me!

PR #1289, opened 2023-04-27
feat(discovery): CSS update to handle newlines better in Discovery modal

Avantol13:
### Improvements
- CSS update to handle newlines better in Discovery modal

### DIFF
--- a/src/Discovery/Discovery.css
+++ b/src/Discovery/Discovery.css
@@ -569,6 +569,7 @@
 
 .discovery-modal__field {
   display: flex;
+  white-space: pre-wrap;
   justify-content: space-between;
   padding: 2px 0;
 }


src/Discovery/Discovery.css
```diff
@@ -569,6 +569,7 @@
 
 .discovery-modal__field {
   display: flex;
+  white-space: pre-wrap;
```
Avantol13, line 572: pre-wrap "Sequences of white space are preserved. Lines are broken at newline characters, at [<br>](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/br), and as necessary to fill line boxes." https://developer.mozilla.org/en-US/docs/Web/CSS/white-space

mfshao (approved):

PR #1294, opened 2023-04-28
Feat/results home table filtering

jarvisraymond-uchicago:
Jira Ticket: [VADC-504](https://ctds-planx.atlassian.net/browse/VADC-504)

### New Features
This adds filtering to the home table in Results App V2 for the first five columns and allows for the user to filter by text input, dropdown selection, and date range selection. The states for the filtering and sorting are maintained in a ```homeTableState``` so that the user selections are maintained while they navigate within the application and during data refetches for the home table. Note that to maintain a consistent UX and avoid problems that could arise after the implementation of the column management UI, sorting and filtering resets the user's current page in the table back to page one. 

### Improvements
Adds ignore for *.conf files in eslint-new command. 

### Design
image

#### In storybook

output

#### In Local QA: 

output

[VADC-504]: https://ctds-planx.atlassian.net/browse/VADC-504?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/Analysis/GWASResultsV2/Views/Home/Home.jsx
```diff
@@ -13,35 +13,26 @@ const Home = () => {
     const getWorkflows = await fetch(workflowsEndpoint);
     return getWorkflows.json();
   }
-  const GWASWorkflows = () => {
```
jarvisraymond-uchicago, line 16: It turns out if you use a function here to output the needed JSX, it causes a re-render when the states associated with the homeTable inputs change, which causes them to lose focus. Placing the function inline fixes that problem.

src/Analysis/GWASResultsV2/Views/Home/Home.test.jsx
jarvisraymond-uchicago, line 1: Ticket to update tests to test for new functionality: https://ctds-planx.atlassian.net/browse/VADC-554

jarvisraymond-uchicago, line 1: Ticket to update tests to test for new functionality: https://ctds-planx.atlassian.net/browse/VADC-554

src/Analysis/GWASResultsV2/Views/Home/HomeTable/HomeTable.css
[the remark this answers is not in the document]
pieterlukasse: The actual app shows better results, with the date components not quickly affected. However, the job status column has another side-effect when selecting multiple options: the column grows in width at each added option, eventually affecting the date columns depending on their width:

Screenshot 2023-05-03 at 17 00 30

Somehow I was also able to get a wrong result after filtering:

Screenshot 2023-05-03 at 16 59 42

This seems to be related to the warning I get in Console:
```
Warning: Encountered two children with the same key, `gwas-workflow-0137223645`.
Keys should be unique ...
```
I think this could be related to some tests I've executed using the retry option, which is a feature we'll be supporting soon. To solve this you should use the `uid` fields as the key instead of `workflow_name`

src/Analysis/GWASResultsV2/Views/Home/HomeTable/HomeTable.jsx
```diff
@@ -1,94 +1,295 @@
...
 const HomeTable = ({ data }) => {
-  const { setCurrentView, setSelectedRowData } = useContext(SharedContext);
+  const {
+    setCurrentView,
+    setSelectedRowData,
+    homeTableState,
+    setHomeTableState,
+  } = useContext(SharedContext);
+
+  const handleTableChange = (pagination, filters, sorter) => {
```
pieterlukasse, line 27: `filters` not used here. Remove?

src/Analysis/GWASResultsV2/Views/Home/HomeTable/HomeTable.jsx
```diff
@@ -1,94 +1,295 @@
...
+        ...homeTableState,
+        currentPage: pagination.current,
+      });
+    }
+    // When the user updates sorting or filtering, set page to first page
+    return setHomeTableState({
+      ...homeTableState,
+      currentPage: 1,
+      sortInfo: sorter,
+    });
```
pieterlukasse, line 40: question: if this already takes care of setting to page 1, why is it repeated in the handlers below?

pieterlukasse (commented):
nice PR, great features 👍 

@jarvisraymond-uchicago thanks for the PR, it looks very nice and the end result is a very intuitive UI. I did find some small issues and had some questions in some parts of the code. Please take a look.

src/Analysis/GWASResultsV2/Views/Home/HomeTable/filterTableData.js
```diff
@@ -0,0 +1,58 @@
+import moment from 'moment';
+
+const filterBySearchTerm = (initData, key, searchTerm) => initData.filter((obj) => obj[key]
```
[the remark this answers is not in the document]
jarvisraymond-uchicago: Good idea. This has been revised in each of the filtering methods to use ```data``` as a variable name instead of ```initData```

```

const filterBySearchTerm = (data, key, searchTerm) => data.filter((obj) => obj[key]
  .toString()
  .toLowerCase()
  .includes(searchTerm.toLowerCase()),
);

const filterByJobStatuses = (data, jobStatusSelections) => data.filter((item) => jobStatusSelections.includes(item.phase));

const filterByDateRange = (data, key, dateSelection) => data.filter((obj) => {
  const utcDate = moment.utc(obj[key]);
  return (
    utcDate.isSameOrAfter(dateSelection[0])
      && utcDate.isSameOrBefore(dateSelection[1])
  );
});

const filterTableData = (data, homeTableState) => {
  let filteredDataResult = data;
  if (homeTableState.nameSearchTerm.length > 0) {
    filteredDataResult = filterBySearchTerm(
      filteredDataResult,
      'name',
      homeTableState.nameSearchTerm,
    );
  }

```

src/Analysis/GWASResultsV2/Views/Home/HomeTable/HomeTable.css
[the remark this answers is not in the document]
jarvisraymond-uchicago: > I noticed some minor styling "issues" where options and dates are not completely readable when the screen is a bit smaller. This is how it looks in Storybook on a maximized window on my laptop:
> 
> Screenshot 2023-05-03 at 16 50 40
> 
> Maybe the status and date fields can have a minimal width and the other fields can be flexible width?

I've updated the select job status column to have a hard coded width. I think this should solve the problems with both smaller screens and the user adding many statuses: 
image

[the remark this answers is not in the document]
jarvisraymond-uchicago: > 



> The actual app shows better results, with the date components not quickly affected. However, the job status column has another side-effect when selecting multiple options: the column grows in width at each added option, eventually affecting the date columns depending on their width:
> 
> Screenshot 2023-05-03 at 17 00 30
> 
> Somehow I was also able to get a wrong result after filtering:
> 
> Screenshot 2023-05-03 at 16 59 42
> 
> This seems to be related to the warning I get in Console:
> 
> ```
> Warning: Encountered two children with the same key, `gwas-workflow-0137223645`.
> Keys should be unique ...
> ```
> 
> I think this could be related to some tests I've executed using the retry option, which is a feature we'll be supporting soon. To solve this you should use the `uid` fields as the key instead of `workflow_name`

**Response**: I've updated the keys for the workflow name to use the uid. 

Line 289: ```rowKey={(record) => record.uid}```
Line 112: ```key: 'uid',```

src/Analysis/GWASResultsV2/Views/Home/HomeTable/HomeTable.jsx
jarvisraymond-uchicago: No. This is a sent as part of the event, removing breaks sorting because then the sorter object wouldn't be in the right order. It could be renamed if desired to ``` _ ``` for example, but I think it makes sense to keep it here so we know what's being sent to the function

src/Analysis/GWASResultsV2/Utils/isIterable.js
jarvisraymond-uchicago, line 1: I noticed sometimes the server is responding with a non-iterable object and this crashes the application. I've added a check for this now.

src/Analysis/GWASResultsV2/Views/Home/HomeTable/HomeTable.jsx
pieterlukasse: ok, got it

pieterlukasse (approved):
issues are resolved

PR #1306, opened 2023-05-17
Feat/input details view

jarvisraymond-uchicago:
Jira Ticket: [VADC-557](https://ctds-planx.atlassian.net/browse/VADC-557)

### New Features
This adds the input view to the Results App including the Job Details table and a button to access the input view from the results app home table. Note that in the input view the attrition table and the final size calculation for the Job Details table will be done as part of subsequent tickets. The minimum and maximum outlier rows are no longer scoped.
### Design

image

#### App
image
image

#### Storybook
image

image
image
image

[VADC-557]: https://ctds-planx.atlassian.net/browse/VADC-557?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/Analysis/GWASResults/TestData/InputViewData/MockedFailureJSON.js
pieterlukasse, line 1: The content of this file looks like a successful reply. Why is it named "FailureJSON"?

pieterlukasse: Also: do we really need all these details in the file or is there a small subset of fields that suffices? 
This comment also applies to the next file src/Analysis/GWASResults/TestData/InputViewData/MockedSuccessJSON.js

src/Analysis/GWASResults/Views/Input/JobDetails.jsx
```diff
@@ -0,0 +1,128 @@
...
+    if (
+      getParameterData('outcome')
+      && IsJsonString(getParameterData('outcome'))
+    ) {
+      return (
+        JSON.parse(getParameterData('outcome'))?.concept_name
+        || JSON.parse(getParameterData('outcome'))?.provided_name
+      );
+    }
+    return 'Data not found';
```
pieterlukasse, line 51: Maybe say 'Data not found or not in expected JSON format'?

pieterlukasse: I think it should be indicated as an error...maybe even raise Error instead of returning a string here.

src/Analysis/GWASResults/Views/Input/Input.stories.jsx
```diff
@@ -0,0 +1,128 @@
...
+        value={{
+          selectedRowData: selectedRowData,
+          setCurrentView,
+        }}
+      >
+        <Input />
+      </SharedContext.Provider>
+    </QueryClientProvider>
+  );
+};
```
pieterlukasse, line 65: can we keep just one template?

src/Analysis/GWASResults/Views/Input/Input.stories.jsx
```diff
@@ -0,0 +1,128 @@
...
+MockedErrorObject.parameters = {
+  msw: {
+    handlers: [
+      rest.get(
+        `http://:argowrapperpath/ga4gh/wes/v2/status/${name}?uid=${uid}`,
+        (req, res, ctx) => {
+          const { argowrapperpath } = req.params;
+          return res(
+            ctx.delay(100),
+            // Some errroneous responses can return an error object
```
pieterlukasse, line 107: hm...then this is a bug in argo-wrapper. I would propose fixing that instead.

src/Analysis/GWASResults/Views/Input/JobDetails.jsx
```diff
@@ -0,0 +1,128 @@
...
+  };
+
+  const processCovariates = () => {
+    const input = JSON.stringify(getParameterData('variables'));
+    let covariatesString = input.replaceAll('\\n', '');
+    covariatesString = covariatesString.substring(
+      1,
+      covariatesString.length - 1,
+    );
+    const strToRemove = '\\"';
```
pieterlukasse, line 61: why is this needed? Is our argo workflow code or argo-wrapper doing something strange with this data? I'm wondering if we can try to fix that in argo-wrapper itself.

pieterlukasse (commented):
thanks @jarvisraymond-uchicago !

I added some questions for parts that can perhaps be simplified and some ideas for possible tickets/issues we could raise for argo-wrapper

src/Analysis/GWASResults/TestData/InputViewData/MockedFailureJSON.js
jarvisraymond-uchicago: It is named failureJSON because some of the fields are assigned null values. 

The file contains a complete server response. I'm not sure what the advantage would be in mocking a response with fewer fields.

PR #1307, opened 2023-05-19
remove covid 19 dashboard standalone client

ocshawn:
### New Features
- remove covid 19 dashboard standalone client


ocshawn:
this code is still being used by occ devs, will leave it in

PR #1330, opened 2023-06-07
(HP-1157): fix bug where filters dont reset table to page one

ocshawn:
Jira Ticket: [HP-1157](https://ctds-planx.atlassian.net/browse/HP-1157)
### Bug Fixes
- fix bug where filters dont reset table to page one

[HP-1157]: https://ctds-planx.atlassian.net/browse/HP-1157?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Discovery/Discovery.tsx
+++ b/src/Discovery/Discovery.tsx
@@ -204,6 +204,7 @@ export interface Props {
     [accessLevel: number]: boolean
   },
   accessSortDirection: AccessSortDirection,
+  onAdvancedSearch: (advancedSearch: any[]) => any,
   onSearchChange: (arg0: string) => any,
   onTagsSelected: (arg0: any) => any,
   onAccessFilterSet: (arg0: object) => any,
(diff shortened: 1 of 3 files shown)


paulineribeyre (approved):

src/Discovery/Discovery.tsx
```diff
@@ -745,7 +746,10 @@ const Discovery: React.FunctionComponent<Props> = (props: Props) => {
                   config={props.config}
                   studies={props.studies}
                   filterState={filterState}
-                  setFilterState={setFilterState}
+                  setFilterState={(event) => {
+                    props.onAdvancedSearch(event);
```
mfshao, line 750: better to only trigger this if the pagniation page is not 1?

ocshawn: i dont think the extra code is needed (as the state would not change if already on page 1), would like to see setFilterState using the existing reducers but dont think a refactor is worth our time

PR #1350, opened 2023-07-03
Feat/two attrition tables

jarvisraymond-uchicago:
Jira Ticket: [VADC-613](https://ctds-planx.atlassian.net/browse/VADC-613)

### New Features
This refactors the Attrition Table component in the Results App so that the input view renders two tables, one above the other, when the returned JSON data contains objects for both the "table_type":case  and "table_type": control. The case table renders first, above the control table. If the case table data is missing an error message will be shown. The control table is optional and depends on the user's input when creating their GWAS submission. 

The appearance and labels matches the existing Attrition tables in the GWAS App, except for the “Chart” column (which is only shown in GWAS App)

### Design
image

### Implementation
output

[VADC-613]: https://ctds-planx.atlassian.net/browse/VADC-613?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Analysis/GWASResults/TestData/InputViewData/AttritionTableJSON.js
+++ b/src/Analysis/GWASResults/TestData/InputViewData/AttritionTableJSON.js
@@ -96,6 +96,57 @@ const AttritionTableJSON = [
       },
     ],
   },
+  {
+    "table_type": "control",
+    "rows": [
+      {
+        "type": "cohort",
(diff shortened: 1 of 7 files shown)


pieterlukasse (approved):
looks good and works well 👍

PR #1367, opened 2023-07-27
Fix/results app home table fix

jarvisraymond-uchicago:
Jira Ticket: [VADC-632](https://ctds-planx.atlassian.net/browse/VADC-632)

### Bug Fixes
This ticket addresses 3 bug fixes related to the home table date sorting: 

1.) When a user reran a workflow, and then submitted a new workflow, in some instances the order of the table would show them out of order in terms of date finished. This is addressed now by providing the data to the table with an initial sort based on the date finished before the data is rendered. 
2.) When a user reran a workflow and tried to sort based on date finished, the application would crash because the date finished value was null and the sorting function didn't handle null values. This is addressed by adding logic to sort null values as coming before anything else. 
3.) When a user tried to sort a date range consisting of only one date, in some cases the selected date would not render any results even if there were workflows that were submitting or finished on that day. This was due to the date range filtering function being too granular and including filtering based on time of day. The filtering function has been made less granular and now only accounts for the date, and not the time of day. 

Note that the functions associated with the home table data processing have been moved into a separate folder and can all be imported using tableDataProcessing.js
#### Bugs 1 & 2: 
output
#### Bugs 3: 
output

[VADC-632]: https://ctds-planx.atlassian.net/browse/VADC-632?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Analysis/GWASResults/Views/Home/HomeTable/HomeTable.jsx
+++ b/src/Analysis/GWASResults/Views/Home/HomeTable/HomeTable.jsx
@@ -1,7 +1,5 @@
 import React, { useContext, useEffect, useState } from 'react';
-import {
-  Button, Table, Space, Input, DatePicker, Select,
-} from 'antd';
+import { Button, Table, Space, Input, DatePicker, Select } from 'antd';
 import { SearchOutlined } from '@ant-design/icons';
 import PropTypes from 'prop-types';
 import moment from 'moment';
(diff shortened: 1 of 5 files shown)


vzpgb (approved):
Looks nice, thank you!

PR #1370, opened 2023-07-27
cleaning up package and package-lock

ocshawn:
### Improvements
- cleaning up package and package-lock


ocshawn:
outdated will start again

PR #1374, opened 2023-08-02
Fix: Portal updates to match new hatchery changes

nss10:
Link to Jira ticket : [BRH-627](https://ctds-planx.atlassian.net/browse/BRH-627)

## Improvements
* Paymodel resets as soon as a workspace session is terminated
* Workspace launch is restricted when current paymodel is not set for commons where pay models are available. 

[BRH-627]: https://ctds-planx.atlassian.net/browse/BRH-627?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/Workspace/index.jsx
```diff
@@ -328,6 +329,7 @@ class Workspace extends React.Component {
         path: `${workspaceTerminateUrl}`,
         method: 'POST',
       }).then(() => {
+        this.state.initiatePayModelCallUponTermination = true;
```
mfshao, line 332: minor: i don't think this has to be a state variable, should be able to make it as an argument that passes into `checkWorkspaceStatus()`
this can save some re-render cycles caused by unnecessary state updates

src/Workspace/index.jsx
```diff
@@ -370,6 +372,14 @@ class Workspace extends React.Component {
         if (this.workspaceStates.includes(data.status)) {
           const workspaceLaunchStepsConfig = this.getWorkspaceLaunchSteps(data);
           let workspaceStatus = data.status;
+          if (this.state.initiatePayModelCallUponTermination && workspaceStatus === 'Not Found') {
+            this.state.initiatePayModelCallUponTermination = false;
+            this.getWorkspacePayModel().then((payModelData) => {
+              this.setState({
+                payModel: payModelData,
```
mfshao, line 379: just curious when resetting paymodels, this will be reset to vwhat value? `{}` or something else? sry I didn't check the hatchery codes

src/Workspace/index.jsx
nss10: I think this makes sense. We can avoid unnecessary state updates if we make it as an argument to `checkWorkspaceStatus()`. I will work on that.

nss10: I've made the proposed changes and pushed a new commit. Please review and let me know if everything looks good. @mfshao

src/Workspace/index.jsx
```diff
@@ -370,6 +371,13 @@ class Workspace extends React.Component {
         if (this.workspaceStates.includes(data.status)) {
           const workspaceLaunchStepsConfig = this.getWorkspaceLaunchSteps(data);
           let workspaceStatus = data.status;
+          if (args?.triggerPayModelCall && workspaceStatus === 'Not Found') {
```
mfshao, line 374: minor: might be better to read and reduce complexityu a bit by just passing in a boolean arg called `triggerPayModelCall` instead of expecting a generic `arg` object and parse it, unless there is a specific reason to do so?

mfshao (commented):
2 minor comments, otherwise lgtm

src/Workspace/index.jsx
nss10: I thought this could be better to extend the functionality in the future with optional params.

src/Workspace/index.jsx
```diff
@@ -59,6 +59,7 @@ class Workspace extends React.Component {
       workspaceID: null,
       hasWorkspaceAccess: true,
       workspaceIsFullpage: false,
+      initiatePayModelCallUponTermination: false,
```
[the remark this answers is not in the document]
nss10: Thanks! I will remove this one and make a fresh commit

mfshao (approved):

PR #1380, opened 2023-08-09
PXP-11040: add more options to Discovery page isAuthorized check

ocshawn:
Jira Ticket: [PXP-11040](https://ctds-planx.atlassian.net/browse/PXP-11040)
### Improvements
- add more options to Discovery page isAuthorized check 

[PXP-11040]: https://ctds-planx.atlassian.net/browse/PXP-11040?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


paulineribeyre (approved):
lgtm. @mfshao could you have a look and see if this makes sense to you? Explanation for this change [here](https://ctds-planx.atlassian.net/browse/PXP-11040)

mfshao:
I think for now most of the envs are expecting this check to actually represents `“method read-storage” + “service fence”` , but this solution looks like a ok one to me. Maybe in the future we will have to make it configurable so each env can decided how to display authz check info based on their own needs

PR #1397, opened 2023-08-22
feat(blockSubmissionGroupSizeZero): initial commit

jarvisraymond-uchicago:
Jira Ticket: [VADC-672](https://ctds-planx.atlassian.net/browse/VADC-672)

### New Features
Updates GWAS App so when user has a finalPopulationSize containing an object with a size value of zero, it renders a warning message and prevents the user from advancing to the next step (viewing the modal). 

Note that the finalPopulationSize array contains objects for one or more of the following: case, control and the total based on the selected HARE Ancestry. 

### Implementation
image

[VADC-672]: https://ctds-planx.atlassian.net/browse/VADC-672?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Analysis/GWASApp/GWASContainer.jsx
+++ b/src/Analysis/GWASApp/GWASContainer.jsx
@@ -19,76 +19,76 @@ const GWASContainer = () => {
 
   const generateStep = () => {
     switch (state.currentStep) {
-      case 0:
-        return (
-          <div data-tour='cohort-intro'>
-            <SelectStudyPopulation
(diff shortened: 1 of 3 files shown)


pieterlukasse (approved):
code looks good and works well. 
I found a separate issue where the "Next" button does stay enabled. This is likely  a pre-existing bug, not introduced in this PR, so can be fixed in a separate ticket.

PR #1398, opened 2023-08-23
(PXP-7377): add banner to portal

ocshawn:
Jira Ticket: [PXP-7377](https://ctds-planx.atlassian.net/browse/PXP-7377)
### New Features
- add banner to portal

Screenshot from 2023-08-24 12-09-57

[PXP-7377]: https://ctds-planx.atlassian.net/browse/PXP-7377?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Layout/reducers.js
+++ b/src/Layout/reducers.js
@@ -1,4 +1,4 @@
-const bar = (state = {}, action) => {
+export const bar = (state = {}, action) => {
   switch (action.type) {
   case 'ACTIVE_INIT': {
     return { ...state, active: window.location.pathname };
@@ -11,4 +11,23 @@ const bar = (state = {}, action) => {
   }
 };
 
-export default bar;
+export const banner = (state = {}, action) => {
+  switch (action.type) {
+  case 'RESET_BANNER': {
+    const closedBanners = state?.closedBanners?.[action.data.id];
+    delete closedBanners?.[action.data.id];
+    return { ...state, closedBanners };
+  }
+  case 'CLOSE_BANNER': {
+    return {
+      ...state,
+      closedBanners: {
+        ...state?.closedBanners,
+        [action.data.id]: action.data.resetDate,
+      },
+    };
+  }
+  default:
+    return state;
+  }
+};
--- a/src/Layout/reducers.js
+++ b/src/Layout/reducers.js
@@ -14,7 +14,7 @@ export const bar = (state = {}, action) => {
 export const banner = (state = {}, action) => {
   switch (action.type) {
   case 'RESET_BANNER': {
-    const closedBanners = state?.closedBanners?.[action.data.id];
+    const closedBanners = state?.closedBanners;
     delete closedBanners?.[action.data.id];
     return { ...state, closedBanners };
   }
--- a/src/components/Banner.jsx
+++ b/src/components/Banner.jsx
@@ -1,0 +1,52 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { Alert } from 'antd';
+
+class Banner extends React.Component {
+  /**
+ * Determine if we need to display any global banners
+ * This is added by defining a object in
+ * gitops.json:
+ * "components": {
+    "banner": [
+      {
+        "type": "info",
+        "message": "I'm a banner",
+        "resetMsgDays": 100
+      }
+    ],
+
+ * "type": "info", // Type of Alert styles, options: success, info, warning, error
+ * "message": "I'm a banner", // message to be displayed
+ * "resetMsgDays": // optional; set to number of days until displaying banner again, set to 0 to make it a browser session
+ */
+
+  render() {
+    const onBannerClose = () => {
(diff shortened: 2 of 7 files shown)


src/Layout/reducers.js
```diff
@@ -11,4 +11,23 @@ const bar = (state = {}, action) => {
   }
 };
 
-export default bar;
+export const banner = (state = {}, action) => {
+  switch (action.type) {
+  case 'RESET_BANNER': {
+    const closedBanners = state?.closedBanners?.[action.data.id];
```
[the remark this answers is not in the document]
ocshawn:
```suggestion
    const closedBanners = state?.closedBanners;
```

src/reduxStore.js
```diff
@@ -53,7 +53,7 @@ const getReduxStore = () => {
       }
 
       persistStore(store,
-        { whitelist: ['certificate'] },
+        { whitelist: ['certificate', 'banner'] },
```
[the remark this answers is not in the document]
ocshawn: it will just sit there forever, thoughts on if i should handle this?

src/components/Banner.jsx
```diff
@@ -0,0 +1,52 @@
...
+      />
+    );
+  }
+}
+
+Banner.propTypes = {
+  id: PropTypes.string.isRequired,
+  type: PropTypes.oneOf(['success', 'info', 'warning', 'error']).isRequired,
+  message: PropTypes.string.isRequired,
+  resetDate: PropTypes.number.isRequired,
```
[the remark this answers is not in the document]
ocshawn: its currently being added by reduxer.js, so it will always be there, i could refactor the code to make it option but right now its required for it to work

src/reduxStore.js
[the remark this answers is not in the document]
mfshao: I think you can supply `autoRehydrate()` with a customized config which contains a customized `stateReconciler` so you can preprocess the banner states before they made their way back to redux https://www.npmjs.com/package/redux-persist/v/4.6.0-es#autorehydrateconfig
But it might be complex to do, and there might be more elegant way exists in `redux-persist` V5+ ...
So I'm also okay if we take this `resetDate` feature out from this PR for now and backlog a new ticket for it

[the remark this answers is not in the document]
ocshawn: after looking into i added code to TopBar to look through the store and remove expired banners

src/components/layout/TopBar.jsx
```diff
@@ -18,6 +19,18 @@ const isEmailAddress = (input) => {
  * NavBar renders row of nav-items of form { name, icon, link }
  */
 class TopBar extends Component {
+  componentDidMount() {
+    // clear global store of expired banners
+    if (this.props.closedBanners) {
+      // eslint-disable-next-line no-restricted-syntax
+      for (const [key, value] of Object.entries(this.props.closedBanners)) {
+        if (Date.now() > value) {
+          this.props.onResetBanner({ id: key });
```
mfshao, line 28: I think this place is good for placing the logic but I didn't quite get the logic: we should only want to remove a banner from tthe store if the banner no longer exists in the config. If the banner still exists in the config and `Date.now() > its_reste_date`, the bannere should reappear right?

ocshawn: the logic does not check the config, but rather checks the store for expired banners and removes them. If an admin removed a banner the users browser will still remember that the banner was closed until the reset date has been hit.

mfshao: Right, and that is what I'm confused at now. Doesn't this `resetMsgDays` determines the time period that we wait until this banner show up again? So under this assumption, if the current time is greater than `resetMsgDays` it doesn't means the banner is expired, it means it should re-appear. Or did I misunderstand this logic? 🤔

ocshawn: you misunderstand, i think it will be easier if i just go threw the logic:

1. config is checked and displays banners that dont match record in local storage 
2. When user exits a banner object is stored in local storage that has timestamp of when it expires (current time plus resetMsgDays)
3. when user current time is greater than expiration time record is removed from local storage 
4. back to 1

mfshao: I see, thanks for the explaination, then I think we should update this descriptive text here https://github.com/uc-cdis/data-portal/pull/1398/files#diff-b44409708be4e154876122765a9fd3ada56184bb299149d53f7227a5db35c2eaR55 🤔 because it says the banner will re-appear

ocshawn: banner will re appear once record is removed from local storage if its still in the config

mfshao: oh I see! thanks for the walk through 😅

mfshao (approved):
🎉

PR #1399, opened 2023-08-29
(HP-1194): Allow access to specific open data based on IdP

ocshawn:
Jira Ticket: [HP-1194](https://ctds-planx.atlassian.net/browse/HP-1194)

### New Features
- only allow certain data on HEAL to be available for download with InCommon IdP login

[HP-1194]: https://ctds-planx.atlassian.net/browse/HP-1194?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Discovery/DiscoveryActionBar.tsx
+++ b/src/Discovery/DiscoveryActionBar.tsx
@@ -304,10 +304,14 @@ const handleDownloadZipClick = async (
   ).catch(() => setDownloadStatus(DOWNLOAD_FAIL_STATUS));
 };
 
-const handleDownloadManifestClick = (config: DiscoveryConfig, selectedResources: any[]) => {
+const handleDownloadManifestClick = (config: DiscoveryConfig, selectedResources: any[], healICPSRLoginNeeded: boolean) => {
   const { manifestFieldName } = config.features.exportToWorkspace;
   if (!manifestFieldName) {
     throw new Error('Missing required configuration field `config.features.exportToWorkspace.manifestFieldName`');
+  }
+
+  if (healICPSRLoginNeeded) {
+    return;
   }
   // combine manifests from all selected studies
   const manifest:any = [];
@@ -346,10 +350,15 @@ const handleExportToWorkspaceClick = async (
   setDownloadStatus: (arg0: DownloadStatus) => void,
   history: any,
   location: any,
+  healICPSRLoginNeeded: boolean,
 ) => {
   const { manifestFieldName } = config.features.exportToWorkspace;
   if (!manifestFieldName) {
     throw new Error('Missing required configuration field `config.features.exportToWorkspace.manifestFieldName`');
+  }
+
+  if (healICPSRLoginNeeded) {
+    return;
   }
 
   if (config.features.exportToWorkspace.verifyExternalLogins) {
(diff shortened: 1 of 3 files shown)


src/Discovery/DiscoveryActionBar.tsx
```diff
@@ -506,7 +527,7 @@ const DiscoveryActionBar = (props: Props) => {
             }
             type='default'
             className={`discovery-action-bar-button${(props.discovery.selectedResources.length === 0) ? '--disabled' : ''}`}
-            disabled={props.discovery.selectedResources.length === 0 || downloadStatus.inProgress}
+            disabled={props.discovery.selectedResources.length === 0 || downloadStatus.inProgress || healICPSRLoginNeeded}
```
mfshao, line 530: I think we should move this line of logic outside because it will be more complicated than this one-liner
The expected behavior should be: 
1. If user is not logged in, this button is enabled, the HEAL-ICPSR tooltip can be visible, clicking on this button redirects to the login page
2. If user is logged in via non-incommons IDP, this button is disabled, the HEAL-ICPSR tooltip is visible
3. If user is logged in via incommons IDP, this button is enabled, the HEAL-ICPSR tooltip is invisible

src/Discovery/DiscoveryActionBar.tsx
```diff
@@ -483,7 +502,9 @@ const DiscoveryActionBar = (props: Props) => {
           arrowPointAtCenter
           content={(
             <React.Fragment>
-          Directly download data (up to 250Mb) from selected studies
+              {healICPSRLoginNeeded
+                ? 'This dataset is only accessible to users who have authenticated via InCommons. Please log back in using the InCommons option.'
```
mfshao, line 506:
```suggestion
                ? 'This dataset is only accessible to users who have authenticated via InCommons. Please log in using the InCommons option.'
```
Changed my mind with this minor text update

mfshao (commented):
Overall the implementation lgtm, left a comment about an update to the button disable/enable logic

src/Discovery/DiscoveryActionBar.tsx
ocshawn: should be fixed now

src/Discovery/DiscoveryActionBar.tsx
```diff
@@ -483,21 +512,24 @@ const DiscoveryActionBar = (props: Props) => {
           arrowPointAtCenter
           content={(
             <React.Fragment>
-          Directly download data (up to 250Mb) from selected studies
+              {healICPSRLoginNeeded
+                ? 'This dataset is only accessible to users who have authenticated via InCommons. Please log in using the InCommons option.'
```
mfshao, line 516:
```suggestion
                ? 'This dataset is only accessible to users who have authenticated via InCommon. Please log in using the InCommon option.'
```
Some typos I made in the original ticket, apologies

src/Discovery/DiscoveryActionBar.tsx
```diff
@@ -483,21 +512,24 @@ const DiscoveryActionBar = (props: Props) => {
...
+                : 'Directly download data (up to 250Mb) from selected studies'}
             </React.Fragment>
           )}
         >
           <Button
             onClick={
               async () => {
-                if (props.user.username) {
+                if (props.user.username && !healICPSRLoginNeeded) {
```
mfshao, line 524: can we make the same restriction applied to all 3 buttons in this component, in other words, can we apply the same logic also to the manifest download button and the export to workspace button?

mfshao (commented):
some feedback after reviewing with @briennal

src/Discovery/DiscoveryActionBar.tsx
ocshawn: yes, i will do that

mfshao (approved):
looks great! 🎉

PR #1403, opened 2023-09-01
Removing configs for data commons.

k-burt-uch:
### Improvements
- Removing stale configs from data/config. These configs are managed in uc-cdis/cdis-manifests.


k-burt-uch:
Seems much more involved than just deleting some json files... closing for now.

PR #1407, opened 2023-09-12
(PXP-11062): fix study viewer alphabetical ordering

ocshawn:
### Bug Fixes
- fix study viewer alphabetical ordering

### DIFF
--- a/src/StudyViewer/StudyViewer.jsx
+++ b/src/StudyViewer/StudyViewer.jsx
@@ -127,9 +127,10 @@ class StudyViewer extends React.Component {
           });
         }
         if (desc) {
-          return (bVal > aVal) ? 1 : -1;
+          return bVal.localeCompare(aVal);
         }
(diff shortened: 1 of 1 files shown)


src/StudyViewer/StudyViewer.jsx
```diff
@@ -127,9 +127,10 @@ class StudyViewer extends React.Component {
           });
         }
         if (desc) {
-          return (bVal > aVal) ? 1 : -1;
+          return bVal.localeCompare(aVal);
         }
-        return (aVal > bVal) ? 1 : -1; // asc by default
+        // asc by default
+        return aVal.localeCompare(bVal);
```
paulineribeyre, line 133: since we can configure any field to sort on, i think we need it to work for types other than str, mostly int - can we add a type check

paulineribeyre (approved):

PR #1408, opened 2023-09-12
update react-relay to v 10 to match relay-compiler and relay-runtime

ocshawn:
### Dependency updates
- update react-relay to v 10.0.0 to match relay-compiler and relay-runtime


mfshao:
did you verify the submission page and query page still works after this update?

PR #1402, opened 2023-08-31
PXP-11072 fix graphiql styling

mfshao:
Jira Ticket: [PXP-11072](https://ctds-planx.atlassian.net/browse/PXP-11072)

### Bug Fixes
- Fix query page styling issue

[PXP-11072]: https://ctds-planx.atlassian.net/browse/PXP-11072?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #1421, opened 2023-09-28
(PXP-11107): fix loading of gitops.css locally

ocshawn:
Jira Ticket: [PXP-11107](https://ctds-planx.atlassian.net/browse/PXP-11107)

### Bug Fixes
- fix loading of gitops.css locally

[PXP-11107]: https://ctds-planx.atlassian.net/browse/PXP-11107?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


custom/customize.sh
```diff
@@ -29,6 +29,8 @@ fi
 
 if [ -f custom/css/$APP.css ]; then
   cp custom/css/$APP.css src/css/themeoverrides.css
+  mkdir -p public/src/css/
```
mfshao, line 32: is this because our setup for `publicPath` in webpack config is off?

ocshawn: i dont know webpack that well, it looks like we want themeoverrides.css to be passed through as a separate file

ocshawn: do you have a suggestion on how this should be handled?

mfshao: Honestly I think the issue discuribed in PXP-11107 should be fixed by https://github.com/uc-cdis/data-portal/pull/1402/files#diff-997f6766dcd75e3c8f1940f37f581261863b8c581e1a7e21e12568e5504cbf3cR31
With that being side, the loading logic for local css in portal is still a bit off: no matter what the `basename` value is, the path to local dev server css should always be `https://localhost:9443/src/css/themeoverrides.css`. This gliche should only affect you if you are working with an env whose portal has a customized `basename` value like HEAL. But for envs whose `basename == '/'`, local dev server dev loading should be working after that fix

paulineribeyre: This ^ was merged on 09/14, I tried using a local gitops.css on 09/20 using master branch and couldn't get it to work

mfshao: hmmm can you try again now or let me know which env are you using? I'll try to replicate it

ocshawn: its not working for anything that i have tested locally, currently testing https://ibdgc.datacommons.io/dev.html/, where is the file being served from? from what i see webpack does not expose any files not in the public directory so unless you place the file there its not being loaded locally

mfshao: hmm I think you are right, I must be testing something wrong last night, so this looks good
then can you update https://github.com/uc-cdis/data-portal/blob/master/boot.js#L31 so that it uses `https://localhost:9443/src/css/themeoverrides.css` only? It shouldn't be changing base on basenames

custom/customize.sh
```diff
@@ -29,6 +29,8 @@ fi
...
+  cp custom/css/$APP.css public/src/css/themeoverrides.css
```
mfshao, line 33: it would be good if this script can clean up the old css file under `/public` each time it runs, and only copy new ones over if there is one. So old css files won't interference with new runs

ocshawn: looking into fixing this, good call

boot.js
```diff
@@ -19,18 +19,9 @@
...
   scriptNode.type = 'text/javascript';
   document.body.appendChild(scriptNode);
-  // create theme override node in body
-  const themeOverridesSrc = (basename && basename !== '/') ? `${basename}/src/css/themeoverrides.css?_=${cssVersion}` : `/src/css/themeoverrides.css?_=${cssVersion}`;
-  let linkNode = document.createElement('link');
-  linkNode.href = themeOverridesSrc;
-  linkNode.type = 'text/css';
-  linkNode.id = 'gen3-theme-overrides';
-  linkNode.rel = 'stylesheet';
-  document.body.appendChild(linkNode);
```
ocshawn, line 29: removed this because i don't think the relative file path is necessary

runWebpack.sh
```diff
@@ -108,6 +108,7 @@ gitops_config() {
         cp $copySource/*.* $copyDest/
       else
         echo "INFO: gitops_config - no $copySource in gitops"
+        rm $copyDest
```
ocshawn, line 111: if file is not found it removes any copy that could already exist

boot.js
mfshao: i think part of this logic is still needed because it will forces the browser to reload the css files each time when there is a change

runWebpack.sh
mfshao: if `copyDest` doesn't exists this will throw a bunch of errors?

boot.js
ocshawn: put it back

runWebpack.sh
ocshawn: added flag to not throw errors if file does not exist

boot.js
ocshawn: on further inspection this loads the css file from the server, you need to rerun the dev script to see your changes as it has to copy the file so its never going to have up to date changes. Should i leave this code in or take it out? Im thinking removing it is  the better call so locale development does not get confused if they are seeing server code or local code

mfshao: yeah so if we remove this section them basically in local dev mode we will only be loading `themeoverrides.css` from local side, and not from remote side right?
I think that is a good idea and we should be good toi remove them then

ocshawn: thats correct, ill run some tests in different environments to make sure its good but thats how it should work

mfshao (approved):

PR #1422, opened 2023-09-28
fix value is null

mfshao:
### Bug Fixes
- Discovery: fixed a bug causing the details page to crash when encountering null values


jarvisraymond-uchicago (approved):
Looks good to me!

PR #1434, opened 2023-10-12
Fix/bugs

mfshao:
### Bug Fixes
- Misc: fixed a bug caused by the webpack startup script cannot correctly execute if there are directories in config path
- SystemUsePopup: fixed a bug causing the acceptance cookie cannot be set due to incorrect domain values

### DIFF
--- a/src/Popup/SystemUseWarningPopup.jsx
+++ b/src/Popup/SystemUseWarningPopup.jsx
@@ -3,7 +3,7 @@ import { connect } from 'react-redux';
 import PropTypes from 'prop-types';
 import Popup from '../components/Popup';
 import { components } from '../params';
-import { hostname } from '../localconf';
+import { hostnameWithSubdomain } from '../localconf';
 import { updateSystemUseNotice } from '../actions';
 
 const handleAcceptWarning = () => {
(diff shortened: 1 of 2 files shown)


src/Popup/SystemUseWarningPopup.jsx
```diff
@@ -3,29 +3,24 @@ import { connect } from 'react-redux';
...
   /**
    * If policy is accepted then set systemUseWarnPopup to false, hiding the
    * warning.
    */
-  // set a new cookie indicating we have accepted this policy
-
-  const hostnameParts = hostname.split('.');
-  const hLen = hostnameParts.length;
-  const hostnameNoSubdomain = (hLen > 2) ? hostnameParts.splice(hLen - 2).join('.') : hostname;
-  const domain = hostnameNoSubdomain.endsWith('/') ? hostnameNoSubdomain.slice(0, -1) : hostnameNoSubdomain;
```
mfshao, line 19: for example, with `hostname='https://healdata.org'`, the value of `domain` from above code will be `https://healdata`, which is incorrect

mfshao, line 19: for example, if `hostname='https://healdata.org'`, then the value of `domain` will be `https://healdata.org`, which causes a problem  down the road

paulineribeyre: i don't understand the example, can you give an example with a subdomain? does it strip off the subdomain?

mfshao: yeah, so if `hostname='https://preprod.healdata.org'`, then `domain` will be `healdata.org`

mfshao: but then this is also problematic, because later when we were setting the cookie we were appending a `.` to it, so it became `.healdata.org`
which means if you accept a DUA on `preprod.healdata.org`, that cookie will also prevent the DUA to popup on `healdata.org`

mfshao: and in the case of no subdomain, for example `hostname='https://healdata.org'`, `domain` here will be `https://healdata.org`, and then down the road when setting the cookie it will became `.https://healdata.org`, which is incorrect
And this PR is to fix this

paulineribeyre (approved):

PR #1450, opened 2023-11-09
Feat/data download list actions

jarvisraymond-uchicago:
Jira Ticket: [HP-1176](https://ctds-planx.atlassian.net/browse/HP-1176)
* Download Study-Level Metadata and Download File Manifest download associated JSON files
* Adds button for Download All Files using batch export job, but is not reproducible in QA environment; the download never successfully completes
* Download Variable-Level Metadata button will be done in subsequent ticket
* Test with studies: hdp99999 ICPSR TEST and hdp88888 ICPSR TEST

output

### New Features
* Adds buttons for  Download Study-Level Metadata, Download File Manifest and Download All Files

[HP-1176]: https://ctds-planx.atlassian.net/browse/HP-1176?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/Discovery/DiscoveryDetails/DataDownloadList/ActionButtons/ActionButtons.css
```diff
@@ -0,0 +1,7 @@
+.discovery-modal_buttons-row button {
+  height: auto;
+  min-height: 54px;
```
mfshao, line 3: This seems to be too high, I think it would be better if the button height can fill parent container height

src/Discovery/DiscoveryDetails/DataDownloadList/ActionButtons/ActionButtons.tsx
```diff
@@ -0,0 +1,77 @@
...
+        )}
+        {discoveryConfig?.features.exportToWorkspace.enableDownloadManifest && (
+          <Col flex='1 0 auto'>
+            <Button
+              className='discovery-action-bar-button'
+              onClick={() => {
+                HandleDownloadManifestClick(
+                  discoveryConfig,
+                  [resourceInfo],
+                  false,
```
mfshao, line 55: This button should behaves exactly the same as its siblings on the Discovery table, that is: if user is not logged in, display button label as `Login to Download Manifest`, and redirects user to the login page on clicking, then redirects user back after loggin in. Right now this button is enabled even for unauthenticated users

src/Discovery/DiscoveryDetails/DataDownloadList/ActionButtons/ActionButtons.tsx
```diff
@@ -0,0 +1,77 @@
...
+            </Button>
+          </Col>
+        )}
+        {discoveryConfig?.features.exportToWorkspace.enableDownloadZip && (
+          <Col flex='1 0 auto'>
+            <Button
+              className='discovery-action-bar-button'
+              onClick={() => DownloadAllFiles(studyIDs, downloadStatus, setDownloadStatus)}
+            >
+              Download All Files
```
mfshao, line 69: Same as above, but this button should also obey any additional repository specific auth restrictions for data access, see how `checkFederatedLoginStatus()` and `healRequiredIDPLogic()` is being used in https://github.com/uc-cdis/data-portal/blob/master/src/Discovery/DiscoveryActionBar.tsx

src/Discovery/DiscoveryDetails/DataDownloadList/ActionButtons/DownloadUtils/DownloadAllFiles.tsx
```diff
@@ -0,0 +1,165 @@
...
+          uid,
+          downloadStatus,
+          setDownloadStatus,
+        );
+      }
+    },
+  );
+};
+
+const DownloadAllFiles = (
```
mfshao, line 107: We probably should add above-mentioned `checkFederatedLoginStatus()` and `healRequiredIDPLogic()` related logics from https://github.com/uc-cdis/data-portal/blob/master/src/Discovery/DiscoveryActionBar.tsx into here

src/Discovery/DiscoveryDetails/DataDownloadList/DataDownloadList.tsx
```diff
@@ -1,16 +1,25 @@
...
+import ActionButtons from './ActionButtons/ActionButtons';
 
-const DataDownloadList = ({ sourceFieldData }: any) => {
+const DataDownloadList = ({
+  discoveryConfig,
+  resourceInfo,
+  sourceFieldData,
+}) => {
   const data = ProcessData(sourceFieldData);
   if (data.length === 0) {
```
mfshao, line 14: By the design https://docs.google.com/presentation/d/1W0ACAiegMY3QfdUt-8ISQoDjKw6yVj9DqLSl42gCmtU if a type of button is configured to show up in this dataDownloadList from portal config, then they should alwyas be visible across all the study details pages. For example, if `enableDownloadManifest` is set to true in config, then for studies that doesn't have a `__manifest` value in its metadata (means `data.length === 0` here), the "Download File Manifest" button should still be visible, but disabled.
In order to do that we need to change the logic here, to let this component renders the `ActionButtons` child even if `data.length === 0`, but don't render the actual `List` of files

mfshao (commented):
good progress, buttons seem to be working. Left some comment wrt some behavior and/or appearance concerns

src/Discovery/DiscoveryDetails/DataDownloadList/ActionButtons/ActionButtons.css
jarvisraymond-uchicago: Good idea. Removed min-height, set height to 100% and set line-height to 1.2 to decrease overall button height while retaining a consistent height across all buttons. 

Design: 
image

Before: 
image

After: 
image

src/Discovery/DiscoveryDetails/DataDownloadList/ActionButtons/ActionButtons.tsx
jarvisraymond-uchicago: Created a custom hook for this, UseHandleRedirectToLoginClick. Passed prop isUserLoggedIn from DiscoveryDetails to DataDownloadList and have buttons for each use case now.

jarvisraymond-uchicago: Wrote a custom hook UseHandleRedirectToLoginClick.ts to redirect the user to login page on clicking if logged out, then to download once logged in. Created unique buttons for each use case, which conditionally render based on new prop for component ```isUserLoggedIn``` determined in DiscoveryDetails

src/Discovery/DiscoveryDetails/DataDownloadList/ActionButtons/ActionButtons.tsx
jarvisraymond-uchicago: Added logic to check user is both logged in and _healLoginNeeded_ is not true. Moved logic to _ProcessHealRequiredIdentityProviderInfo.ts_ into a separate function called from _DiscoveryDetails.tsx_, to help avoid ActionsButtons.tsx from doing processing not directly related to the action buttons.

jarvisraymond-uchicago: Note: todo Dec. 7th -> address missing checkFederatedLoginStatus() here.

src/Discovery/DiscoveryDetails/DataDownloadList/ActionButtons/DownloadUtils/DownloadAllFiles.tsx
jarvisraymond-uchicago: Note: todo Dec. 7th -> address missing checkFederatedLoginStatus() here.

src/Discovery/DiscoveryDetails/DataDownloadList/DataDownloadList.tsx
jarvisraymond-uchicago: This has been updated. The conditional for data.length === 0 has been wrapped around only the ```List``` component

```
 <ActionButtons
        isUserLoggedIn={isUserLoggedIn}
        discoveryConfig={discoveryConfig}
        resourceInfo={resourceInfo}
        healLoginNeeded={healLoginNeeded}
      />
      {data.length !== 0 && (
        <List
          itemLayout='horizontal'
          dataSource={data}
          renderItem={(item: DataDownloadListItem) => (
            <List.Item>
              <List.Item.Meta
                title={item.title}
                description={item.description || ''}
              />
            </List.Item>
          )}
          ```

File: src/Discovery/DiscoveryDetails/DataDownloadList/DataDownloadList.tsx [hunk 6]
```diff
@@ -1,28 +1,53 @@
...
+  discoveryConfig,
+  resourceInfo,
+  sourceFieldData,
+  healLoginNeeded,
+}: DataDownloadListProps) => {
   const data = ProcessData(sourceFieldData);
-  if (data.length === 0) {
-    return null;
-  }
+  const noData = data.length === 0;
```
mfshao, line 26: I think this is really hard to show up if we are on the discovery details page, because this basically means the selected study has an empty `__manifest` field. But in this case, the `DataDownloadList` component won't be showing up in the first place due to https://github.com/uc-cdis/data-portal/blob/bd988afb03b2271a000d44cb5117ccf7a12554d5/src/Discovery/DiscoveryDetails/DiscoveryDetails.tsx#L261-L266
For now maybe this is an acceptable behavior, since these buttons are part of the `DataDownloadList` component, make only part of them to show up is kind of tricky. And we can defer them to a later ticket. But can you update to code to add some comments in there to say why this popover won't show up and how can we fix it later?

src/Discovery/DiscoveryDetails/DataDownloadList/ActionButtons/ActionButtons.tsx
```diff
@@ -0,0 +1,165 @@
...
+  const { HandleRedirectToLoginClick } = UseHandleRedirectToLoginClick();
+  const [downloadStatus, setDownloadStatus] = useState({
+    inProgress: false,
+    message: { title: '', content: <React.Fragment />, active: false },
+  });
+  const history = useHistory();
+  const location = useLocation();
+  const showDownloadStudyLevelMetadataButtons: boolean | undefined = discoveryConfig?.features.exportToWorkspace.studyMetadataFieldName
+    && discoveryConfig?.features.exportToWorkspace.enableDownloadStudyMetadata
+    && resourceInfo?.study_metadata;
```
mfshao, line 38: instead of hard coding `study_metadata`, this should be the value of `discoveryConfig?.features.exportToWorkspace.studyMetadataFieldName`

src/Discovery/DiscoveryDetails/DataDownloadList/ActionButtons/ActionButtons.tsx
```diff
@@ -0,0 +1,165 @@
...
+        */}
+        {showDownloadStudyLevelMetadataButtons && (
+          <Col flex='1 0 auto'>
+            <ConditionalPopover>
+              <Button
+                className='discovery-action-bar-button'
+                disabled={noData}
+                onClick={() => DownloadJsonFile(
+                  'study-level-metadata',
+                  resourceInfo.study_metadata,
```
mfshao, line 75: same as above

mfshao (commented):
I think this looks great! deployed and tested in QA, seems to be working. And the refactoring and adding tests are brillent
I do have some comments, see below

src/Discovery/DiscoveryDetails/DataDownloadList/DataDownloadList.tsx
jarvisraymond-uchicago: Instead of adding code comments that might end up living in the code for a long time and end up out of date, could we instead create a ticket in JIRA that explains the requirement and identifies in the technical details this section of code that needs to be changed? I've found this approach helps avoid code bases with comments that are no longer relevant, difficult to understand and situations where developers are unsure if they can or should delete the comments. Let me know what you think; I can add the code comments if you prefer.

mfshao: sounds good, i can create a ticket

src/Discovery/DiscoveryDetails/DataDownloadList/ActionButtons/ActionButtons.tsx
jarvisraymond-uchicago: I updated this to reference this variable: 
```
  const studyMetadataFieldNameReference: string | undefined = discoveryConfig?.features.exportToWorkspace.studyMetadataFieldName;

```

In this manner: 
```
  const showDownloadStudyLevelMetadataButtons = Boolean(
    discoveryConfig?.features.exportToWorkspace.enableDownloadStudyMetadata
      && studyMetadataFieldNameReference
      && resourceInfo?.[studyMetadataFieldNameReference],
  );
  ```

File: src/Discovery/DiscoveryDetails/DataDownloadList/ActionButtons/ActionButtons.tsx (hunk 8 above)
Comment (jarvisraymond-uchicago) [reply src/Discovery/DiscoveryDetails/DataDownloadList/ActionButtons/ActionButtons.tsx]: This has been updated to use this variable: 
```
const studyMetadataFieldNameReference: string | undefined = discoveryConfig?.features.exportToWorkspace.studyMetadataFieldName;
```
In this manner: 
```
                onClick={() => DownloadJsonFile(
                  'study-level-metadata',
                  studyMetadataFieldNameReference
                      && resourceInfo[studyMetadataFieldNameReference],
                )}
                ```

mfshao (approved):
✨

PR #1455, opened 2023-11-17
PXP-11183 feat dd filter bot on intake

mfshao:
Jira Ticket: [PXP-11183](https://ctds-planx.atlassian.net/browse/PXP-11183)

### Improvements
- Misc: filter known bot seesion on intake for DD RUM

### Dependency updates
- Added `crawler-user-agents`

[PXP-11183]: https://ctds-planx.atlassian.net/browse/PXP-11183?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/localconf.js
+++ b/src/localconf.js
@@ -1,3 +1,5 @@
+import crawlers from 'crawler-user-agents';
+
 /* eslint-disable prefer-destructuring */
 const { components, requiredCerts, config } = require('./params');
 
@@ -160,6 +162,8 @@ function buildConfig(opts) {
       ddSampleRate = config.ddSampleRate;
     }
   }
+  const ddKnownBotPattern = crawlers.map((c) => c.pattern).join('|');
+  const ddKnownBotRegex = new RegExp(ddKnownBotPattern, 'i');
 
   // backward compatible: homepageChartNodes not set means using graphql query,
   // which will return 401 UNAUTHORIZED if not logged in, thus not making public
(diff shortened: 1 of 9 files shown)


src/localconf.js
```diff
@@ -160,6 +162,8 @@ function buildConfig(opts) {
       ddSampleRate = config.ddSampleRate;
     }
   }
+  const ddKnownBotPattern = crawlers.map((c) => c.pattern).join('|');
+  const ddKnownBotRegex = new RegExp(ddKnownBotPattern, 'i');
```
k-burt-uch, line 166: So this takes all the patterns from the crawler-user-agents.json here: https://github.com/monperrus/crawler-user-agents/blob/master/crawler-user-agents.json
 
and creates a regex ORing them all together to give us our ddKnownBotRegex

Then our conditional sample rate test if the user agent from a "user" matches on that regex - if it does then we don't count the sample otherwise we do? 

Is my understanding on that correct?

mfshao: yeah you are right, if the value of user agent matches the pattern of known crawlers, then we set the DD RUM session sample rate to 0, which mean no sessions are being send to DD

this method is documented in the official DD RUM user doc https://docs.datadoghq.com/real_user_monitoring/guide/identify-bots-in-the-ui/?s=userAgent#filter-out-bot-sessions-on-intake

PR #1467, opened 2023-12-19
add additional typecheck to Discovery/renderAggregation

craigrbarnes:
Jira Ticket: [PXP-11229](https://ctds-planx.atlassian.net/browse/PXP-11229)

### Bug Fixes
Adds an additional check to DiscoverySummary `renderAggregation` to only sum strings or number values.  If the value is something other than a string or number, an array, for example, the sum value explodes.

### Deployment changes

[PXP-11229]: https://ctds-planx.atlassian.net/browse/PXP-11229?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -18,7 +18,7 @@
         "@fortawesome/free-brands-svg-icons": "^5.14.0",
         "@fortawesome/free-solid-svg-icons": "^5.2.0",
         "@fortawesome/react-fontawesome": "^0.2.0",
-        "@gen3/guppy": "^0.17.1",
+        "@gen3/guppy": "^0.18.0",
         "@gen3/ui-component": "^0.11.4",
         "@reactour/tour": "^2.12.0",
         "@upsetjs/venn.js": "^1.4.2",
@@ -3761,18 +3761,17 @@
       "dev": true
     },
     "node_modules/@gen3/guppy": {
-      "version": "0.17.1",
-      "resolved": "https://registry.npmjs.org/@gen3/guppy/-/guppy-0.17.1.tgz",
-      "integrity": "sha512-Vr5N11p2vi5iVrmXkTkfuUHpx3vvNj8X2fdnIUtFqWqQVg0K8w7G3L4rTCEhTreSs+JDgj0Q9qRZMFFWCcfymA==",
+      "version": "0.18.0",
+      "resolved": "https://registry.npmjs.org/@gen3/guppy/-/guppy-0.18.0.tgz",
+      "integrity": "sha512-6G7lUn17G08RxxuG3pWDvjI06duBvLVyTVFcVhTq8Jv/nrH7utnfoTswV162+nSOk8z7sFNn79nzCAdKcPfAOg==",
       "dependencies": {
-        "@apollo/server": "^4.7.5",
+        "@apollo/server": "^4.9.4",
         "@elastic/elasticsearch": "~7.13.0",
         "@gen3/ui-component": "^0.11.4",
-        "array.prototype.flat": "^1.2.2",
-        "array.prototype.flatmap": "^1.3.1",
+        "array.prototype.flat": "^1.3.2",
+        "array.prototype.flatmap": "^1.3.2",
         "body-parser": "^1.20.2",
         "cors": "^2.8.5",
-        "csurf": "^1.11.0",
         "dd-trace": "^4.5.0",
         "express": "^4.18.2",
         "file-saver": "^2.0.5",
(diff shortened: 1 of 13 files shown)


src/Discovery/DiscoverySummary.tsx
```diff
@@ -21,7 +21,16 @@ const renderAggregation = (aggregation: AggregationConfig, studies: any[] | null
   switch (type) {
   case 'sum': {
     // parse any string representation of an integer
-    fields = fields.map((item) => (typeof item === 'string' ? parseInt(item, 10) || 0 : item));
+    // if it's not a number, return 0 so as not to break the sum
```
jarvisraymond-uchicago, line 24: Nitpick: instead of using a comment here, the code could be wrapped in a function with an explanatory name. Reduces likelihood that code changes and the comment does not.  

Examples: 
```
const checkForNonNumericInput = (input) => {
  ....
}
```
or
```
const handleNonNumericInput = () => {
   ...
}
```

src/Discovery/DiscoverySummary.tsx
```diff
@@ -21,7 +21,16 @@ const renderAggregation = (aggregation: AggregationConfig, studies: any[] | null
...
+    fields = fields.map((item) => {
+      if (typeof item === 'number') {
+        return item;
+      }
+      if (typeof item === 'string') {
```
jarvisraymond-uchicago, line 29: Nitpick: changing this to an ```else if``` is more efficient and reduces number of comparisons executed (but practically this doesn't matter).

src/Discovery/DiscoverySummary.tsx
```diff
@@ -21,7 +21,16 @@ const renderAggregation = (aggregation: AggregationConfig, studies: any[] | null
...
+        return parseInt(item, 10) || 0;
```
jarvisraymond-uchicago, line 30: Nitpick: I believe the second parameter for parseInt sets the base, however the default for parseInt is base 10 so the second parameter isn't needed.

jarvisraymond-uchicago (approved):
Great catch! This will hopefully reduce runtime errors! 

* Left some nitpick comments, update if deemed necessary. 
* Please add JIRA ticket to PR description using format: 
```Jira Ticket: [VADC-764](https://ctds-planx.atlassian.net/browse/VADC-764)```

**Comment by mfshao** [comment]
what is the relationship between this PR and VADC-764?

**Comment by jarvisraymond-uchicago** [comment]
> what is the relationship between this PR and VADC-764?

None, however this is the format needed for PR's
* Please add JIRA ticket to PR description using format:
Jira Ticket: [VADC-764](https://ctds-planx.atlassian.net/browse/VADC-764)

**Review by mfshao — COMMENTED** [review]
trying to clarify the goal here: if the target field are arrays of strings or numbers, shouldn't the expected behavior of this `sum` function is to sum the value of all numeric elements in those arrays?

**Comment by craigrbarnes** [comment]
> trying to clarify the goal here: if the target field are arrays of strings or numbers, shouldn't the expected behavior of this `sum` function is to sum the value of all numeric elements in those arrays?

True, I can add support for arrays as well.

File: src/Discovery/DiscoverySummary.tsx (hunk 1 above)
Comment (craigrbarnes) [reply src/Discovery/DiscoverySummary.tsx]: used your first suggestion

File: src/Discovery/DiscoverySummary.tsx (hunk 2 above)
Comment (craigrbarnes) [reply src/Discovery/DiscoverySummary.tsx]: Since I added handling of an array and it's a return. I left it as 3 if..then blocks

File: src/Discovery/DiscoverySummary.tsx (hunk 3 above)
Comment (craigrbarnes) [reply src/Discovery/DiscoverySummary.tsx]: It is but ESlint wants to make it required

File: package-lock.json [hunk 4]
```diff
@@ -61625,7 +61625,7 @@
         "fb-watchman": "^2.0.0",
         "fbjs": "^3.0.0",
         "glob": "^7.1.1",
-        "immutable": "*",
+        "immutable": "~3.7.6",
```
jarvisraymond-uchicago, line 61628: Is this change intentional? If not this file should be reverted. 

If you want to revert the file to its state in master in git:

```git checkout origin/master [filename]```

craigrbarnes: It was not, I will do that now.

jarvisraymond-uchicago (approved):

mfshao (approved):

mfshao:
those failed CI tests doesn't look like related to this PR

PR #1468, opened 2023-12-20
MIDRC-550 Replace nested empty values with 'null' in explorer display

paulineribeyre:
Jira Ticket: [MIDRC-550](https://ctds-planx.atlassian.net/browse/MIDRC-550)

Before:
Screenshot 2023-12-20 at 2 46 45 PM
Screenshot 2023-12-20 at 2 46 34 PM

After:
Screenshot 2023-12-20 at 2 42 49 PM
Screenshot 2023-12-20 at 2 42 32 PM

### Improvements
- Replace nested empty values with 'null' in explorer display

[MIDRC-550]: https://ctds-planx.atlassian.net/browse/MIDRC-550?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/GuppyDataExplorer/ExplorerTable/index.jsx
+++ b/src/GuppyDataExplorer/ExplorerTable/index.jsx
@@ -76,8 +76,13 @@ class ExplorerTable extends React.Component {
           const nestedChildFieldName = fieldStringsArray.slice(1, fieldStringsArray.length).join('.');
           // some logic to handle depends on wether the child field in raw data is an array or not
           if (_.isArray(row.value)) {
-            // replace empty nested values with null so we can still join cleanly and display 1 value per nested row
-            valueStr = row.value.map((x) => _.get(x, nestedChildFieldName) || 'null').join(', ');
+            valueStr = row.value.map((x) => {
+              let val = _.get(x, nestedChildFieldName);
+              // replace empty, null and undefined nested values with explicit "null" strings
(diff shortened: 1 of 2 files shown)


src/GuppyDataExplorer/ExplorerTable/index.jsx
```diff
@@ -76,7 +76,8 @@ class ExplorerTable extends React.Component {
           const nestedChildFieldName = fieldStringsArray.slice(1, fieldStringsArray.length).join('.');
           // some logic to handle depends on wether the child field in raw data is an array or not
           if (_.isArray(row.value)) {
-            valueStr = row.value.map((x) => _.get(x, nestedChildFieldName)).join(', ');
+            // replace empty nested values with null so we can still join cleanly and display 1 value per nested row
+            valueStr = row.value.map((x) => _.get(x, nestedChildFieldName) || 'null').join(', ');
```
k-burt-uch, line 80: Is it possible for `_.get(x, nestedChildFieldName)` to return a value other than string? If so, then falsy values like 0 for numeric types or even a "false" boolean will show up as 'null'.

k-burt-uch (approved):
value == null will catch null and undefined, val==='' catches empty nested values. other falsy values such as 0 will not get caught which is good. Approved.

PR #1476, opened 2024-01-18
Feat/download variable metadata button

jarvisraymond-uchicago:
Jira Ticket: [HP-1224](https://ctds-planx.atlassian.net/browse/HP-1224)

Note: If the first endpoint fails, button is disabled. However if the second endpoint fails, the fail message shows the data dictionary ID that failed. If the download is successful, no message is shown and the ZIP file downloads. 

Screenshots:
output

### New Features
* Adds download variable metadata button to action buttons within data download list 

### Improvements
* Makes the _DownloadAllModal_ a more reusable and generic _DownloadModal_ within data download list component
* Adds additional type checking to CheckHealLoginNeeded to avoid JS errors now that function is being called in more places. 

[HP-1224]: https://ctds-planx.atlassian.net/browse/HP-1224?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Discovery/DiscoveryDetails/DataDownloadList/ActionButtons/ActionButtons.tsx
+++ b/src/Discovery/DiscoveryDetails/DataDownloadList/ActionButtons/ActionButtons.tsx
@@ -1,5 +1,7 @@
 import React, { useEffect, useState } from 'react';
-import { Col, Row, Button, Popover } from 'antd';
+import {
+  Col, Row, Button, Popover,
+} from 'antd';
 import { useHistory, useLocation } from 'react-router-dom';
 import { DiscoveryConfig } from '../../../DiscoveryConfig';
 import { DiscoveryResource } from '../../../Discovery';
@@ -34,28 +36,26 @@ const ActionButtons = ({
   const history = useHistory();
   const location = useLocation();
 
-  const studyMetadataFieldNameReference: string | undefined =
-    discoveryConfig?.features.exportToWorkspace.studyMetadataFieldName;
-  const manifestFieldName: string | undefined =
-    discoveryConfig?.features.exportToWorkspace.manifestFieldName;
+  const studyMetadataFieldNameReference: string | undefined = discoveryConfig?.features.exportToWorkspace.studyMetadataFieldName;
+  const manifestFieldName: string | undefined = discoveryConfig?.features.exportToWorkspace.manifestFieldName;
   const showDownloadStudyLevelMetadataButton = Boolean(
-    discoveryConfig?.features.exportToWorkspace.enableDownloadStudyMetadata &&
-      studyMetadataFieldNameReference &&
-      resourceInfo?.[studyMetadataFieldNameReference]
+    discoveryConfig?.features.exportToWorkspace.enableDownloadStudyMetadata
+      && studyMetadataFieldNameReference
+      && resourceInfo?.[studyMetadataFieldNameReference],
   );
   const showDownloadFileManifestButtons = Boolean(
-    discoveryConfig?.features.exportToWorkspace.enableDownloadManifest
+    discoveryConfig?.features.exportToWorkspace.enableDownloadManifest,
   );
   const showDownloadAllFilesButtons = Boolean(
-    discoveryConfig?.features.exportToWorkspace.enableDownloadZip
+    discoveryConfig?.features.exportToWorkspace.enableDownloadZip,
   );
   const verifyExternalLoginsNeeded = Boolean(
-    discoveryConfig?.features.exportToWorkspace.verifyExternalLogins
+    discoveryConfig?.features.exportToWorkspace.verifyExternalLogins,
   );
 
   const showDownloadVariableMetadataButton = Boolean(
-    discoveryConfig.features.exportToWorkspace.variableMetadataFieldName &&
-      discoveryConfig.features.exportToWorkspace.enableDownloadVariableMetadata
+    discoveryConfig.features.exportToWorkspace.variableMetadataFieldName
+      && discoveryConfig.features.exportToWorkspace.enableDownloadVariableMetadata,
   );
   const [dataDictionaryInfo, setDataDictionaryInfo] = useState({
     noVariableLevelMetadata: true,
@@ -67,18 +67,17 @@ const ActionButtons = ({
       discoveryConfig,
       resourceInfo,
       showDownloadVariableMetadataButton,
-      setDataDictionaryInfo
+      setDataDictionaryInfo,
     );
   }, [resourceInfo]);
 
-  const ConditionalPopover = ({ children }) =>
-    noData ? (
-      <Popover title={'This file is not available for the selected study'}>
-        {children}
-      </Popover>
-    ) : (
-      children
-    );
+  const ConditionalPopover = ({ children }) => (noData ? (
+    <Popover title={'This file is not available for the selected study'}>
+      {children}
+    </Popover>
+  ) : (
+    children
+  ));
 
   return (
     <div className='discovery-modal_buttons-row' data-testid='actionButtons'>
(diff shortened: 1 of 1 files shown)


src/Discovery/DiscoveryDetails/DataDownloadList/ActionButtons/DownloadUtils/DownloadVariableMetadata/DownloadVariableMetadata.tsx
```diff
@@ -0,0 +1,104 @@
...
+    if (statusResponse.status !== 200 || !data) {
+      setDownloadStatus(DOWNLOAD_FAIL_INFO);
+    } else {
+      try {
+        const dataDictionaries: IdataDictionaries = data.data_dictionaries;
+        if (Object.keys(dataDictionaries).length !== 0) {
+          fetchDataForAllFiles(dataDictionaries);
+        }
+      } catch (error) {
+        setDownloadStatus(DOWNLOAD_FAIL_INFO);
```
mfshao, line 96: So you have set the download status by using `createUniqueDownloadErrorMsg()` when there is an error downloading a particular DD in `fetchData()`. And then you called `fetchData()` in `fetchDataForAllFiles` and surronded it with a try-catch block. Would that make this catch block here redundent?

src/Discovery/DiscoveryDetails/DataDownloadList/DataDownloadList.tsx
```diff
@@ -33,7 +36,7 @@ const DataDownloadList = ({
         healLoginNeeded={healLoginNeeded}
         noData={noData}
       />
-      {!noData && (
+      {!noData && showList && (
```
mfshao, line 39: Can you explain a bit why we need this `showList` variable here in addition to `noData`?

src/Discovery/DiscoveryDetails/DataDownloadList/ActionButtons/ActionButtons.tsx
```diff
@@ -62,20 +70,32 @@ const ActionButtons = ({
...
-        */}
-        {showDownloadStudyLevelMetadataButtons && (
+        {showDownloadVariableMetadataButton && (
+          <Col flex='1 0 auto'>
+            <Button
+              className='discovery-action-bar-button'
+              onClick={() => {
+                DownloadVariableMetadata(
+                  resourceInfo,
+                  setDownloadVariableMetadataStatus,
```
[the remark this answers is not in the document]
jarvisraymond-uchicago: My concern is that if they both share the same state, unexpected UI behavior could occur that would make it difficult to use the application in some edge cases. For example, a user could press the _Download All Files_ button, and while that is downloading they could press the _Download Variable Metadata_ button. I think it would be  possible that they could both fail one after the other, and the application would hide/show and change the text of the modal while the user is reading it if they share the same state. 

One alternative to hard coding individual Download Modal components would be to dynamically generate them by storing all the Download Modal statuses in a state containing an array of these statuses _DownloadModalStatuses_. Then we could create each DownloadModal by mapping through the DownloadModalStatuses. 

This might be a good idea if there are going to be many download statuses. Let me know what you think.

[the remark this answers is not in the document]
mfshao: Yeah the edge case you mentioned makes sense. I think the regular way we were using to deal with this situation is to let only one button in this group being able to get triggered at any certain time. That is, when user clicks on a button, that button goes into `pending` + `disabled` state, while all the other buttons will go into `disable` state. And they don't get re-enabled until the pervious button action has finished. This can be done bu using the shared state variable for download status across this button group.
We can discuss further about this approach if you would like.

src/Discovery/DiscoveryDetails/DataDownloadList/ActionButtons/DownloadUtils/DownloadVariableMetadata/DownloadVariableMetadata.tsx
jarvisraymond-uchicago: It is possible that _fetchDataForAllFiles_ will not execute if the _Object.keys(dataDictionaries).length !== 0_ is false (i.e. there's no data dictionaries associated with the study). In that case the catch block at line 95 executes and shows the warning. 

An example: Study  HDP00258 is not configured for Data Dictionaries, so its data.datadictionaries is undefined, causing the catch block to execute.
output

An alternative approach would be to disable the button for these studies, but I think that would require an API call each time. Let me know what you think.

mfshao: I see, in this case, I think we can move this line that is getting DD keys/GUIDs from the input metadata https://github.com/uc-cdis/data-portal/pull/1476/files#diff-c6cbd176398b0ad195cfb42288fcf60a404070635158552da44e2e4442f40a89R91 into src/Discovery/DiscoveryDetails/DataDownloadList/ActionButtons/ActionButtons.tsx
So it will be something like

> const dataDictionaries: IdataDictionaries = data[discoveryConfig.features.exportToWorkspace.variableMetadataFieldName];
> const noVLMDData = !dataDictionaries || dataDictionaries.length == 0
> // use noVLMDData to disable/enable the download VLMD button at L82

In this way, we can disable the download VLMD button for studies that doesn't have VLMD attached to them in metadata without making additional API calls, and we can also simplify the try-catch blocks in this file

src/Discovery/DiscoveryDetails/DataDownloadList/DataDownloadList.tsx
jarvisraymond-uchicago: This could ensure against a study having a valid  resource field value but is missing the correct data after the data is processed. I can update the variable name from _showList_ to _resourceFieldValueIsValid_ which would make more sense

src/Discovery/DiscoveryDetails/DataDownloadList/ActionButtons/ActionButtons.tsx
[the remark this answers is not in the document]
jarvisraymond-uchicago: I couldn't find a pending state in AntD's documentation, but I implemented this using a loading state. The _inProgress_ data type in the _downloadStatus_ is now a string and either references what is currently in progress or is an empty string. The Download All Files button and the Download Variable-level Metadata disable all buttons and display the loading icon while downloading and then reset after. The other two buttons do not, as they download instantaneously.

src/Discovery/DiscoveryDetails/DataDownloadList/DataDownloadList.tsx
mfshao: hmm okay I thin i get it now, thanks

src/Discovery/DiscoveryDetails/DataDownloadList/ActionButtons/DownloadUtils/DownloadVariableMetadata/DownloadVariableMetadata.tsx
jarvisraymond-uchicago: This has been updated: the try catch block in _DownloadVariableMetadata_ has been removed. The logic for downloading the data dictionary info has been placed into a separate file _DownloadDataDictionaryInfo_ and is called in the _ActionsButtons_ component. This updates a state _dataDictionaryInfo_ with values for _noVariableMetadata_ and _dataDictionaries_ which is passed down as a parameter to the _DownloadVariableMetadata_ function. A unit test has been written for the new file and the unit tests for the other files have been updated as needed.

src/Discovery/DiscoveryDetails/DataDownloadList/DataDownloadList.tsx
jarvisraymond-uchicago: This has been updated: the variable _showList_ has been renamed in the _DataDownloadList_ component to _resourceFieldValueIsValid_ to improve clarity.

src/Discovery/DiscoveryDetails/DataDownloadList/ActionButtons/ActionButtons.tsx
```diff
@@ -62,25 +81,39 @@ const ActionButtons = ({
...
+                  setDownloadStatus,
+                );
+              }}
+            >
+              Download <br />
+              Variable-Level Metadata
+            </Button>
+          </Col>
+        )}
+        {showDownloadStudyLevelMetadataButton && (
```
mfshao, line 111: when user has not logged in, this button should show as `Login to To Download Study-Level Metadata`, and redirect to user to the login page upon clicking, just like the download manifest button

src/Discovery/DiscoveryDetails/DataDownloadList/ActionButtons/ActionButtons.tsx
```diff
@@ -62,25 +81,39 @@ const ActionButtons = ({
...
-        {/*
-        <Col flex='1 0 auto'>
-          <Button className='discovery-action-bar-button'>
-            Download <br />
-            Variable-Level Metadata
-          </Button>
-        </Col>
-        */}
-        {showDownloadStudyLevelMetadataButtons && (
+        {showDownloadVariableMetadataButton && (
```
mfshao, line 89: when user has not logged in, **and if this study has a VLMD**, this button should show as `Login to To Download Variable-Level Metadata`, and redirect to user to the login page upon clicking

src/Discovery/DiscoveryDetails/DataDownloadList/DataDownloadList.tsx
```diff
@@ -16,6 +17,7 @@ interface DataDownloadListProps {
 }
 
 const DataDownloadList = ({
+  resourceFieldValueIsValid,
```
mfshao, line 20: I think you need to use this value to check if you can pass `sourceFieldData` into `ProcessData()` at https://github.com/uc-cdis/data-portal/pull/1476/files#diff-1dbfc504269445ba617c8cf99d89caf9865795dc6db82e0d0fa12b14313a8770R27 below. Because there is no guarantee that the `sourceFieldData` here will be an array, which will cause problems down the road in `ProcessData()` when it tries to process the input as an array

mfshao (commented):
deployed and tested in QA, looking good mostly, have 3 comments

src/Discovery/DiscoveryDetails/DataDownloadList/DataDownloadList.tsx
jarvisraymond-uchicago: That is a good point, the assignment to the const ```data``` has been updated to only process the data if the ```resourceFieldValueIsValid``` is true. 

```
  const data: [] = resourceFieldValueIsValid
    ? ProcessData(sourceFieldData)
    : [];
```

src/Discovery/DiscoveryDetails/DataDownloadList/ActionButtons/ActionButtons.tsx
jarvisraymond-uchicago: Per discussion with stakeholders Feb 22, will not implement this functionality—users will be able to download study metadata whether they are logged in or not.

mfshao (approved):
LGTM 👍 thanks, we can merge after CI passes

PR #1496, opened 2024-03-08
fix array handling

mfshao:
### Bug Fixes
- Discovery: make sure arrays that passed into list components are always single-level


jarvisraymond-uchicago (approved):
Looks good, testing running the code locally with QA Dev.

PR #1498, opened 2024-03-13
Feat/incorporate sort search highlight pagination

jarvisraymond-uchicago:
Jira Ticket: [VADC-851](https://ctds-planx.atlassian.net/browse/VADC-851)

Notes: 
* Table events such as opening & closing the dropdowns, updating the current page, changing the number of entries shown per page, changing the search term, and changing the sorting order are now all controlled by a function ```handleTableChange```

This allows for: 
* Maintaining open / closed dropdown across table events
* Resetting of the current page back to page one following searches, changes of the entries shown and sorting

output

### New Features
* Updates Atlas Data Dictionary to use an integrated approach to handling UI events, allowing different UI table updates to update each other. This improves UX by enabling pagination resets and avoids problems such as being on page 3 when there is no longer enough data to populate 3 pages. 

### Bug Fixes
* Fixes issue with entries header showing incorrect values
* Fixes stylelint issues
* Fixes issue with incorrect number of entries shown on the second page when there are more entries than one page can hold but less than three pages 

[VADC-851]: https://ctds-planx.atlassian.net/browse/VADC-851?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/Analysis/AtlasDataDictionary/AtlasDataDictionaryContainer.tsx
```diff
@@ -1,73 +1,123 @@
...
+        ...dataDictionaryTableState,
+        searchTerm: eventData,
+        currentPage: 1,
+      });
+    } else if (event === 'sortConfig') {
+      setDataDictionaryTableState({
+        ...dataDictionaryTableState,
+        sortConfig: eventData,
+        currentPage: 1,
+      });
```
pieterlukasse, line 89: can we perhaps combine these? They are doing the same thing here.

pieterlukasse (approved):
code looks good. Just a few minor questions

src/Analysis/AtlasDataDictionary/Components/TableRow/ValueSummaryChart/ValueSummaryChart.tsx
```diff
@@ -95,4 +95,4 @@ const ValueSummaryChart = ({
   );
 };
 
-export default ValueSummaryChart;
+export default React.memo(ValueSummaryChart);
```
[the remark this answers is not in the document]
jarvisraymond-uchicago: Correct, this is an optimization. This prevents redraws of the charts in the value summary column when the props for them _haven't changed_. For example, when the search term changes,  this prevents redraws of all the unchanged charts, potentially up to 30 on the page. Without this optimization all the unnecessary redraws makes using the search bar significantly slower. This also prevents redraws of the chart in the dropdown.

src/Analysis/AtlasDataDictionary/AtlasDataDictionaryContainer.tsx
jarvisraymond-uchicago: Great idea! Updated to use square brackets so the event variable is evaluated as the key of the key value pair assignment: 
```
} else if (event === 'entriesShown' || event === 'searchTerm' || event === 'sortConfig') {
      setDataDictionaryTableState({
        ...dataDictionaryTableState,
        [event]: eventData,
        currentPage: 1,
      });
```

PR #1508, opened 2024-03-22
(MIDRC-666): add guppy defaultFilter functionality

ocshawn:
Link to JIRA ticket if there is one: [MIDRC-666](https://ctds-planx.atlassian.net/browse/MIDRC-666)

### New Features
- add guppy defaultFilter functionality

[MIDRC-666]: https://ctds-planx.atlassian.net/browse/MIDRC-666?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/GuppyDataExplorer/GuppyDataExplorer.jsx
```diff
@@ -38,10 +39,32 @@ class GuppyDataExplorer extends React.Component {
...
+        }
+        return tabAccumulator;
+      }, {});
+    }
+
     this.state = {
       aggsData: {},
       filter: {},
-      initialFilterFromURL: initialFilter,
+      initialFilterFromURL: { ...defaultFilters, ...initialFilter },
```
paulineribeyre, line 67: i don't know if we even use `explorerStoreFilterInURL` anywhere - but it's raising a couple of questions
- if I send a URL with filters, should it include the default filters?
- if I open a URL with filters, should the default filters be applied on top of those?

paulineribeyre: if `explorerStoreFilterInURL` isn't used anywhere, maybe we can leave this as-is and document in the config doc that `explorerStoreFilterInURL` and `defaultFilters` don't work well together

paulineribeyre: we can fix it if we ever need to

mfshao: I don't think `explorerStoreFilterInURL` has been enabled anywhere in any existing project. But in terms of the relationships of `explorerStoreFilterInURL` and `defaultFilters`, I would propose:
1. if `explorerStoreFilterInURL` is enabled, `defaultFilters` will always be placed in the URL as part of the initial filters
2. if a page is opened by pasting an URL, then portal should add those `defaultFilters` when parsing the input if the input filters from URL doesn't have them, even if this will alter the URL

src/GuppyDataExplorer/GuppyDataExplorer.jsx
```diff
@@ -38,10 +39,32 @@ class GuppyDataExplorer extends React.Component {
...
+      }, []);
+
+      defaultFilters = defaultFilterArr.reduce((tabAccumulator, tabDefaultFilter) => {
+        if (tabDefaultFilter.field && tabDefaultFilter.values.length > 0) {
+          // if it already exists add values to it
+          if (tabAccumulator[tabDefaultFilter.field]) {
+            tabAccumulator[tabDefaultFilter.field].selectedValues.push(...tabDefaultFilter.values);
+            return tabAccumulator;
+          }
+          return { ...tabAccumulator, [tabDefaultFilter.field]: { selectedValues: tabDefaultFilter.values } };
```
mfshao, line 58: I think these logic makes sense to me, but I also think this implies that `tabDefaultFilter.values` is required for each objects in `defaultFilters` and they have to be arrays
But I don't think we should do the validation here, maybe we should update the portal config validation script in our GH actions to catch this?

src/GuppyDataExplorer/GuppyDataExplorer.jsx
paulineribeyre: I agree with #1 but not #2, i think we should allow users to send a URL for a cohort where default filters have been changed. Otherwise the cohort when you open the URL will be different from the cohort the user meant to share.

mfshao: Hmm I see. Yes that makes sense. I agree

ocshawn: so its good?, or should i change the code to make it more explicit of whats going on

paulineribeyre: i think we should update it to do this:
- if explorerStoreFilterInURL disabled: just check the defaultFilters
- if explorerStoreFilterInURL enabled:
  - add defaultFilters to the URL by default
  - when a user lands on the explorer page from a URL with filters, respect the URL filters only and do not check the defaultFilters

It's up to you @ocshawn and @mfshao if this can be done in this PR or if it's too much scope creep, in that case let's create a separate ticket

ocshawn: I think that's the right approach

src/GuppyDataExplorer/GuppyDataExplorer.jsx
ocshawn: the vales are required for setting the defaults, and must be in array so that the user can select multiple default values. This part of the code is added to handle the edge case of setting defaults for the same field in multiple tabs

Ill look into validating these new fields in the portal config validation script

paulineribeyre: https://github.com/uc-cdis/gen3utils/blob/master/gen3utils/gitops/gitops_validator.py

src/GuppyDataExplorer/GuppyDataExplorer.jsx
```diff
@@ -31,6 +32,7 @@ class GuppyDataExplorer extends React.Component {
           const isValidJSON = IsValidJSONString(decodedFilter);
           if (isValidJSON) {
             initialFilter = JSON.parse(decodedFilter).filter;
+            hasInitialFilter = true;
```
paulineribeyre, line 35: if we land on this page without url filters, is `isValidJSON` false? iirc an empty string is valid json 🤔

ocshawn: if its empty it wont pass the check for stateFromURL

PR #1513, opened 2024-04-03
VADC-940 Feat/vadc manage columns data dictionary

jarvisraymond-uchicago:
https://ctds-planx.atlassian.net/browse/VADC-940

**Dev Notes**
* Uses mantine V6 components with Typescript with unique instances of custom icons to aid in porting to frontend framework
* Updates column headers and table row logic and their tests to support hiding / showing selected columns 
* User settings are retained between sessions and reloads within the same browser via local storage
* A11y: Panel and controls are keyboard controllable 
* Screen shots: 
_Includes multiple refreshes to demonstrate local storage retaining user selections_
output

### New Features
* Adds column management to VADC Atlas Data Dictionary


src/Analysis/AtlasDataDictionary/Components/ColumnHeaders/ColumnHeaders.tsx
```diff
@@ -1,125 +1,154 @@
...
-      ))}
-    </tr>
-  </thead>
-);
+  return (
+    <thead className={'column-headers'} data-testid='column-headers'>
+      <tr>
+        {headerItems.map((item) => (
+          <React.Fragment key={item.headerKey}>
+            {item.show && (
```
pieterlukasse, line 136: maybe this could also be written as follows?
```suggestion
            {item.columnManagementData[item.headerKey] && (
```
Or am I overlooking something? Just wondering why this new `show` attribute was needed...

src/Analysis/AtlasDataDictionary/Components/ManageColumns/ManageColumns.tsx
```diff
@@ -0,0 +1,177 @@
...
+
+  const columnManagementReset = () => {
+    handleTableChange(
+      'columnManagementReset',
+      'columnManagementReset',
+    );
+    setShowNotification(true);
+  };
+
+  const columnControls = [
```
pieterlukasse, line 55: this list seems to repeat some of the same attributes of the items in the list in ColumnHeaders. Is there a way to have the list of columns for this table defined once and then perhaps refer to it here?

pieterlukasse (commented):
thanks @jarvisraymond-uchicago . I added two minor comments/questions. Please take a look.

src/Analysis/AtlasDataDictionary/Components/ColumnHeaders/ColumnHeaders.tsx
jarvisraymond-uchicago: This is a very good point. 
To prevent or enable and thus hide / show the columns this has been updated to: 
```          {columnManagementData[item.headerKey] && ( ```

The column management headerKeys have been updated to use the same naming convention as what is being used elsewhere in the application, such as for sorting.

src/Analysis/AtlasDataDictionary/Components/ManageColumns/ManageColumns.tsx
jarvisraymond-uchicago: This is a very good point. 

The columnControls have been moved to the utils folder and renamed ```ColumnItems```
This is now used by both the ManageColumns component and the ColumnHeaders component. The ```label``` key value pair is now computed in ManageColumns by stripping out the HTML elements used in the JSX for the ColumnHeaders components, which need line breaks in specific places to match the design. The ```ID``` key value pair is now ```headerKey``` for both components.

pieterlukasse (approved):

PR #1516, opened 2024-04-04
Fix/slmd button not clickable

jarvisraymond-uchicago:
https://ctds-planx.atlassian.net/browse/HP-1454

**Screen Shots**
Before:
image

After: 
image

### Bug Fixes
* Updates the logic for the study level metadata button so it is not disabled even when there is noData 

### Improvements
* Adds unit test to ensure study level metadata button does not render when the resourceInfo does not include
    the study metadata field name reference from the discovery config


mfshao (approved):
👍 nice to cover it with tests

PR #1532, opened 2024-05-08
bugfix(teamProjectStateManagementRefresh): Initial commit

jarvisraymond-uchicago:
https://ctds-planx.atlassian.net/browse/VADC-1134

**Developer Notes**
Before
output

After 
output

### Bug Fixes
* Removes the delay  that occurs when opening GWAS Results after switching teams, such as after Selecting team1, Open GWAS Results, wait 5s, Select team2, Open GWAS Results. Before you would see team1 results for first 5s or so

### DIFF
--- a/src/Analysis/GWASResults/Views/Home/Home.jsx
+++ b/src/Analysis/GWASResults/Views/Home/Home.jsx
@@ -1,11 +1,11 @@
 import React from 'react';
 import { Spin } from 'antd';
 import { useQuery } from 'react-query';
+import PropTypes from 'prop-types';
 import HomeTable from './HomeTable/HomeTable';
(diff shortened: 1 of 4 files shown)


src/Analysis/GWASResults/Views/Home/Home.jsx
```diff
@@ -32,4 +34,9 @@ const Home = () => {
     </React.Fragment>
   );
 };
+
+Home.propTypes = {
+  selectedTeamProject: PropTypes.string.isRequired,
```
pieterlukasse, line 39: I think storybook and test will need to be updated to contain this as well.

jarvisraymond-uchicago: Good catch, updated and confirmed unit tests and storybook are working now.

pieterlukasse (approved):

PR #1537, opened 2024-05-20
editing the portal nginx conf

EliseCastle23:
editing the regex pattern for file extensions, so it won't match for project ids that have a "." in them

### DIFF
--- a/nginx.conf
+++ b/nginx.conf
@@ -88,7 +88,7 @@ server {
     }
 
     # Any route containing a file extension (e.g. /devicesfile.js)
-    location ~ ^.+\.[^/]+$ {
+    location ~* \.(jpg|jpeg|png|gif|ico|css|js|pdf|txt|bmp|svg)$ {
       try_files $uri =404;
     }
 


nginx.conf
```diff
@@ -88,7 +88,7 @@ server {
     }
 
     # Any route containing a file extension (e.g. /devicesfile.js)
-    location ~ ^.+\.[^/]+$ {
+    location ~* \.(jpg|jpeg|png|gif|ico|css|js|pdf|txt|bmp|svg)$ {
```
paulineribeyre, line 91: I'm wondering if we could make this more generic

We could do `^.+\.[^/]{1,3}+$` which assumes file extensions have between 1 and 3 characters, but it only mitigates the problem - if we have a project ID that matches it we'll have the same problem again

`^.+\.[^\/]{1,3}+\/?$` same as above but it would recognize file extensions even when there's a trailing slash

Or maybe nginx's `try_files` could be used to say "try to find a file, if not try to load the page, if not return 404"? But i'm not sure i understand why `try_files $uri =404` doesn't work already - it should be able to load the project page just like [here](https://github.com/uc-cdis/data-portal/blob/416c180078b9f0ef18c61937af8ed63719780471/nginx.conf#L95-L98)

mfshao: Where can I test this issue?
Also It might be a good idea to specify some characters as reserved and cannot be used as part of project names. I thought we already have some rules like this but maybe I'm wrong

m0nhawk: It doesn't work this way, [`try_files`](https://nginx.org/en/docs/http/ngx_http_core_module.html#try_files) only checks if file exist, not if specific path is being routed by anything.

paulineribeyre: copying my question from slack - could we just remove the "file extension -> 404" block? if a file doesn't exist, we redirect to portal (`index.html`), which will redirect to 404 anyway?

paulineribeyre: > Also It might be a good idea to specify some characters as reserved and cannot be used as part of project names. I thought we already have some rules like this but maybe I'm wrong

Currently the only restriction that i know of is that program names can't contain `-`. I don't think we should add restrictions if we can avoid it

mfshao: > could we just remove the "file extension -> 404" block?
so basically just use `try_files $uri /index.html;` ? I think it worth a try 🤔

m0nhawk: Removing "file extension -> 404" won't work because then it will route `/bundle.js` and other files through portal, which cannot serve JS.

PR #1538, opened 2024-05-21
Remove the nginx routing regex pattern specific to file extensions

paulineribeyre:
Closes #1537

### Improvements
- Remove the nginx routing regex pattern specific to file extensions, so it won't match for project IDs that contain a "."

### DIFF
--- a/nginx.conf
+++ b/nginx.conf
@@ -88,9 +88,9 @@ server {
     }
 
     # Any route containing a file extension (e.g. /devicesfile.js)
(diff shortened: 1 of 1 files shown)


mfshao (approved):
👍

m0nhawk (approved):
👍 Yes, this will work, I was not correct [here](https://github.com/uc-cdis/data-portal/pull/1537/files#r1608551711).

PR #1545, opened 2024-05-31
use file manifest for batch download

mfshao:
### Bug Fixes
- Discovery: fixed a bug that button incorrectly disabled because of duplicated required IDP tags

### Improvements
- Discovery: use file manifest instead of study IDs for direct download


piotrsenkow (approved):
New method assembleFileManifest's logic is easy to follow. Refactoring changes in other files make sense to me. Approving PR.

PR #1549, opened 2024-06-06
(VADC-836): remove already selected covariates from list to eliminate ability to add multiple times

ocshawn:
Link to JIRA ticket if there is one: [VADC-836](https://ctds-planx.atlassian.net/browse/VADC-836)

### Bug Fixes
- remove already selected covariates from list to eliminate ability to add multiple times

[VADC-836]: https://ctds-planx.atlassian.net/browse/VADC-836?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/Analysis/GWASApp/Components/Covariates/ContinuousCovariates.jsx
```diff
@@ -25,7 +25,11 @@ const ContinuousCovariates = ({
     <React.Fragment>
       <div className='GWASUI-flexRow continuous-covariates'>
         <div className='continous-covariates-table'>
-          <Covariates selected={selected} handleSelect={setSelected} />
+          <Covariates
+            selected={selected}
+            handleSelect={setSelected}
+            selectedCovariates={outcome?.concept_id ? [outcome.concept_id, ...selectedCovariates.map((obj) => obj.concept_id)] : []}
```
pieterlukasse, line 31: would it be possible to use the already existing `selected` instead? If not, can you please add a comment explaining why not? And perhaps consider another name so `selectedCovariates` and `selected` are not easily mixed up?

pieterlukasse (commented):
thanks @ocshawn . I added a question.

src/Analysis/GWASApp/Components/Covariates/ContinuousCovariates.jsx
ocshawn: the selected is for what the user currently has selected in the list, my new variable is for what they already have selected and submitted i will try to think of a new better name for this. Probably `submittedCovariates` if that works for you

pieterlukasse: Ok, `submittedCovariates` sounds good to me.

pieterlukasse (approved):

PR #1555, opened 2024-06-20
(VADC-293): Refactor use of inline SVG to use SVG utils for Chart Icons in GWAS

ocshawn:
Link to JIRA ticket if there is one: [VADC-293](https://ctds-planx.atlassian.net/browse/VADC-293)

### Improvements
- Refactor use of inline SVG to use SVG utils for Chart Icons in GWAS


ocshawn:
old icon code does not work with storybook, look into why, may need to just drop this update

PR #1560, opened 2024-06-26
HP-1565 fix repo info overwrite

mfshao:
Link to JIRA ticket if there is one:  https://ctds-planx.atlassian.net/browse/HP-1565

### Bug Fixes
- Study reg: fixed a bug causing pre-defined data repository metadata got overwritten by user


george42-ctds (approved):
Looks good to me.

PR #1562, opened 2024-07-01
(VADC-1228): add access-denied page and flag to enable

ocshawn:
Link to JIRA ticket if there is one: [VADC-1228](https://ctds-planx.atlassian.net/browse/VADC-1228)

### New Features
- add access-denied page and flag to enable when user does not have access to any site resources 

[VADC-1228]: https://ctds-planx.atlassian.net/browse/VADC-1228?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #1564, opened 2024-07-01
(VADC-1228): add access-denied page and flag to enable

ocshawn:
Link to JIRA ticket if there is one: [VADC-1228](https://ctds-planx.atlassian.net/browse/VADC-1228)

### New Features
- add access-denied page and flag to enable when user does not have access to any site resources 

[VADC-1228]: https://ctds-planx.atlassian.net/browse/VADC-1228?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Login/ProtectedContent.jsx
+++ b/src/Login/ProtectedContent.jsx
@@ -223,9 +223,20 @@ class ProtectedContent extends React.Component {
    */
   checkSiteAccess = (initialState) => {
     // check for setup property
-    if (userAccessToSite?.enabled && initialState.user.authz) {
+    if (userAccessToSite?.enabled && initialState.user.resources) {
+      const checkIfUserHasPermission = (resources) => {
+        if (userAccessToSite?.userAccessIncludes?.length > 0) {
+          const permissionMatchs = userAccessToSite.userAccessIncludes.filter((path) => (
+            resources.includes(path)
+          ));
+          // check if any user permissions matched
+          return permissionMatchs.length === 0;
+        }
+        return false;
+      };
       // check user authorization
-      if (!Object.keys(initialState.user.authz).length > 0) {
+      if (initialState.user.resources.length === 0
+        || checkIfUserHasPermission(initialState.user.resources)) {
         const redirectPage = userAccessToSite.deniedPageURL || '/access-denied';
         // only allow user to see access denied page
         if (redirectPage !== this.props.match.path) {
--- a/docs/portal_config.md
+++ b/docs/portal_config.md
@@ -720,7 +720,8 @@ Below is an example, with inline comments describing what each JSON block config
   "userAccessToSite": { // optional: user must have access to a resorces to acess the site
     "enabled": true,// optional: enable ristricted access
     "noAccessMessage": "Access to this site requires special permission.",// optional: defaults to this value, first email addresses will be turned into mailto link if used
-    "deniedPageURL": "/access-denied"//optional: defaults to this value
+    "deniedPageURL": "/access-denied",//optional: defaults to this value
+    "userAccessIncludes": ["/argo", "/workspace"] //optional: defaults to any, otherwise must have access to one item in array
   },
   "connectSrcCSPWhitelist": [ // optional; Array of urls to add to the header CSP (Content-Security-Policy) connect-src 'self'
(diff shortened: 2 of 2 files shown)


src/Login/ProtectedContent.jsx
```diff
@@ -53,13 +53,16 @@ class ProtectedContent extends React.Component {
...
           .then(
-            () => this.checkLoginStatus(store, this.state)
-              .then((newState) => ((this.props.public) ? { ...newState, redirectTo: null } : this.checkQuizStatus(newState))) // don't redirect for public pages
-              .then((newState) => ((this.props.public) ? { ...newState, redirectTo: null } : this.checkApiToken(store, newState))),
-          ).then(
             (newState) => this.checkUseWarning(store, newState), // check for existence of cookie to popup Use Warning
           )
+          .then(
+            // check for site access if property set
+            (newState) => this.checkSiteAccess(newState),
```
mfshao, line 64: might better check for `userAccessToSite?.enabled` here so we don't need always call `this.checkSiteAccess()`

mfshao: And this feature doesn't really works for public pages right? If a page is public, it suppose to be accessible for all users. But if this feature is turned on, it will block user's access?

src/Login/ProtectedContent.jsx
```diff
@@ -215,6 +218,25 @@ class ProtectedContent extends React.Component {
...
 
+  /**
+   * Display access contol message if userAccessToSite property set and user does not have sufficient privileges
+   */
+  checkSiteAccess = (initialState) => {
+    const newState = { ...initialState };
+    // check for setup property
+    if (userAccessToSite?.enabled && newState.user.authz) {
+      // check user authorization
+      if (!Object.keys(newState.user.authz).length > 0) {
```
mfshao, line 229: I feel this is a very specific use case. If there are any `all_users_policies` or `anonymous_policies` specified in user.yaml, then this won't work?
Maybe better considering adding a specific policy to those user's who should have access, and reject if the user logged in doesn't have that policy

src/index.jsx
```diff
@@ -509,6 +510,12 @@ async function init() {
                     path='/not-found'
                     component={NotFound}
                   />
+                  { userAccessToSite?.enabled && (
```
mfshao, line 513: Might add this to all other portal bundles? Or add something in doc saying this feature is available for Gen3 bundle only

mfshao (commented):
Do we have a feature doc for this change? I feel this change will have some product-level impact once it got into master, and there are some things that should be better illustrated with a feature doc

mfshao:
maybe I should leave my comments within https://github.com/uc-cdis/data-portal/pull/1562 but I'm not sure which one will be merged

src/Login/ProtectedContent.jsx
ocshawn: this feature is to block access to all pages, there is no feature doc as they want it live by Friday. This PR is not to merge it to master but instead to a separate release branch for VA. i will get to your performance improvements

nmetokishlubsky (approved):

src/Login/ProtectedContent.jsx
ocshawn: @mfshao i think the logic should stay in the this.checkSiteAccess function, but i will refactor that function so it does not do any additional work if the feature is not enabled

src/Login/ProtectedContent.jsx
ocshawn: the current way works for this use case, additional features could easily be added if anyone else ever requests this functionality

src/index.jsx
ocshawn: ill add to the other index pages

src/Login/ProtectedContent.jsx
mfshao: Yeah my point was there might be alternative solutions that is more generic if we consider this as a product-level feature instead of a project-level feature

ocshawn: made this area a little more robust by adding userAccessIncludes option

PR #1572, opened 2024-07-23
Feat/vadc workflow dashboard

jarvisraymond-uchicago:
https://ctds-planx.atlassian.net/browse/VADC-1267

**Dev Notes**
Dashboard updates every 5 seconds using polling

Design: 
image

Implementation: 
In GWAS App: 
image

In Results App: 
image

In Storybook: 
output

### New Features
* Adds a dashboard to show the real time status of the user's workflow limit while using the Gen 3 GWAS App and Results App.


src/Analysis/SharedUtils/WorkflowLimitsDashboard/WorkflowLimitsUtils.js
```diff
@@ -0,0 +1,31 @@
...
+  }
+  // validate data contains expected keys and they're numeric
+  if (
+    typeof data?.workflow_run !== 'number'
+    || typeof data?.workflow_limit !== 'number'
+  ) {
+    return false;
+  }
+  return true;
+};
```
pieterlukasse, line 31: because of `workflowRun / workflowLimit` done in other part of the code, I think this function should also return `false` if `workflowLimit === 0`

pieterlukasse (commented):
one small comment

src/Analysis/SharedUtils/WorkflowLimitsDashboard/WorkflowLimitsUtils.js
jarvisraymond-uchicago: Good catch, function updated along with unit tests for this edge case. 
```
  // validate data contains expected keys and they're numeric
  // and workflow limit is greater than 0
  if (
    typeof data?.workflow_run !== 'number'
    || typeof data?.workflow_limit !== 'number'
    || !(data.workflow_limit > 0)
  ) {
    return false;
  }
  ```

pieterlukasse (approved):

PR #1573, opened 2024-07-30
Feat: login features

pieterlukasse:
https://ctds-planx.atlassian.net/browse/VADC-1286, 
https://ctds-planx.atlassian.net/browse/VADC-1228,  
https://ctds-planx.atlassian.net/browse/VADC-1221

### New Features
- add access-denied page and flag to enable when user does not have access to any site resources

### Improvements
- fix: use openid URL for Atlas to get automatic login behavior

### DIFF
--- a/docs/portal_config.md
+++ b/docs/portal_config.md
@@ -718,7 +718,7 @@ Below is an example, with inline comments describing what each JSON block config
       "service": "query_page"
     }
   },
-  "userAccessToSite": { // optional: user must have access to a resorces to acess the site
+  "userAccessToSite": { // optional: user must have access to a resorces to acess the site, including public pages
     "enabled": true,// optional: enable ristricted access
     "noAccessMessage": "Access to this site requires special permission.",// optional: defaults to this value, first email addresses will be turned into mailto link if used
     "deniedPageURL": "/access-denied",//optional: defaults to this value
--- a/docs/portal_config.md
+++ b/docs/portal_config.md
@@ -718,6 +718,12 @@ Below is an example, with inline comments describing what each JSON block config
       "service": "query_page"
     }
   },
+  "userAccessToSite": { // optional: user must have access to a resorces to acess the site
+    "enabled": true,// optional: enable ristricted access
+    "noAccessMessage": "Access to this site requires special permission.",// optional: defaults to this value, first email addresses will be turned into mailto link if used
+    "deniedPageURL": "/access-denied",//optional: defaults to this value
+    "userAccessIncludes": ["/argo", "/workspace"] //optional: defaults to any, otherwise must have access to one item in array
+  },
   "connectSrcCSPWhitelist": [ // optional; Array of urls to add to the header CSP (Content-Security-Policy) connect-src 'self'
     "https://example.s3.amazonaws.com" // full url to be added
   ],
--- a/src/Login/ProtectedContent.jsx
+++ b/src/Login/ProtectedContent.jsx
@@ -8,7 +8,7 @@ import {
 } from '../actions';
 import Spinner from '../components/Spinner';
 import getReduxStore from '../reduxStore';
-import { requiredCerts } from '../configs';
+import { requiredCerts, userAccessToSite } from '../configs';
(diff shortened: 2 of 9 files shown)


docs/portal_config.md
```diff
@@ -718,6 +718,12 @@ Below is an example, with inline comments describing what each JSON block config
       "service": "query_page"
     }
   },
+  "userAccessToSite": { // optional: user must have access to a resorces to acess the site
```
mfshao, line 721: shoudl add something here to explain that this access restriction feature takes precedence over public pages

src/Login/ProtectedContent.jsx
```diff
@@ -83,7 +86,8 @@ class ProtectedContent extends React.Component {
             },
           ),
       );
-    if (this.props.public) {
+    // if userAccessToSite is enabled do not load homepage until access is checked
+    if (this.props.public && !(this.props.match.path === '/' && userAccessToSite?.enabled)) {
```
mfshao, line 90: why check the path here? shouldn't it block access to any public pages if this feature is enabled?

docs/portal_config.md
mfshao: I think for now this is okay. But it might be debateable in the future if a site has both this option enabled and public pages, which should takes precedence

src/Login/ProtectedContent.jsx
ocshawn: good catch, this was done because this needs to run on the login page. Will reverse the logic to look for that page instead and update comment

docs/portal_config.md
ocshawn: updated

ocshawn (approved):
looks good to me

src/ecosystemIndex.jsx
```diff
@@ -583,6 +583,12 @@ async function init() {
                     path='/not-found'
                     component={NotFound}
                   />
+                  {userAccessToSite?.enabled && (
+                    <Route
+                      path={userAccessToSite.deniedPageURL || '/access-denied'}
+                      component={AccessDenied}
```
mfshao, line 589: you need to import `AccessDenied` and `userAccessToSite` before using them
this also applies to other bundle homepages

docs/portal_config.md
```diff
@@ -718,6 +718,12 @@ Below is an example, with inline comments describing what each JSON block config
       "service": "query_page"
     }
   },
+  "userAccessToSite": { // optional: user must have access to a resorces to acess the site, including public pages
+    "enabled": true,// optional: enable ristricted access
+    "noAccessMessage": "Access to this site requires special permission.",// optional: defaults to this value, first email addresses will be turned into mailto link if used
+    "deniedPageURL": "/access-denied",//optional: defaults to this value
+    "userAccessIncludes": ["/argo", "/workspace"] //optional: defaults to any, otherwise must have access to one item in array
```
mfshao, line 725: is this true? it seems from the logic of `checkIfUserHasPermission()`, it will returns false is no `userAccessIncludes` has been defined

mfshao (changes requested):

src/Login/ProtectedContent.jsx
```diff
@@ -215,6 +219,37 @@ class ProtectedContent extends React.Component {
     return newState;
   };
 
+  /**
+   * Display access contol message if userAccessToSite property set and user does not have sufficient privileges
+   */
+  checkSiteAccess = (initialState) => {
+    // check for setup property
+    if (userAccessToSite?.enabled && initialState.user.resources) {
+      const checkIfUserHasPermission = (resources) => {
```
mfshao, line 228: the name of this function is a bit confusing, so if it returns true it means user doesn't have premission and should be redirected, and false means user has premission
This is conter-intuitive, suggest to get it a better name

mfshao (approved):
👍

src/Login/ProtectedContent.jsx
ocshawn: changed the function name to userHasNoPermissions, need to redo the whole section to make it more readable

PR #1578, opened 2024-08-07
Feat/results app workflow limit error UI

jarvisraymond-uchicago:
https://ctds-planx.atlassian.net/browse/VADC-1269

**Dev Notes**
* Functionality can be tested using storybook, i.e. ```npm run storybook```
* Updates test data to make storybook easier to use
* Per discussion with stakeholder, uses AntD notification component instead of AntD Modal. This allows for display of a loading state, a consistent UI with other Action Dropdown events and multiple notifications when retrying multiple workflows simultaneously 

DESIGN
image

IMPLEMENTATION
output
output

### New Features
* Adds messaging for VADC Results app for workflow limits and prevents submission of retries if workflow limit service is unavailable, returns malformed data or if user's workflow limit is exceeded


tianj7 (approved):
tested and lgtm

PR #1580, opened 2024-08-13
Bugfix/fallback for null unprocessed chart data

jarvisraymond-uchicago:
https://ctds-planx.atlassian.net/browse/VADC-1300

**Screenshots**
Before fix VADC Atlas Data Dictionary application crashes due to it being unable to process null data for the preview chart. After the fix, the preview charts do not render: 
output

### Bug Fixes
* prevents Atlas Data Dictionary app from crashing when unProcessChartData input to processNumericChartData function is null.


jarvisraymond-uchicago:
Per discussion with @pieterlukasse, PR is good to merge.

pieterlukasse:
👍 ✅

PR #1590, opened 2024-08-28
Fix/aws s3 csp

mfshao:
### Improvements
- Generic: update CSP to allow both `virtual` and `path` addessing styled presigned url to work


mfshao:
no long needed becuase we implemented the fix from fence side https://github.com/uc-cdis/fence/pull/1179

PR #1592, opened 2024-08-29
Chore/change faro metadata

mfshao:
### Improvements
- Generic: Grafana Faro's `environment` value in its app metadata is set to the hostname of the env
- Generic: Grafana Faro's `namespace` value in its app metadata is set to the application’s namespace, for example: prod, pre-prod, staging
- Generic: Set CSP dynamically for faro enabled envs


jawadqur (approved):
Works fine in staging.midrc.org

PR #1593, opened 2024-09-06
bugfix(discoveryModalButtonsMediaQueries): Initial commit

jarvisraymond-uchicago:
https://ctds-planx.atlassian.net/browse/HP-1687

**Before**
study record permalink doesn't wrap

**After**
output

### Bug Fixes
* Adds media query for Discovery page so buttons are visible in discovery modal (drawer) at screen sizes below 1080px width


mfshao (approved):
👍

PR #1594, opened 2024-09-10
Feat/globalize vadc charts

jarvisraymond-uchicago:
https://ctds-planx.atlassian.net/browse/VADC-694

### New Features
* Globalizes data viz components (histogram and Simple3Sets ) for use between VADC apps

### Improvements
* In the case where the histogram bin data is missing (undefined) for the histogram, application now shows an error message. Before it would only show the error message when the value was null, and not when it is undefined (i.e. it doesn't exist at all in the sent data).


tianj7 (approved):
lgtm

PR #1596, opened 2024-09-17
 Bugfix/duplicate dichotomous covariates

jarvisraymond-uchicago:
https://ctds-planx.atlassian.net/browse/VADC-1363

**Dev Notes**
This bug fix prevents the user from adding duplicate covariates by both hiding already selected covariates from the end user and also by ensuring the reducer used for state management rejects covariates already stored as part of the shared state. 

**Defect**
User able to add duplicate covariates when using dichotomous outcome: 
output

**Fix**
User unable to add duplicate covariates when using dichotomous outcome: 
output

### Bug Fixes
* Prevents a user from submitting duplicate covariates for both continuous and dichotomous outcome types in the VADC Gen3 GWAS app


src/Analysis/GWASApp/Components/Covariates/ContinuousCovariates.jsx
```diff
@@ -21,14 +21,22 @@ const ContinuousCovariates = ({
...
+  // If they selected a outcome phenotype that is dichotomous
+  // the outcome doesn't need to be included as a submitted covariate
+  const submittedCovariateIds = outcome?.concept_id
+    ? [outcome.concept_id, ...selectedCovariates.map((obj) => obj.concept_id)]
+    : [...selectedCovariates.map((obj) => obj.concept_id)];
+
   return (
     <React.Fragment>
       <div className='GWASUI-flexRow continuous-covariates'>
         <div className='continous-covariates-table'>
```
tianj7, line 35: Actually it's here too, maybe we can get these updated?

src/Analysis/GWASApp/Components/Covariates/ContinuousCovariates.jsx
```diff
@@ -21,14 +21,22 @@ const ContinuousCovariates = ({
     concept_name: selected.concept_name,
   });
 
+  // when a user has selected a outcome phenotype that is a continous covariate with a concept ID,
```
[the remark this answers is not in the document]
jarvisraymond-uchicago: Good catch, thank you! Should be fixed now.

tianj7 (approved):

src/Analysis/GWASApp/Components/Covariates/ContinuousCovariates.jsx
jarvisraymond-uchicago: Good catch, thank you! Should be fixed now.

PR #1579, opened 2024-08-09
Feat/discovery load time spike

jarvisraymond-uchicago:
# Spike for Improve Discovery Page Loading UX

Ticket: https://ctds-planx.atlassian.net/browse/HP-1643

**Instructions:**
Pull local, then run in QA Dev. Three different UIs are available with batch loading: 

No Progress Bar: 
https://qa-heal.planx-pla.net/portal/dev.html

Mini Progress Bar by Studies Count: 
https://qa-heal.planx-pla.net/portal/dev.html?showMini

Big Progress Bar at top of Application: 
https://qa-heal.planx-pla.net/portal/dev.html?showBig

**Video showing all three**
https://github.com/user-attachments/assets/27eb5de4-5e8e-444b-ae19-e82fad7a5a57

PR #1599, opened 2024-09-26
Feat/discovery enhanced loading

jarvisraymond-uchicago:
https://ctds-planx.atlassian.net/browse/HP-1679

**Dev Notes**
* Implements feature doc defined as part of https://ctds-planx.atlassian.net/browse/HP-1643 and prototyped here: https://github.com/uc-cdis/data-portal/pull/1579     
    *  batch loading consists of an initial small batch of 10 studies, enabling the user to see and interact with the application faster. A secondary larger batch of studies will be downloaded and processed following it. 
    * Because HEAL uses `studyRegistration`, the smaller batch consists of two batches of 5 studies each from `loadStudiesFromAggMDS` and `getSomeStudiesFromMDS`

**UX Before Update** 
Load time is approximately 5 seconds. Searches entered during loading do not update once loading is completed. 
output

**UX After Update**
UX now uses an initial batch loading of 10 studies and followed by loading the rest. The initial load of first 10 studies takes around 200-300ms (95% improvement). UX shown includes a mini progress bar. 
* Searches entered during loading update with all available results after loading is completed. 
* Filters and Data Repository selection is disabled until loading is completed. 
output

### New Features
* Provides an enhanced page loading experience for Discovery page using batch loading

### Improvements
* Adds unit tests for MDSUtils and aggMDSUtils


src/Discovery/index.tsx
```diff
@@ -167,16 +200,22 @@ const DiscoveryWithMDSBackend: React.FC<{
...
+  }, [props, numberOfBatchesLoaded, numberOfStudiesForSmallerBatch]);
 
   let studyRegistrationValidationField = studyRegistrationConfig?.studyRegistrationValidationField;
   if (!isEnabled('studyRegistration')) {
     studyRegistrationValidationField = undefined;
   }
+
+  const allBatchesAreLoaded = studies === null
+    ? false
+    : (studies as Array<any>)?.length > totalNumberOfStudiesFromSmallBatches;
```
mfshao, line 212: better to check if `numberOfBatchesLoaded === expectedNumberOfTotalBatches`? since the current logic won't work if the total number of studies in MDS is less then `totalNumberOfStudiesFromSmallBatches` ?
although if that is the case we probably will disable this feature, but just to be on the safe side

src/Discovery/index.tsx
```diff
@@ -73,21 +72,55 @@ const DiscoveryWithMDSBackend: React.FC<{
...
+  const [numberOfBatchesLoaded, setNumberOfBatchesLoaded] = useState(0);
+  const expectedNumberOfTotalBatches = 2;
+  // if loading with both unregistered and registered studies, load 5 for each (10 total)
+  const numberOfStudiesForSmallerBatch = isEnabled('studyRegistration') ? 5 : 10;
+  const numberOfSmallBatchesToBeLoaded = isEnabled('studyRegistration') ? 2 : 1;
+  const totalNumberOfStudiesFromSmallBatches = numberOfStudiesForSmallerBatch * numberOfSmallBatchesToBeLoaded;
+  const numberOfStudiesForAllStudiesBatch = 2000;
+
   useEffect(() => {
+    // If batch loading is Enabled, update the numberOfBatchesLoaded to enable calling of different batch sizes
```
mfshao, line 88: is there a way to disable batch loading?

src/Discovery/Utils/MDSUtils/MDSUtils.jsx
```diff
@@ -43,4 +43,29 @@ const loadStudiesFromMDS = async (guidType = 'discovery_metadata') => {
   }
 };
 
-export default loadStudiesFromMDS;
+export const getSomeStudiesFromMDS = async (guidType = 'discovery_metadata', numberOfStudies = 10) => {
```
mfshao, line 46: I don't think we should re-write this function since a good chunk of its logic already exists in `loadStudiesFromMDS()` above
we should be able to make `loadStudiesFromMDS()` to take `LIMIT` as an argument, something like
```
export const loadStudiesFromMDS = async (guidType = 'discovery_metadata', limit=2000)
```
and if we need partial loading, just call this function with a smaller `limit` value

src/Discovery/index.tsx
jarvisraymond-uchicago: I don't believe this will work - the proposed approach sets the allBatchesAreLoaded variable to true before allowing the browser to process all the studies (such as with the function `populateStudiesWithConfigInfo`). This can take a couple seconds due to the number of studies. Setting this variable true before this has completed causes the progress bar to finish too soon.

However to improve clarity I have updated to the variable name to:
`allBatchesAreReady`

src/Discovery/index.tsx
jarvisraymond-uchicago: Good catch. That was initially scoped, but then we decided to have this functionality for all instances of the discovery page. The comment has been updated to: 

`    // Update the numberOfBatchesLoaded to enable calling of different batch sizes with different parameters`

src/Discovery/Utils/MDSUtils/MDSUtils.jsx
jarvisraymond-uchicago: Good idea! Per our discussion, the loadStudiesFromMDS() has been refactored to take in a numeric `fetchSize` as a parameter and a boolean `loadAllMetadata`, both with default assignments. 

When called with `loadAllMetadata === false` the while loop exits after a single fetch. The calls to `getSomeStudiesFromMDS` have been replaced to calls with `loadAllMetadata` with additional parameters. The unit tests for MDSUtils have been updated accordingly.

mfshao (approved):
code -wise i think it lgtm
need to figure out why some CI tests are failing 🤔

PR #1608, opened 2024-10-22
fix(downloadVLMDButtonReset): Initial commit

jarvisraymond-uchicago:
https://ctds-planx.atlassian.net/browse/HP-1759

**Dev Notes**
_Before_
State is incorrectly retained for VLMD button after choosing a differet study; a study with VLMD available incorrectly makes other studies show that they have VLMD. 

output

_After_
State is updated for VLMD button after choosing a different study; one study's VLMD availability does not impact other studies. 
output

### Bug Fixes
Resets noVariableLevelMetadata when choosing different studies so VLMD button behaves as expected.


mfshao:
code wis lgtm, the failed unit tests is because an env has been removed, so this should fix it https://github.com/uc-cdis/data-portal/pull/1609

PR #1611, opened 2024-10-23
Bugfix/combo box labels

jarvisraymond-uchicago:
https://ctds-planx.atlassian.net/browse/VADC-1443

**Before**
Markup for AntD combobox component is missing aria-labels 
image

**After**
AntD combobox replaced with select element with a label element
image

### Bug Fixes
Replaces non-compliant AntD combo box component with 508 compliant TeamDropdown component using a labeled HTML select element.


cemile-ctds (commented):
Labels should be concise and clear for all users. The term "Combo Box" is typically used in programming and UI development, and not all end users may be familiar with it. Please update the label to "Select Team Project" or "Team Project Selection," as these options are more straightforward.

PR #1613, opened 2024-10-23
Feat: enable old and new data dictionary for VA

pieterlukasse:
Link to JIRA ticket if there is one:  https://ctds-planx.atlassian.net/browse/VADC-1519

### Improvements
- new logic to allow for coexistence of old and new VA data dictionary apps.
Both apps are available under the same /analysis/AtlasDataDictionary location.
- removed unnecessary `ProtectedContent` wrapper for one of the Data dictionary options since /analysis wrapper is already protected (see https://github.com/uc-cdis/data-portal/blob/7d03acf24c498f57748d670773790b834412d3f4/src/index.jsx#L274-L291). 

### Deployment changes
- new config option `dataDictionaryVersion` that can be set to "new" or something else. If "new", then the new data dictionary component is loaded. This is a temporary solution until we have fully deprecated the old one.

### DIFF
--- a/src/Analysis/AnalysisApp.jsx
+++ b/src/Analysis/AnalysisApp.jsx
@@ -138,7 +138,9 @@ class AnalysisApp extends React.Component {
     case 'AtlasDataDictionary': {
       return (
         <div className='analysis-app_flex_row'>
-          <AtlasDataDictionaryContainer />
+          <AtlasDataDictionaryContainer
+            dataDictionaryVersion={analysisApps[app].dataDictionaryVersion}
+          />
         </div>
       );
     }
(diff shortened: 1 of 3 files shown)


src/Analysis/AnalysisApp.jsx
```diff
@@ -138,7 +138,9 @@ class AnalysisApp extends React.Component {
     case 'AtlasDataDictionary': {
       return (
         <div className='analysis-app_flex_row'>
-          <AtlasDataDictionaryContainer />
+          <AtlasDataDictionaryContainer
+            title={analysisApps[app].dataDictionaryVersion}
```
jarvisraymond-uchicago, line 142: This is the wrong prop name. I believe this needs to be `dataDictionaryVersion`. I was unable to get this to work locally with `title` as the prop name.

jarvisraymond-uchicago (commented):
Please run linter, I'm seeing this error: 

AtlasDataDictionaryContainer.tsx
  35:3  error  propType "dataDictionaryVersion" is not required, but has no corresponding defaultProps declaration  react/require-default-props
  
  and
  
  Unnecessary 'else' after 'return'.eslint[no-else-return](https://eslint.org/docs/rules/no-else-return)
  
  
 Also flagged an issue with prop name `title`

m0nhawk (approved):

jarvisraymond-uchicago (approved):
Looks good to me.

PR #1617, opened 2024-10-25
(VADC-1444): add please login to access text to login page

ocshawn:
Link to JIRA ticket if there is one: [VADC-1444](https://ctds-planx.atlassian.net/browse/VADC-1444)

### New Features
- add please login to access text to login page

Screenshot 2024-10-30 at 4 47 04 PM

[VADC-1444]: https://ctds-planx.atlassian.net/browse/VADC-1444?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


craigrbarnes (dismissed):

cemile-ctds (changes requested):
Please change the title and description to the text below:
ACCESS {{PAGE NAME}}
The {{Page name}} requires access. Please login using one of the options blow to continue.

Screenshot 2024-10-29 at 16 59 49

cemile-ctds (approved):
Looks good. Thank you Shawn.

src/Login/Login.jsx
```diff
@@ -96,6 +96,21 @@ class Login extends React.Component {
...
+      if (location.state && location.state.from) {
+        next = location.state.from;
+      }
+      if (!next) {
+        return undefined;
+      }
+      // Lookup next to get actuale page name, if item is not in main navigation display default messaging
+      const nextItem = components.navigation.items.filter((item) => item.link === next)[0];
+
+      return nextItem ? nextItem.name : 'Restricted';
```
mfshao, line 110: This doesn't take care of homepage right? If logging in from homepage it will show as `Restricted` since `/` is not in `navigation.items`? And the message will just say `The page requires access`?

ocshawn: I will look into, i was not aware that user can login from homepage. I think it should probably just show the default login page in this instance what are your thoughts? 

Do you know which environment has the homepage login? otherwise i will find it

mfshao: If you check on how `indexPublic` is being set in config that should work...
But you are right, it was implenent as a way of working with a very old Penegrine
Probably no body is using that any more and no need to worry

mfshao: I think taht is an edge case we can drop, I'll approve this

mfshao (approved):

src/Login/Login.jsx
ocshawn: I fixed, data.kidsfirstdrc.org still uses it and i made the change please review again

mfshao (approved):

PR #1631, opened 2024-11-04
Fix: improve error message for Euler diagram

pieterlukasse:
Link to JIRA ticket if there is one: https://ctds-planx.atlassian.net/browse/VADC-1490

### Improvements
- improve error message for Euler diagram, replacing it with a more generic error message that applies better to all possible error scenarios


jarvisraymond-uchicago (approved):
Looks good to me!

PR #1632, opened 2024-11-08
fix discovery batch loading

mfshao:
### Bug Fixes
- Discovery: fixed a bug causing the discovery page cannot correctly loads metadata in batches if neither AggMDS or StudyRegsistration has been enabled

### DIFF
--- a/src/Discovery/index.tsx
+++ b/src/Discovery/index.tsx
@@ -86,7 +86,11 @@ const DiscoveryWithMDSBackend: React.FC<{
 
   useEffect(() => {
     // Update the numberOfBatchesLoaded to enable calling of different batch sizes with different parameters
-    if (numberOfBatchesLoaded < expectedNumberOfTotalBatches) setNumberOfBatchesLoaded(numberOfBatchesLoaded + 1);
+    if (numberOfBatchesLoaded < expectedNumberOfTotalBatches) {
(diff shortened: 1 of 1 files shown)


src/Discovery/index.tsx
```diff
@@ -99,7 +99,8 @@ const DiscoveryWithMDSBackend: React.FC<{
           : numberOfStudiesForAllStudiesBatch);
       } else {
         loadStudiesFunction = loadStudiesFromMDS;
-        loadStudiesParameters = [props.config?.features?.guidType, 10, false];
+        loadStudiesParameters = (numberOfBatchesLoaded === 1
+          ? [props.config?.features?.guidType, 10, false] : [props.config?.features?.guidType, 2000, true]);
```
nss10, line 103: Correct me if I misunderstood. In line 80, we have `expectedNumberOfTotalBatches` set to 2, and here we have a ternary which checks if `numberOfBatchesLoaded` === 1, but since we have line 89, which doesn't let us ever reach `numberOfBatchesLoaded` as `1`, the true condition never works. Right? Excuse me if i'm hugely mistaken 😅

mfshao: `numberOfBatchesLoaded` will reach 1 because of L89, but you are right, this logic can be improved to save an unncecssary loop cycle. I'll update

nss10 (approved):

PR #1636, opened 2024-11-13
(VADC-1448): add logout warning

ocshawn:
Link to JIRA ticket if there is one: [VADC-1448](https://ctds-planx.atlassian.net/browse/VADC-1448)

### New Features
- add logout warning to session manager 

Screenshot 2024-11-13 at 3 00 09 PM

[VADC-1448]: https://ctds-planx.atlassian.net/browse/VADC-1448?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/SessionMonitor/index.js
+++ b/src/SessionMonitor/index.js
@@ -24,7 +24,9 @@ export class SessionMonitor {
     this.mostRecentSessionRefreshTimestamp = Date.now();
     this.mostRecentActivityTimestamp = Date.now();
     this.interval = null;
-    this.popupShown = false;
+    this.logoutPopupShown = false;
+    this.inactivityWarningPopupShown = false;
+    this.inactiveWarningTime = 5 * 60 * 1000; // 5 min
   }
 
   start() {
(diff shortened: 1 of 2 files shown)


src/SessionMonitor/index.js
```diff
@@ -49,18 +51,38 @@ export class SessionMonitor {
...
+
+    getReduxStore().then((store) => {
+      store.dispatch({
+        type: 'UPDATE_POPUP',
+        data: {
+          inactivityWarningPopup: true,
+          inactivityWarningTime: warnTime,
+        },
+      });
+      this.inactivityWarningPopupShown = true;
```
mfshao, line 78: could both warning popup and logout popup be visible at the smae time and interference with each other?
if so maybe dismiss warning popup when the logout popup is activated?

ocshawn: In the src/Popup/ReduxAuthTimeoutPopup.jsx code i have an if else statement insuring that the logged out popup always has priority over the warning and that they are never shown at the same time

ocshawn: well i did but i think the code linter removed it (will look into why)

ocshawn: it removed it because of the return no need for an else statement because second if will never run if first is true

mfshao (approved):

PR #1639, opened 2024-11-15
Feat/spike ant d strategy

jarvisraymond-uchicago:
https://ctds-planx.atlassian.net/browse/VADC-1400

## Draft PR for Discussion Purposes

Prototype of Tech Approaches for AntD Remediation. 
- Creates prototype library AntDAccessibilityFixes
  - Consists of a variety of functions to inject aria labels and/or otherwise manipulate non-508 compliant markup created by AntD Library
- Alternative approach to recreating AntD components from scratch or porting VADC applications to use library that is accessible out-of-box (both approaches likely exceeding a year of dev time)
- This prototype adds aria labels to AntD table radio buttons, but the library can be expanded using a similar approach  to add and remove attributes such as aria-labels, titles and summaries. 
- Uses descriptive functions to handle the unique requirements for each ticket
  - helps organize and separate logic for 508 compliance
- Trade Offs: DOM manipulation should probably be kept as minimal while meeting compliance standards to reduce complexity and maintainability issues. 
  - i.e. adding one aria label vs. adding a wrapping `<label>` element, unique `id` attributes, `for` attributes referencing other DOM nodes etc.

### DIFF
--- a/docs/portal_config.md
+++ b/docs/portal_config.md
@@ -204,6 +204,7 @@ Below is an example, with inline comments describing what each JSON block config
     "studyRegistration": true, // optional, whether to enable the study registration feature
     "workspaceRegistration": true, // optional, whether to enable the workspace registration feature
     "workspaceTokenServiceRefreshTokenAtLogin": true, // optional, whether to refresh the WTS token directly at portal login (recommended mode). If not set, this refresh happens only when the user enters the workspace section of the portal (default/old/previous mode).
+    "legacyVADCDataDictionary": false, // optional, VADC specific. Set to "true" to ensure the new version of the /analysis/AtlasDataDictionary data dictionary is displayed the user navigates to /analysis/AtlasDataDictionary, and when the user clicks on the data dictionary button in when in the Atlas app page.
   },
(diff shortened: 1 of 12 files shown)


jarvisraymond-uchicago:
> maybe we can try to update to `antd` v5, might be easier than migrating everyting to `mantine`

Interesting idea! Do you know if antd v5 is 508 compliant?

mfshao:
> > maybe we can try to update to `antd` v5, might be easier than migrating everyting to `mantine`
> 
> Interesting idea! Do you know if antd v5 is 508 compliant?

I'm not sure, and tbh I don't have a high hope on that. But maybe it will make our patching easier (although it can also goes to the other way)
Franking I think replacing with something like mantine might be the best solution. But that level of effort makes me feeling uncertain

jarvisraymond-uchicago:
Investigate RC option (building block of AntD)

https://github.com/react-component/table

jarvisraymond-uchicago:
Closing PR. Decision has been made to instead port Data Portal code of Gen3.2.

PR #1642, opened 2024-11-19
ARIA dialog and alertdialog nodes should have an accessible name

tianj7:
https://ctds-planx.atlassian.net/browse/VADC-1419

### Improvements
Add aria label for popup message

### DIFF
--- a/src/components/Popup.jsx
+++ b/src/components/Popup.jsx
@@ -64,7 +64,7 @@ class Popup extends React.Component {
               )
             }
           </div>
-          <div className='popup__message' id='popup__message'>
+          <div className='popup__message' id='popup__message' aria-labelledby='popup__message'>
             { this.props.message && <div className='high-light'>{this.props.message.map((text, i) => <p key={i}>{text}</p>)}</div> }
             {
               this.props.lines.length > 0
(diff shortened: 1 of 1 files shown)


ocshawn (commented):
unsure what this is accomplishing as it looks like you are adding a pointer that points to itself

tianj7:
> unsure what this is accomplishing as it looks like you are adding a pointer that points to itself

Updated code

src/components/Popup.jsx
```diff
@@ -100,6 +100,7 @@ class Popup extends React.Component {
                       enabled={(btn.enabled !== undefined) ? btn.enabled : true}
                       buttonType='default'
                       value={btn.value}
+                      aria-labelledby='popup__message'
```
jarvisraymond-uchicago, line 103: @tianj7 could you please touch base with @cemile-ctds regarding the business requirements? Usually the aria-labelledby acts by referencing the id of another element that provides the label text for screen readers. I believe if we have aria-labelledby we would not need the aria-labels you have added. Also make sure the element with the id `popup_message` is the correct element to act as the label.

jarvisraymond-uchicago (changes requested):

PR #1644, opened 2024-11-20
Fix/radio buttons508

jarvisraymond-uchicago:
https://ctds-planx.atlassian.net/browse/VADC-1457
https://ctds-planx.atlassian.net/browse/VADC-1461

**DEV NOTES**
- Uses a more complex instantiation of AntD tables so aria label can be applied to Radio buttons in Cohort select and concept (outcome and covariates) select tables 
- Before (aria labels missing from radio buttons): 

- After (aria labels added aria labels)
### Bug Fixes
- Improves 508 compliance


vzpgb (approved):
Looks good to me, I see area labels for GWAS Step 1 and GWAS Step 2 radio buttons. Thank you!

cemile-ctds (approved):

PR #1647, opened 2024-11-22
Fix/aria selected attribute removal

jarvisraymond-uchicago:
https://ctds-planx.atlassian.net/browse/VADC-1462

**Dev Notes**
- Creates library `AntDAccessibilityFixes`, which will include other programmatic fixes for AntD accessibility issues going forward
- **Description of the problem**: When an end user toggles the AntD table pagination component button (the button that shows 10 / page by default) either by click or keyboard interaction to open the dropdown, AntD creates new DOM elements to show the end user their pagination options (20 / page, 50 / page etc). However within these DOM elements the elements with class name `.ant-select-item` have the attribute `aria-selected`. This is identified as an error in axe accessibility checker. 
- **Solution**: This adds a new function `AntDTableAccessibilityFix` that is now called once the data for the AntD table is available and this calls the function `FixPaginationDropdown` after a delay allowing everything time to render. This function adds an event handler to AntD table pagination component button. When the end user open the dropdown and new DOM elements are created by AntD, a secondary script runs to remove all attributes of type `aria-selected` from the elements with class name `.ant-select-item`. Following this axe accessibility no longer identities errors for the `aria-selected` attributes. 
- **How to duplicate the defect**: First install the [axe accessibility extension](https://chromewebstore.google.com/detail/axe-devtools-web-accessib/lhdoppojpmngadmnindnejefpokejbdd). 
&ensp;1. Load the view containing the table
&ensp;2. Click start new scan
&ensp;3. Click the dropdown with text `10 / page` or navigate to using your keyboard with `tab` and then trigger it with `return` or spacebar.
&ensp;3. Then with that open click run the axe full page scan button
&ensp;4. The errors show up under the heading: _Elements must only use supported ARIA attributes_
- **Before**: axe finds 5 errors related to `Elements must only use supported ARIA attributes` and  attribute `aria-selected`

https://github.com/user-attachments/assets/2e25b193-57f0-406d-b2e8-827469cf3e38

- **After**: axe does not find errors related `Elements must only use supported ARIA attributes` and attribute `aria-selected`

https://github.com/user-attachments/assets/0777ed96-e45f-46fc-9777-9aaa9b43acd5

### Bug Fixes
- Improves accessibility

### DIFF
--- a/src/Analysis/GWASApp/Components/Covariates/Covariates.jsx
+++ b/src/Analysis/GWASApp/Components/Covariates/Covariates.jsx
@@ -7,6 +7,7 @@ import queryConfig from '../../../SharedUtils/QueryConfig';
 import { useFetch, useFilter } from '../../Utils/formHooks';
 import { useSourceContext } from '../../Utils/Source';
 import SearchBar from '../SearchBar/SearchBar';
+import { AntDTableAccessibilityFix } from '../../../SharedUtils/AccessibilityUtils/AntDAccessibilityFixes';
 
 const Covariates = ({ selected, handleSelect, submittedCovariateIds }) => {
   const { source } = useSourceContext();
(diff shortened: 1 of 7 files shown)


jarvisraymond-uchicago:
Closing PR. Decision has been made to instead port Data Portal code to Gen3.2 as of Jan 2025.

PR #1657, opened 2024-12-04
HP-1816 update batch export result texts

mfshao:
Link to JIRA ticket if there is one: https://ctds-planx.atlassian.net/browse/HP-1816
Screenshot 2024-12-04 at 3 28 25 PM

### Improvements
- Discovery: display link with title for batch export result


george42-ctds (approved):
Looks good. The correct short link was shown when running portal locally against QA_HEAL.

PR #1658, opened 2024-12-05
update cookie

mfshao:
Link to JIRA ticket if there is one: https://ctds-planx.atlassian.net/browse/PPS-1605

### Dependency updates
- update `cookie`

### DIFF
--- a/.storybook/main.js
+++ b/.storybook/main.js
@@ -3,10 +3,6 @@ module.exports = {
       "../src/**/*.stories.mdx",
       "../src/**/*.stories.@(js|jsx|ts|tsx)"
     ],
-    "addons": [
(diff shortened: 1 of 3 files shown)


AlbertSnows (approved):

.storybook/main.js
```diff
@@ -3,10 +3,6 @@ module.exports = {
       "../src/**/*.stories.mdx",
       "../src/**/*.stories.@(js|jsx|ts|tsx)"
     ],
-    "addons": [
```
AlbertSnows, line 6: Why are these being removed? Just double checking.

mfshao: I don't think we ever needed them. So would be better simply removing them

PR #1659, opened 2024-12-10
Create list from dropdown

k-burt-uch:

### Bug Fixes
- Create list from Discovery dropdown.


craigrbarnes (approved):
That should fix it.

PR #1670, opened 2025-02-10
MIDRC-921: disable "download manifest" for more terms than the ES limit

paulineribeyre:
Link to JIRA ticket if there is one: https://ctds-planx.atlassian.net/browse/MIDRC-921

### Improvements
- Explorer page: Do not allow users to click "download manifest" button when the query to ElasticSearch would fail due to having too many terms in the query. Instead, let the user know: "too many rows to download at once; make a smaller selection"


ocshawn (dismissed):
looks good

frankliuao (approved):
This will resolve the issue of forever-bouncing ••• in the portal, that MIDRC is facing now.

PR #1676, opened 2025-02-19
Fix/discovery load order wrong

k-burt-uch:

### Bug Fixes
- Fix issue on Discovery where the smaller batch for studies overwrites the larger batch, causing studies to not appear on Discovery.

### DIFF
--- a/src/Discovery/index.tsx
+++ b/src/Discovery/index.tsx
@@ -46,6 +46,8 @@ const populateStudiesWithConfigInfo = (studies, config) => {
   });
 };
 
+let allStudies;
+
 const DiscoveryWithMDSBackend: React.FC<{
     userAggregateAuthMappings: any,
     config: DiscoveryConfig,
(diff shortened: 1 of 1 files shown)


src/Discovery/index.tsx
k-burt-uch, line 1: I am open to other ideas.

mfshao: I think this can also be done at https://github.com/uc-cdis/data-portal/blob/70794fcfbda405048858c7b0fc9988f3b211a9d7/src/Discovery/index.tsx#L140 by adding some checks there to compare the length of `rawStudies` and `studies`. If `studies` already exists and the length of  `rawStudies` is less then `studies`, means the larger batch has already been processed and we should stop processing further but to return instead.
This way we can potentially skip some steps downstream.

mfshao (dismissed):

k-burt-uch:
This seems to be fixed in current portal. Closing

mfshao (approved):

PR #1677, opened 2025-02-25
fix use fractional height value

mfshao:
Link to JIRA ticket if there is one: https://ctds-planx.atlassian.net/browse/HP-1959

### Improvements
- use fractional height value for external footer


jarvisraymond-uchicago (approved):
Looks good to me!

mfshao:
the failed ci test is not related to the change in this pr

PR #1679, opened 2025-03-05
HP-1956 fix: discovery table checkbox enabled bug

mfshao:
Link to JIRA ticket if there is one: https://ctds-planx.atlassian.net/browse/HP-1956

### Bug Fixes
- Discovery: fixed a bug causing checkbox in discovery table gets enabled incorrectly for studies doesn't have file manifest and external file metadata


jarvisraymond-uchicago (approved):
Looks good!

* Pulled branch local and confirmed that fix disables checkbox per requirements, and was able to reproduce bug in QA. 
* New logic is easier to understand

PR #1683, opened 2025-03-28
Fix "Switch to Graph/Flat model" button

paulineribeyre:
### Bug Fixes
- The "Switch to Graph/Flat model" button on the Query page is now displayed when using the `explorerConfig` setting instead of the deprecated `dataExplorerConfig` setting in the data-portal configuration

### Improvements
- The data-portal configuration now describes the `explorerConfig` setting instead of the deprecated `dataExplorerConfig` and `fileExplorerConfig` settings

### DIFF
--- a/.github/workflows/integration_tests.yaml
+++ b/.github/workflows/integration_tests.yaml
@@ -9,7 +9,7 @@ concurrency:
 jobs:
   integration_tests:
     name: Integration tests
-    uses: uc-cdis/.github/.github/workflows/integration_tests.yaml@fix/md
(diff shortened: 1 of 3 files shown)


SpencerAxelrod:
This brought back the graph/flat model button with `staging.midrc.org` testing with portal locally

nss10 (approved):
Looks good to me. Tested the change in staging.midrc.org and is working as expected. Just to confirm, do we no longer have `FileExplorerConfig`? I see that being taken out of the readme completely.

PR #1685, opened 2025-03-31
Update bpa.json

FayBooker2:
Updates the bpa-support email address to bpa-support@gen3.org
Also, at the request of OCC and Lauren --> change all BloodPAC to BLOODPAC


mfshao:
this is a default config file shipped with portal for bloodpac, it is not being used by that env since the env is on gitops
I can still approve the change but just fyi

PR #1686, opened 2025-04-02
Feature/data library

k-burt-uch:

### New Features
- Adds data library list creation and management to Discovery


src/index.ejs
```diff
@@ -12,7 +12,7 @@
...
+    <meta http-equiv="Content-Security-Policy" content="default-src 'self'  http://localhost:8000 https://login.bionimbus.org https://wayf.incommonfederation.org; child-src blob:; img-src 'self' <%= htmlWebpackPlugin.options.imgSrc %> http://localhost:8000 https://opendata.datacommons.io https://static.planx-pla.net data: https://*.s3.amazonaws.com blob:; script-src 'self' 'unsafe-eval' <%= htmlWebpackPlugin.options.scriptSrc %>; worker-src 'self' blob:; style-src 'self' 'unsafe-inline' https://fonts.googleapis.com; object-src 'none'; font-src 'self' data: https://fonts.googleapis.com https://fonts.gstatic.com; connect-src 'self' https://login.bionimbus.org https://wayf.incommonfederation.org https://opendata.datacommons.io https://static.planx-pla.net https://*.zendesk.com <%= htmlWebpackPlugin.options.connectSrc %>; frame-src <%= htmlWebpackPlugin.options.connectSrc %> 'self' https://auspice.planx-pla.net https://auspice.pandemicresponsecommons.org https://*.quicksight.aws.amazon.com;">
```
mfshao, line 15: is this for local testing? if so I suggest moving them to `dev.html` and only use them from there

webpack.config.js
```diff
@@ -20,8 +20,8 @@ if (DAPTrackingURL) {
   connectSrcURLs.push(DAPTrackingURL);
 }
 if (gaTrackingId?.startsWith('UA-') || gaTrackingId?.startsWith('G-')) {
-  scriptSrcURLs.push(...['https://www.google-analytics.com', 'https://ssl.google-analytics.com', 'https://www.googletagmanager.com']);
-  connectSrcURLs.push(...['https://www.google-analytics.com', 'https://*.analytics.google.com', 'https://analytics.google.com', 'https://*.g.doubleclick.net']);
+  scriptSrcURLs.push(...['http://localhost:8000', 'http://www.google-analytics.com','https://www.google-analytics.com', 'https://ssl.google-analytics.com', 'https://www.googletagmanager.com']);
+  connectSrcURLs.push(...['http://localhost:8000', 'https://*.analytics.google.com', 'https://analytics.google.com', 'https://*.g.doubleclick.net']);
```
mfshao, line 24: I don't think you need to do them here with google analytics

src/Discovery/DiscoveryActionBar/DiscoveryDataLibrary.tsx
```diff
@@ -0,0 +1,160 @@
...
+        <ComboboxWithInput
+          items={data}
+          onChange={onChangeListSelection}
+          placeholder={notLoggedIn ? 'Login to save to a list' : 'Select/create a list to save to'}
+          disabled={notLoggedIn}
+        />
+        <Button
+          loading={loading}
+          type='primary'
+          disabled={error !== null || loading || data?.length === 0 || currentList === undefined || selectedResources.length === 0 || notLoggedIn}
```
SpencerAxelrod, line 144: `data?.length === 0` seems to prevent any initial list creation. It does not seem to increase when a user's first list is initially entered (after creation, but before a study is Saved to list), and subsequently can't be Saved to List as the button remains disabled. I have tested removing this and it resolves [this issue](https://ctds-planx.atlassian.net/browse/PXP-11460)

PR #1690, opened 2025-04-14
Feat/disable buttons when notebook access

jarvisraymond-uchicago:
https://ctds-planx.atlassian.net/browse/HP-1995

Following update,  the Download Zip, download all  and both Download Manifest buttons are disabled for HDP00167
output

### New Features
* Updates business logic so that for repositories with which we are interoperating via a workspace/notebook-based solution instead of indexing (e.g., Syracuse QDR and Harvard Dataverse), the Download Zip and Download Manifest buttons are be disabled, even when there is a value in the AuthZ field.


src/Discovery/Utils/isManifestDataMissing.ts
```diff
@@ -0,0 +1,5 @@
+import { DiscoveryResource } from '../Discovery';
+
+const isManifestDataMissing = (resource: DiscoveryResource) => !resource.__manifest || resource.__manifest?.length === 0;
```
mfshao, line 3: let's use the `exportToWorkspace.manifestFieldName` from discovery config instead of hard coding `__manifest` here

src/Discovery/DiscoveryActionBar/components/DownloadZipButton.tsx
```diff
@@ -3,89 +3,98 @@ import { DownloadOutlined } from '@ant-design/icons';
...
-          }
-          icon={<DownloadOutlined />}
-          loading={downloadStatus.inProgress}
+}) => {
+  const checkIfDownloadZipButtonDisabled = () => {
+    const noSelectedResources = props.discovery.selectedResources.length === 0;
+    const downloadInProgress = downloadStatus.inProgress;
+    const eachSelectedResourcesIsMissingManifest = props.discovery.selectedResources.every(
+      (resource: DiscoveryResource) => isManifestDataMissing(resource),
+    );
```
mfshao, line 17: nitpicking: I think we should be able to make this function return early if any of these values evaluate to true, so it can skip checking the truthness of values in the next

src/Discovery/Utils/isManifestDataMissing.ts
jarvisraymond-uchicago: Updated as requested, the function now takes parameter ```manifestFieldName``` and the files DownloadZipButton.tsx, DownloadManifestButton.tsx and ActionButtons.tsx have been updated to pass this parameter when calling ```isManifestDataMissing```

src/Discovery/DiscoveryActionBar/components/DownloadZipButton.tsx
jarvisraymond-uchicago: Both this function and the version of the function in DownloadManifestButton have been updated to return early if any of the values evaluate as true, otherwise they return ```false```

mfshao (approved):
👍

PR #1692, opened 2025-04-15
Add specific class to each login button

tianj7:
https://ctds-planx.atlassian.net/browse/DCF-2048

### Improvements
- Adding specifc classname to each login option to allow application of special styling if needed.

https://github.com/uc-cdis/cdis-manifest/pull/8592


k-burt-uch:
I don't see an equivalent feature that exists in portal that would accomplish DCF's goal of having the google login available but hiding it from portal, so this seems like a fine change. 

I'm curious how DCF will use the google authN   if the button in portal is hidden though 🤔

k-burt-uch (approved):

PR #1697, opened 2025-04-21
bump tar-fs version to 2.1.2 to address security vulnerability

tianj7:
https://ctds-planx.atlassian.net/browse/PPS-1982

### Dependency updates
bump tar-fs version to 2.1.2 to address security vulnerability

### DIFF
--- a/package.json
+++ b/package.json
@@ -156,7 +156,8 @@
       "strip-ansi": "6.0.1"
     },
     "@storybook/cli": {
-      "trim": "0.0.3"
+      "trim": "0.0.3",
+      "tar-fs": "^2.1.2"
     },
     "remark-parse": {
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -156,7 +156,8 @@
       "strip-ansi": "6.0.1"
     },
     "@storybook/cli": {
-      "trim": "0.0.3"
+      "trim": "0.0.3",
+      "tar-fs": "^2.1.2"
```
mfshao, line 160: do we really need this? I think do `npm update tar-fs` should update the dep in lock and satisfies the request
We should try not to manually set this override section unless we have to

mfshao:
addressed by https://github.com/uc-cdis/data-portal/pull/1693

PR #1700, opened 2025-05-06
fix for cdes that doesn't have a drupal ID

mfshao:

### Improvements
- The CDE pick list now can handle CDEs that don't have a drupal ID


george42-ctds (approved):
LGTM.

PR #1712, opened 2025-06-30
Feat/search bar enhancement search

jarvisraymond-uchicago:
https://ctds-planx.atlassian.net/browse/HP-2141

**Dev Notes**
Documentation changes are added to Portal Config
Unit test for Discovery updated to account for Search reinitializing
Note that UX will optimized in subsequent ticket
This is implemented such that if `searchableFields` are configured, search is enabled over only those fields, however if `searchableAndSelectableTextFields` is configured and fields are selected, search is enabled over only those fields.

Implementation: 
output

### New Features
* Re-triggers search following check-box selection in Discovery Page Search Bar


jarvisraymond-uchicago:
> feedback from testing: it would be ideal if when field selection is in use, the highlighting feature will only highlight any matches on UI if they are from the fields that are selected

This has been updated. A new function has been added, ```isColumnSearchable``` which is used to compute ```columnIsSearchable``` to determine if renderedCell should have a ```highlightSearchTerm```

If there are configured searchable text fields and selectedSearchableTextFields, it will only highlight the columns that include that searchable text field. Otherwise it will highlight allConfiguredSearchableTextFields

If there are no allConfiguredSearchableTextFields, it will highlight all columns with a falsy contentType or type of String (this was the fallback before).

src/Discovery/Discovery.tsx
```diff
@@ -319,6 +321,9 @@ const Discovery: React.FunctionComponent<Props> = (props: Props) => {
         const formattedFields = formatSearchIndex(field);
         search.addIndex(formattedFields);
       });
+      if (config.studyPreviewField && searchableFields.includes('studyPreviewField.field')) {
```
mfshao, line 324: I'm not sure why these changes are in this file. The `studyPreviewFieldArr` was already going to be added into search on https://github.com/uc-cdis/data-portal/pull/1712/commits/22718225593b2a23807fde1de428994eec098e83#diff-22156db7794691aa0cc7b7966e41924e7f845742f177afd82d81f1fde56cdd93L332

src/Discovery/Discovery.tsx
```diff
@@ -755,6 +759,7 @@ const Discovery: React.FunctionComponent<Props> = (props: Props) => {
           <div id='discovery-table-of-records' className={`discovery-table-container ${filtersVisible ? 'discovery-table-container--collapsed' : 'discovery-table-container--expanded '}`}>
             <Space direction={'vertical'} style={{ width: '100%' }}>
               <DiscoveryListView
+                selectedSearchableTextFields={selectedSearchableTextFields}
```
mfshao, line 762: I think this is the only change we need in this file

src/Discovery/DiscoveryListView.tsx
```diff
@@ -24,14 +25,23 @@ interface Props {
...
   const [onHeightChange, setOnHeightChange] = useState(true);
 
   const scroll = (
     props.config.tableScrollHeight
       ? { scroll: { y: props.config.tableScrollHeight } } : {}
   );
+
+  const isHighlightingEnabled = () => {
+    if (selectedSearchableTextFields.length > 0
+       && !selectedSearchableTextFields.includes('studyPreviewField.field')) {
```
mfshao, line 39: it should check if `selectedSearchableTextFields` has the value as defined in `studyPreviewField.field` in portal config, for example https://github.com/uc-cdis/cdis-manifest/blob/master/healdata.org/portal/gitops.json#L284
Not the raw string of `'studyPreviewField.field'`

PR #1718, opened 2025-08-04
Feat/search bar enhancement cacheing

jarvisraymond-uchicago:
https://ctds-planx.atlassian.net/browse/HP-2182

**Before** 
Switching back and forth between full text search can take several seconds
output

**After** 
Switching back and forth between full text search is faster due to cache
output

### Improvements
* Improves search UX with cache for search restriction selections

### DIFF
--- a/src/Discovery/Discovery.tsx
+++ b/src/Discovery/Discovery.tsx
@@ -309,7 +309,6 @@ const Discovery: React.FunctionComponent<Props> = (props: Props) => {
     });
     // Check if the search object is already cached
     if (searchCache[cacheKey]) {
-      console.log(searchCache);
       setJsSearch(searchCache[cacheKey]);
       props.onSearchChange(props.searchTerm); // Reinitialize search with the cached object
     } else {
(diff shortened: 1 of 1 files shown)


src/Discovery/Discovery.tsx
```diff
@@ -299,46 +299,60 @@ const Discovery: React.FunctionComponent<Props> = (props: Props) => {
...
       }
+      search.addDocuments(props.studies);
+      // expose the search function
+      setJsSearch(search);
+      // Reinitialize search
+      props.onSearchChange(props.searchTerm);
+      // Cache the newly created search object
+      setSearchCache((prevSearchCache) => ({
+        ...prevSearchCache,
+        [cacheKey]: search,
```
mfshao, line 353: I think i get the idea of this change. But I'm not sure if we need to cache all the different searches, because I believe the way this code works now is it will cache all the combanations each time a different field selection has been applied.
I think the most noticeable slowness we are observing now happens when switching the selected search field back from a subset of seclections to the whole collection of them. Maybe we only need to cache the search that is for the whole collection of all searchable fields and reuse it. Could that be more efficient?

jarvisraymond-uchicago: While testing I noticed that the summary description field also caused slowness, so this approach helps with that too. Let me know if you want me to refactor this to only cache the whole collection.

mfshao: I'm still a big concerned about the size of this cached object since it would keeps growing. I don't know if you have tested this and to see if my concern is valid?

jarvisraymond-uchicago: That's a good point. I will investigate and circle back.

PR #1721, opened 2025-08-19
HP-2269 fix don't update search when data is not loaded yet

mfshao:
Link to JIRA ticket if there is one: https://ctds-planx.atlassian.net/browse/HP-2269

### Bug Fixes
- fixed a bug about search bar is initialized with incorrect fields collection
- fixed a bug cause page to freeze when search bar is active and switching between pages


src/Discovery/Discovery.tsx
```diff
@@ -348,7 +359,7 @@ const Discovery: React.FunctionComponent<Props> = (props: Props) => {
       // Reinitialize search
       props.onSearchChange(props.searchTerm);
       // Cache only the Full Text Search object
-      if(selectedSearchableTextFields.length === 0 ) {
+      if (selectedSearchableTextFields.length === allSearchableFields.length) {
```
jarvisraymond-uchicago, line 362: Will this still only cache the Full Text Search object? If not, the comment on line 361 should be updated.

src/Discovery/Discovery.tsx
```diff
@@ -298,11 +303,17 @@ const Discovery: React.FunctionComponent<Props> = (props: Props) => {
...
+  const { searchableTextFields = [], searchableAndSelectableTextFields = {} } = config?.features?.search?.searchBar || {};
+  const allSearchableFields = [...searchableTextFields, ...Object.values(searchableAndSelectableTextFields)] as string[];
+  const [selectedSearchableTextFields, setSelectedSearchableTextFields] = useState(allSearchableFields);
   // Used to cache generated JS search object for studies and selected fields combinations
   const [searchCache, setSearchCache] = useState({});
+  const [searchMode, setSearchMode] = useState(SearchMode.FULL_TEXT);
 
   useEffect(() => {
+    if (!props.studies.length) {
+      return;
```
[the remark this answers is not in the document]
jarvisraymond-uchicago: I tested this locally. I was able to reproduce this error on master branch but not this branch. Success!

src/Discovery/Discovery.tsx
```diff
@@ -660,9 +671,11 @@ const Discovery: React.FunctionComponent<Props> = (props: Props) => {
                   && (
                     <div className='discovery-search-container discovery-header__dropdown-tags-search'>
                       <DiscoveryMDSSearch
+                        searchableTextFields={config.features.search.searchBar.searchableTextFields}
```
jarvisraymond-uchicago, line 674: Per comment above, should this be assigned searchableTextFields instead of config.features.search.searchBar.searchableTextFields?

src/Discovery/Discovery.tsx
```diff
@@ -726,6 +739,11 @@ const Discovery: React.FunctionComponent<Props> = (props: Props) => {
           && (
             <div className='discovery-search-container discovery-search-container__standalone'>
               <DiscoveryMDSSearch
+                searchableTextFields={config.features.search.searchBar.searchableTextFields}
```
jarvisraymond-uchicago, line 742: Per comment above, should this be assigned searchableTextFields instead of config.features.search.searchBar.searchableTextFields?

src/Discovery/Discovery.tsx
```diff
@@ -726,6 +739,11 @@ const Discovery: React.FunctionComponent<Props> = (props: Props) => {
...
+                searchableAndSelectableTextFields={config.features.search.searchBar.searchableAndSelectableTextFields}
```
jarvisraymond-uchicago, line 743: Per comment above should this be assigned searchableAndSelectableTextFields instead of config.features.search.searchBar.searchableAndSelectableTextFields?

src/Discovery/Discovery.tsx
```diff
@@ -660,9 +671,11 @@ const Discovery: React.FunctionComponent<Props> = (props: Props) => {
                   && (
                     <div className='discovery-search-container discovery-header__dropdown-tags-search'>
                       <DiscoveryMDSSearch
+                        searchableTextFields={config.features.search.searchBar.searchableTextFields}
                         searchableAndSelectableTextFields={config.features.search.searchBar.searchableAndSelectableTextFields}
```
jarvisraymond-uchicago, line 675: Per comment above should this be assigned searchableAndSelectableTextFields instead of config.features.search.searchBar.searchableAndSelectableTextFields?

src/Discovery/Discovery.tsx
```diff
@@ -298,11 +303,17 @@ const Discovery: React.FunctionComponent<Props> = (props: Props) => {
     return indexArr;
   };
 
-  const [selectedSearchableTextFields, setSelectedSearchableTextFields] = useState([] as string[]);
+  const { searchableTextFields = [], searchableAndSelectableTextFields = {} } = config?.features?.search?.searchBar || {};
+  const allSearchableFields = [...searchableTextFields, ...Object.values(searchableAndSelectableTextFields)] as string[];
+  const [selectedSearchableTextFields, setSelectedSearchableTextFields] = useState(allSearchableFields);
```
jarvisraymond-uchicago, line 308: Do we want to change the name of the variable ```selectedSearchableTextFields``` if it no longer represents only the user's selections? It looks like it is being initialized here as allSearchableFields instead of the users selections at page load (an empty array before this update).

src/Discovery/Discovery.tsx
```diff
@@ -322,7 +333,7 @@ const Discovery: React.FunctionComponent<Props> = (props: Props) => {
       // in the table and the study description.
       // ---
       const searchableFields = selectedSearchableTextFields;
-      if (searchableFields.length > 0) {
+      if (searchableFields.length > 0 || searchMode === SearchMode.RESTRICTED) {
```
jarvisraymond-uchicago, line 336: Question: do we still need to check for both? Or can we just check for the ```searchMode === SearchMode.RESTRICTED```

jarvisraymond-uchicago (commented):
Code works great based on local testing. 

I posted some questions regarding naming and maintainability, lets discuss briefly synchronously when available.

src/Discovery/Discovery.tsx
```diff
@@ -298,11 +303,17 @@ const Discovery: React.FunctionComponent<Props> = (props: Props) => {
     return indexArr;
   };
 
-  const [selectedSearchableTextFields, setSelectedSearchableTextFields] = useState([] as string[]);
+  const { searchableTextFields = [], searchableAndSelectableTextFields = {} } = config?.features?.search?.searchBar || {};
```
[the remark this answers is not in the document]
mfshao: Renamed `selectedSearchableTextFields` to `selectedFieldsForSearchIndexing`
Assigned the new variables to `DiscoveryMDSSearch` component props instead of redeclaring them
Reused `selectedSearchableTextFields` instead of creating new variable `searchableFields`

I would keep `searchableTextFields` and `searchableAndSelectableTextFields` since they are coming from portal config directly

src/Discovery/Discovery.tsx
mfshao: we need to check for both since `searchableFields.length > 0` is for the default case for envs that doesn't use restricted search mode

src/Discovery/Discovery.tsx
```diff
@@ -298,14 +303,20 @@ const Discovery: React.FunctionComponent<Props> = (props: Props) => {
...
+  const allSearchableFields = [...searchableTextFields, ...Object.values(searchableAndSelectableTextFields)] as string[];
+  const [selectedFieldsForSearchIndexing, setSelectedFieldsForSearchIndexing] = useState(allSearchableFields);
   // Used to cache generated JS search object for studies and selected fields combinations
   const [searchCache, setSearchCache] = useState({});
+  const [searchMode, setSearchMode] = useState(SearchMode.FULL_TEXT);
 
   useEffect(() => {
+    if (!props.allBatchesAreReady) {
```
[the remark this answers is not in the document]
jarvisraymond-uchicago: Good one!

PR #1734, opened 2025-10-03
render the email address instead of raw oidc id

flashguerdon:
### Improvements
Currently the Portal header and profile popover display the raw OIDC username claim (e.g. google-oauth2|12345678901234567890). This is confusing for users. Fence already returns the user’s email in /user. We should render the email address instead, falling back to other available claims.
**Before:**
image
**After:**
image

### DIFF
--- a/src/components/layout/TopBar.jsx
+++ b/src/components/layout/TopBar.jsx
@@ -35,8 +35,8 @@ class TopBar extends Component {
 
   render() {
     const { user } = this.props;
-    // Prefer email, then preferred_username, then name, then username
-    const userLabel = (user && (user.email || user.preferred_username || user.name || user.username)) || '';
+    // Prefer: preferred_username > display_name > name > username > email
+    const userLabel = (user && (user.preferred_username || user.display_name || user.name || user.username || user.email)) || '';
     return (
       <div className='top-bar'>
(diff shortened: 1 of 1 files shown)


src/components/layout/TopBar.jsx
```diff
@@ -34,6 +34,9 @@ class TopBar extends Component {
   isActive = (id) => this.props.activeTab === id;
 
   render() {
+    const { user } = this.props;
+    // Prefer email, then preferred_username, then name, then username
+    const userLabel = (user && (user.email || user.preferred_username || user.name || user.username)) || '';
```
mfshao, line 39: We might need an internal sync but imo `preferred_username` should have the highest priorities here among these other fields here. Based on their definations in fence https://github.com/uc-cdis/fence/blob/2a0ede3587ca64cc856991ccdf7c39202e15a2c7/fence/resources/user/__init__.py#L91-L94, I think the order should be `preferred_username > display_name > name > username > email`
And for your use case, instead of trying to patch this from FE side, maybe a better solution is to update the `preferred_username` properly in Fence when the user's record gets created

mfshao (approved):
Put through CI by https://github.com/uc-cdis/data-portal/pull/1738 and it has passed

PR #1736, opened 2025-10-14
fix(checkJobStatus_with_undefined_jobId): update checkJobStatus to handle undefined jobId

markxiao:
https://ctds-planx.atlassian.net/browse/PD-22

### Bug Fixes

- If the jobId is falsy, end checkJobStatus process before the API call .

### DIFF
--- a/src/Analysis/AnalysisJob.js
+++ b/src/Analysis/AnalysisJob.js
@@ -39,6 +39,15 @@ export const checkJobStatus = (dispatch, getState) => {
   if (state.analysis.job) {
     jobId = state.analysis.job.uid;
   }
+  // If jobId is null or undefined, clear the interval and exit
+  if (!jobId) {
+    clearInterval(state.analysis.jobStatusInterval);
+    dispatch({
+      type: 'FETCH_ERROR',
+      error: 'Job ID not found',
+    });
+    return Promise.resolve();
+  }
   return fetchWithCreds({
     path: `${jobAPIPath}status?UID=${jobId}`,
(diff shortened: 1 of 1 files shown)


src/Analysis/AnalysisJob.js
```diff
@@ -39,6 +39,15 @@ export const checkJobStatus = (dispatch, getState) => {
...
     jobId = state.analysis.job.uid;
   }
+  // If jobId is null or undefined, clear the interval and exit
+  if (!jobId) {
+    clearInterval(state.analysis.jobStatusInterval);
+    dispatch({
+      type: 'FETCH_ERROR',
+      error: 'Job ID not found',
+    });
+    return Promise.resolve();
```
ocshawn: should this be placed in resolve or return the dispatch? this seems redundant but it has been a while since i have worked with promises

mfshao: Yeah we can try to return the dispatch, I cannot recall if the dispath here is thenable, I think it needs to comes from a redux middleware
But if it wonm't work, I'm also fine with this solution

markxiao: Tested with the following code:

```
  if (!jobId) {
    clearInterval(state.analysis.jobStatusInterval);
    return dispatch({
      type: 'FETCH_ERROR',
      error: 'Job ID not found',
    });
  }
```

Got error:
`utils.js:69 Uncaught TypeError: lambda(...).then is not a function`

Looks like we need to return Promise.

PR #1796, opened 2026-04-15
HP-2603 Feat/manifest filename

mfshao:
Link to JIRA ticket if there is one: https://ctds-planx.atlassian.net/browse/HP-2603

### Improvements
- Discovery: downloaded file manifests now will have hostname, identifier and date information in their filenames
- Discovery: multiple file manifests generated for downloading will be put into a single archive file


src/Discovery/DiscoveryDetails/Components/FieldGrouping/TabField/DataDownloadList/ActionButtons/DownloadUtils/GenerateFilename.test.ts
```diff
@@ -71,6 +71,12 @@ describe('GenerateFilename', () => {
       );
     });
 
+    test('omits studyID segment for "all_manifests"', () => {
```
jarvisraymond-uchicago, line 73: FYI, this file is failing ESLINT with error: 
`  1:8  error  'moment' is defined but never used  @typescript-eslint/no-unused-vars
`
This is because moment is imported but not used, just mocked. 

ESLINT can be run locally with
`npm run eslint`
or
`npm run eslint-new` 
to test only files that differ from master

src/Discovery/DiscoveryActionBar/utils/handleDownloadManifestClick.ts
```diff
@@ -16,11 +18,17 @@ const handleDownloadManifestClick = (
...
+
   if (healIDPLoginNeeded) {
     return;
   }
-  // combine manifests from all selected studies
-  const { fileManifest } = assembleFileMetadata(manifestFieldName, selectedResources);
+
+  const fileManifestsWithNames = processFileMetadata(uidFieldName, manifestFieldName, selectedResources);
+  if (!Object.keys(fileManifestsWithNames).length) {
+    return;
```
jarvisraymond-uchicago, line 29: Question: in this situation would be want to show an error to the end user? Also the same with line 23. 

If this branch executes, clicking the button causes nothing to happen and the UI doesn't alert the user that anything is wrong. 

Image

jarvisraymond-uchicago (commented):
Overall code looks good! Functionality works as expected, identified an edge case we might want to address and a lint error. 

* Confirmed downloading single manifest files results in file name format specified in ticket
* Confirmed downloading multiple manifest files results in file name format specified in ticket
* Confirmed updates to JEST tests pass
* Noted ESLINT issue in `src/Discovery/DiscoveryDetails/Components/FieldGrouping/TabField/DataDownloadList/ActionButtons/DownloadUtils/GenerateFilename.test.ts`
* Noted confusing UX edge case in `src/Discovery/DiscoveryActionBar/utils/handleDownloadManifestClick.ts`

src/Discovery/DiscoveryActionBar/utils/handleDownloadManifestClick.ts
mfshao: I don't think so, because this is a check ofor the component that conceptually should not be triggered, becuase in the actual download button component there is another guard condition https://github.com/uc-cdis/data-portal/blob/master/src/Discovery/DiscoveryActionBar/components/DownloadManifestButton.tsx#L71, which should just disable the download button if there is no manifests in selected studies

jarvisraymond-uchicago (approved):

PR #1815, opened 2026-06-09
chore(updateAnalysisLink): Added external link indicator

jarvisraymond-uchicago:
https://ctds-planx.atlassian.net/browse/HP-2677

output

**Note**
Navigation section of gitops.json will need to be updated with an optional newWindow boolean set to true for any navigation links that need to be opened in a new window, i.e. 
```
      {
          "name": "Example Analyses",
          "link": "/resource-browser",
          "newWindow": true,
          "icon": "bar-chart",
          "tooltip": "Learn how to use Jupyter Notebooks to explore and visualize data available on the HEAL platform by running a static tutorial notebook, or use one of these examples as a launching point for your own custom analysis."
        },
```

### Improvements
- Example Analyses tab: Link now opens in new tab with external link indicator

### DIFF
--- a/src/components/layout/NavButton.jsx
+++ b/src/components/layout/NavButton.jsx
@@ -3,17 +3,12 @@ import PropTypes from 'prop-types';
 import { NavLink } from 'react-router-dom';
 import IconComponent from '../Icon';
 import './NavButton.less';
+import ExternalLinkIndicator from '../ExternalLinkIndicator';
 
-const NavButton = ({
-  dictIcons,
-  item,
-}) => {
+const NavButton = ({ dictIcons, item }) => {
   if (item.link.startsWith('http')) {
     return (
-      <a
-        href={item.link}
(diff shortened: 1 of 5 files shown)


src/components/layout/NavButton.jsx
```diff
@@ -26,11 +21,14 @@ const NavButton = ({
       className={'body-typo nav-button'}
       activeClassName='button-active'
       to={item.link}
+      target={item.newWindow ? '_blank' : '_self'}
```
mfshao, line 24: can you update https://github.com/uc-cdis/data-portal/blob/master/docs/portal_config.md?plain=1#L138-L149 with this new option?

jarvisraymond-uchicago: Good idea, updated the navigation section of the documentation
image

mfshao (approved):

PR #1863, opened 2026-08-05
[PPS-3845] upgrade to relay 14

SpencerAxelrod:
Link to JIRA ticket if there is one: https://ctds-planx.atlassian.net/browse/PPS-3845

### Dependency updates
Upgrade to Relay 14. Relay 13+ does not have the `immutable` dependency, which addresses ticket.

Important changes:
- top level `"relay"` key/config added to package.json.
- Relay 14 requires SDL. Extracted the JSON -> SDL conversion logic from `getSchema.js` into a standalone `generateSchemaSDL.js `module so it can be run at build time.
- Relay 13+ turns `fetchQuery` from Promise to Observable so `.then(...)` is no longer a function on the result. Updated to insert `.toPromise().`


craigrbarnes (approved):
Looks good. Thank you for updateing this dependency

